PR #32, opened 2016-06-04
Fix some docs + thoughts on the doc

Keats:
Hey, found a few typos while the reading the docs.

A few thoughs as well:

- you mention `without making progress` a few times in the grammar but don't explain what you mean by it, does the parsing stop
- you call `parse.exp()` a few times but never say what is it, i guess it's expected? (I like long names)
- does `[""]` match everything?
- when talking about silent rules you say they don't appear in `Parser::queue` or `Parser::expected` . I can imagine what the queue is but not sure about expected
- `// exp is silent since we already know what we're parsing` in the `process` doc. I didn't know, it just says something about nested letter and you see the example only at the end. Also from the point above, I thought silent rules were not added to the stream of token and yet they are there in `process!` so I'm a bit confused

### DIFF
--- a/src/grammar.rs
+++ b/src/grammar.rs
@@ -38,7 +38,7 @@
 /// ab = @{ a ~ b }
 /// ```
 ///
-/// In the rule above, there cannot be any between `a` and `b` for `ab` to match. This rule has a
+/// In the rule above, there cannot be anything between `a` and `b` for `ab` to match. This rule has a
 /// *cascading* effect, so any rules called from an atomic rule will also be atomic while being
 /// matched in an atomic context. In other words, `a` and `b` will also be atomic when being
 /// matched inside of `ab`.
@@ -69,7 +69,7 @@
 /// | `&a`         | matches `a` without making progress                  |
 /// | `!a`         | matches if `a` doesn't match without making progress |
 ///
-/// ## Precendece climbing
+/// ## Precedence climbing
 ///
 /// pest supports a special type of rule that implements precedence climbing in the background.
 ///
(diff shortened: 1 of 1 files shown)


dragostis:
Thanks a lot for the PR. 😃 I'll try and fix points you mentioned on a new branch.

dragostis:
I think the best way to explain the progress would be to have a full example from start to end in the main library. That would explain making progress, because explaining it in grammar seems a bit too much.

The example with `exp` is probably not the best. `exp` stands for expression. I'll [rename](issues/33) it to get rid of the confusion.

`[""]` captures nothing. It's a rule that simply matches no matter what. This again is probably bad and should be [removed](issues/35) from the `expression`/`paren` example.

`Parser::expected` should probably be linked here. Anyway, `expected` may [not be the best name](issues/36) for the method. The method simply returns what the parser would expect to find in terms of rules at the last position of the input where it failed to parse.

`exp` is silent there because you're calling `exp`. So instead of generating a `Token` for it, you don't have to. You're parsing an expression and that's the only thing that this particular parser can parse.

Keats:
> I think the best way to explain the progress would be to have a full example from start to end in the main library. That would explain making progress, because explaining it in grammar seems a bit too much.

I think so too, https://github.com/nikomatsakis/lalrpop/blob/master/doc/tutorial.md is a very clear one for example

dragostis:
@Keats I'm working on the calculator example right now and it's almost done. I'll post here once I push it to master.

dragostis:
@Keats I've added the example in the [crate's doc](http://dragostis.github.io/pest/pest/). I've also fixed a few other issues. Please let me know how this works out and where it needs improvement.

PR #47, opened 2016-06-10
Added separate matchers.

dragostis:
@Keats Please take a look and let me know how this looks.


src/process.rs
```diff
@@ -134,11 +137,147 @@
 ///            Expression::Paren(Box::new(Expression::Paren(Box::new(Expression::Letter('z'))))));
 /// # }
 /// ```
+///
+/// ### Sentence
+///
+/// To showcase the use of multiple matchers, we'll use a setence grammar:
```
Keats: sentence

src/process.rs
```diff
@@ -134,11 +137,147 @@
...
+///     Letter(char)
+/// }
+/// ```
+///
+/// To build the `Token` processor, let's use a bottom up aproach when writing the matchers. Let's
+/// start by building a `word` matcher. We'll call it `_word` in order not to clash with the name
+/// of the rule that also gets defined as a method on the `Parser`.
+///
+/// ```ignore
+/// _word(&self) -> LinkedList<Node> { // return LinkedList<Node> to buil Word with
```
Keats: where do you put that? in `process!`?

Keats: and `to build`

src/process.rs
```diff
@@ -134,11 +137,147 @@
...
+///     (&head: letter, @tail) => { // usual tail recursion; &head is captured,
+///         let mut tail = tail;    // tail is recursive
+///         tail.push_front(Node::Letter(head.chars().next().unwrap()));
+///
+///         tail
+///     },
+///     () => {                // if rule above doesn't match, there are no more letters to
```
Keats: would hving the `match` keyword in there easier to understand that it's doing some kind of pattern matching?

src/process.rs
```diff
@@ -134,11 +137,147 @@
...
+///         LinkedList::new()  // process; return empty list
+///     }
+/// }
+/// ```
+///
+/// Processing a `sentence` is similar, only this time `head` will be a `_word` call.
+///
+/// ```ignore
+/// _sentence(&self) -> LinkedList<Node> {
+///     (_: word, @<_word>head, @tail) => { // match word Token then call _word
```
Keats: `@<_word>head` feels like a very magical syntax

src/process.rs
```diff
@@ -9,18 +9,21 @@
...
+///
+/// | Item        | What it does                                       |
+/// |-------------|----------------------------------------------------|
+/// | `item`      | matches any `Token`                                |
+/// | `item: R`   | matches a `Token` of rule `R`                      |
+/// | `&item`     | captures a `Token`                                 |
+/// | `&item: R`  | captures a `Token` of rule `R`                     |
+/// | `_`         | skips a `Token`                                    |
+/// | `_: R`      | skips a `Token` of rule `R`                        |
+/// | `@item`     | call method recursively and store result in `item` |
```
Keats: i'm not sure i understand that one

src/process.rs
dragostis: You place it outside of the `impl_rdp!` macro. It's specified in the last example.

Also, what do you mean by `to build`?

src/process.rs
```diff
@@ -134,11 +137,147 @@
...
+/// }
+/// ```
+///
+/// To build the `Token` processor, let's use a bottom up aproach when writing the matchers. Let's
+/// start by building a `word` matcher. We'll call it `_word` in order not to clash with the name
+/// of the rule that also gets defined as a method on the `Parser`.
+///
+/// ```ignore
+/// _word(&self) -> LinkedList<Node> { // return LinkedList<Node> to buil Word with
+///     (&head: letter, @tail) => { // usual tail recursion; &head is captured,
```
Keats: so i'm not sure i understand what's happening there and the relation between `head` and `tail`. where do they end up?
At what point does `head` gets redefined? Is that a concept that `pest` knows about and handles automatically?

src/process.rs
Keats: you have a typo (`to buil`)

src/process.rs
dragostis: I've thought about it, but I think having a `match` keyword would be confusing since you're not matching against an object, rather against an arbitrary number of items at the start of the queue. Maybe once [slice patterns](https://doc.rust-lang.org/book/slice-patterns.html) get stable it will be worth changing the syntax, but this remains to be seen.

src/process.rs
dragostis: I know. Any better ideas for it? I was thinking of changing the `@` rule to `head = _word()` and making the function that you want to call explicit. What do you think?

src/process.rs
dragostis: I'm trying to encourage the use of tail recursive calls here. You're basically matching against a letter that you take the slice of (with the help of `&`) and then calling `_word` recursively and saving the result in `tail`.

dragostis: `head` and `tail` are simple variable names.

src/process.rs
dragostis: Fixed.

src/process.rs
dragostis: I kind of like the idea of using `head: _word()` and making calls explicit.

dragostis:
@Keats I've changed the matcher call syntax and I'm pretty satisfied with the way the syntax looks.

Keats:
It looks ok, maybe you can merge that and i'll depend on master branch for now so we can see if some stuff needs to be changed while doing it for Tera?

dragostis:
@Keats I'm going to do a release with this. If anything needs being added, we'll do another one.

PR #61, opened 2016-06-19
v0.3.0

dragostis:


sunng87:
I'm not 100% but it seems in sentence_lifetime.rs the updated rdp._word() results shares same lifetime with `rdp` instead of `file` in master branch.

If you change the test to:

```rust
    let mut parser = Rdp::new(StringInput::new(&file));
    let result = {
        assert!(parser.word());
        parser._word()
    };
```

This may be related to updated `Parser` and `Rdp`.

dragostis:
@sunng87 Yes, it definitely is. The problem is that it would be nice for the lifetime of the of the actual `&str` to be used. The old solution caused problems when I removed all the delegated methods from `Parser` and used `Parser::input()` and `Parser::input_mut()`.

Do you see any way to tie the `Input` reference from `Parser::input()` and `Parser::input_mut()` to that `&str` without adding lifetimes to every single thing? Maybe owning the `Input` inside of `Rdp` is not the best solution.

PR #62, opened 2016-06-22
Adding some lifetime back

sunng87:
Seems the key lifetime symbol is Input slice() -> &str. If we don't specify it, it will use parser's default scope.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -6,7 +6,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 /// A `trait` that defines an input for a `Parser`.
-pub trait Input {
+pub trait Input<'a> {
     /// Returns length of an `Input`.
     fn len(&self) -> usize;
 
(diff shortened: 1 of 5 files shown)


sunng87:
Sorry I have rustfmt configured so there is some format change, just ignore it. (And I recommend you to use rustfmt as well )

dragostis:
Thanks for the PR! I'm wondering if there's a way to make this work without having to add that lifetime to `Node`.

Also, in order not to add confusion, a good idea would be to name it something different from `'a`. This is particularly annoying because the lifetime appears to come from nowhere. I'm still trying to see if this is the best solution.

sunng87:
I'm not sure if it's possible. The lifetime symbol on `Node` means it holds a reference with same lifetime with the string the Input holds. It's the macro hides the `<'a>` on `Rdp`. 

One idea comes to my mind is to add omit `<'a>` within `process!` and add it via macro magic. But that will bring more problems I believe.

dragostis:
Lifetimes are not yet properly capturable in macros, but I'm satiafied with the current implementation.

PR #69, opened 2016-06-30
V0.4.0

dragostis:


Cargo.toml
```diff
@@ -7,3 +7,6 @@ repository = "https://github.com/dragostis/pest"
 documentation = "http://dragostis.github.io/pest/pest/"
 keywords = ["pest", "parser", "peg", "grammar"]
 license = "MPL-2.0"
+
+[dependencies]
+clap = "2.8.0"
```
sunng87: I suggest to use a feature flag `grammar_check` and make clap a optional dependency.

dragostis: Good point. One of the few things that I tried to keep was 0 dependencies. Still, I don't know if features will do the trick here because you can't really do conditional compilation for logic and I need argument parsing in the tool.

Binary dependencies would be great, but they're not yet implemented in Cargo.

If you have a good idea how to do this, please leave a comment.

sunng87: There is an example in handlebars that deals with both rustc_serialize and serde https://github.com/sunng87/handlebars-rust/blob/master/examples/render_file.rs. 

You can wrap most of main mod in a `#cfg(feature=...)` block so it won't compile if the feature is turned off.

dragostis: The only issue I see here is that you probably won't be able to run `cargo install pest` to get the binary since the main function will be behind a feature, so you will probably have to provide the feature when installing.

sunng87: Yes, if you want that you'd better to make it a separated crate say `pest-checker`.

dragostis: @sunng87 I was thinking of making it a general `pest` tool, so the `pest` name would be best. I'll probably have to resort to the `pest-tool` name.

tests/non_silent.rs
```diff
@@ -0,0 +1,48 @@
...
+#[macro_use]
+extern crate pest;
+
+use pest::prelude::*;
+
+impl_rdp! {
+    grammar! {
+        abc = { ["a"] ~ ["b"] | ["a c"] }
+
+        whitespace = { [" "] }
```
Keats: will there be a sigil (like `_` or `@`) to ignore whitespace token in a given rule?

dragostis: You can ignore it with `whitespace = _{...}`.

Keats: I meant inside a rule, let's say i want to keep whitespace in general but not between a given rule, something like `if_tag = { tag_start ~ ["if"] ~ expression ~ tag_end }` where i do not care about whitespace in between but it can exist.

dragostis: @Keats Whitespace will work just as it did until now, without automatically being matched before rules. So you're `if_tag` rule can have whitespace between sub-rules. If you don't want whitespace between sub-rules, you use atomic.

Keats: got it, thanks

PR #73, opened 2016-07-20
Add queue_with_captures(), similar to queue() except it contains the value

### LINKED ISSUE #72: Add a queue_debug on parser
Right now we can get the `.queue()` method to see the list of tokens but when debugging I want to know what's the actual text for that token since I'm doing some pre-processing. If I print the queue I get
```
[
    Token {
        rule: content,
        start: 0,
        end: 27
    },
    Token {
        rule: text,
        start: 0,
        end: 27
    },
    Token {
        rule: content,
        start: 27,
        end: 45
    },
...
```

It would be nice to have a `parser.queue_debug()` that you could print and would look like so:
```
[
    Token {
        rule: content,
        start: 0,
        end: 27,
        text: "blabla",
    },
    Token {
        rule: text,
        start: 0,
        end: 27,
       text: "blabla",
    },
    Token {
        rule: content,
        start: 27,
        end: 45,
       text: " \n hello",
    },
...
```

Keats:
For debugging help.
Fix #72

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -31,6 +31,9 @@ pub trait Parser<'a, T: Input<'a>> {
     /// Returns the mutable queue of all matched `Token`s.
     fn queue_mut(&mut self) -> &mut Vec<Self::Token>;
 
+    /// Returns the queue of all matched `(Token, value)`s.
+    fn queue_with_captures(&self) -> Vec<(Self::Token, String)>;
+
     /// Returns the current index within the queue. Used in `process!`.
     fn queue_index(&self) -> usize;
 
--- a/src/parsers/rdp.rs
+++ b/src/parsers/rdp.rs
@@ -238,13 +238,21 @@ macro_rules! impl_rdp {
             }
 
             #[inline]
-            fn queue(&self) -> &Vec<Token<Rule>>{
+            fn queue(&self) -> &Vec<Token<Rule>> {
                 &self.queue
             }
 
             #[inline]
-            fn queue_mut(&mut self) -> &mut Vec<Token<Rule>>{
+            fn queue_mut(&mut self) -> &mut Vec<Token<Rule>> {
                 &mut self.queue
+            }
+
+            fn queue_with_captures(&self) -> Vec<(Token<Rule>, String)> {
+                self.queue
+                    .clone()
+                    .into_iter()
+                    .map(|t| (t, self.input().slice(t.start, t.end).to_owned()))
+                    .collect()
             }
 
             #[inline]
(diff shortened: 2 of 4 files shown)


src/parsers/rdp.rs
```diff
@@ -237,13 +237,21 @@ macro_rules! impl_rdp {
                 self.fail_pos = 0;
             }
 
+            fn queue_debug(&self) -> Vec<(Token<Rule>, String)> {
+                self.queue
+                    .clone()
+                    .into_iter()
+                    .map(|t| (t, self.input().slice(t.start, t.end).to_string()))
```
dragostis: Even though `to_string` has been optimized in the latest versions, I'd still use `to_owned` since it makes more sense in terms of terminology.

src/parsers/rdp.rs
```diff
@@ -237,13 +237,21 @@ macro_rules! impl_rdp {
...
+                    .collect()
+            }
+
             #[inline]
-            fn queue(&self) -> &Vec<Token<Rule>>{
+            fn queue(&self) -> &Vec<Token<Rule>> {
                 &self.queue
             }
 
             #[inline]
```
dragostis: No need to inline since this will be used only for debugging.

src/parsers/rdp.rs
```diff
@@ -475,4 +483,24 @@ mod tests {
...
+
+    #[test]
+    fn queue_debug() {
+        let mut parser = Rdp::new(StringInput::new("(  ( ))(( () )() )() "));
+
+        assert!(parser.expression());
+        assert!(!parser.end());
+
+        let queue = vec![
+            (Token::new(Rule::paren, 0, 7), "(  ( ))".to_string()),
```
dragostis: Same `to_owned` idea.

src/parsers/rdp.rs
Keats: this is `queue_mut` inline

src/parser.rs
```diff
@@ -25,6 +25,9 @@ pub trait Parser<'a, T: Input<'a>> {
     /// Reset a `Parser`.
     fn reset(&mut self);
 
+    /// Returns the queue of all matched `DebugToken`s for debug purposes
```
dragostis: No more `DebugToken`. 😛 Also don't forget about the full stop.

src/parsers/rdp.rs
dragostis: Nice! 👍 I didn't notice.

src/parsers/rdp.rs
Keats: that's interesting, I find `to_string` more explicit myself. I can change it though

dragostis: It is explicit, but someone new to Rust might find it odd to see `"I am already a string".to_string()`. I still kind of do. And also the fact that most of the code already uses the `to_onwed` method makes me lean towards this one.

src/parser.rs
```diff
@@ -25,6 +25,9 @@ pub trait Parser<'a, T: Input<'a>> {
...
+    fn queue_debug(&self) -> Vec<(Self::Token, String)>;
```
dragostis: This method might actually be useful in other cases apart from debugging, so maybe it would be a good idea to name it `queue_with_captures`?

Keats:
updated

src/parser.rs
```diff
@@ -25,6 +25,9 @@ pub trait Parser<'a, T: Input<'a>> {
     /// Reset a `Parser`.
     fn reset(&mut self);
 
+    /// Returns the queue of all matched `(Token, value)`s
```
dragostis: I usually used full stop at the end of documentatiom descriptions mostly because you might need to use multiple sentences and form paragraphs, so this should have one as well.

src/parser.rs
```diff
@@ -25,6 +25,9 @@ pub trait Parser<'a, T: Input<'a>> {
...
+    fn queue_with_captures(&self) -> Vec<(Self::Token, String)>;
```
dragostis: As I said in an outdates comment, this might actually be useful outside of a debugging context so it might be useful to call it `queue_with_captures`?

src/parsers/rdp.rs
```diff
@@ -237,13 +237,21 @@ macro_rules! impl_rdp {
                 self.fail_pos = 0;
             }
 
+            fn queue_with_captures(&self) -> Vec<(Token<Rule>, String)> {
```
dragostis: Same ordering here.

Keats: done

dragostis:
Looks good. Thank you sl much for the PR! 😄

PR #89, opened 2016-12-05
Show an example how to use different associativities.

ehiggs:
Repeat of @86 to fix @85 on request of @dragostis


dragostis:
Thanks you for your patience.

PR #104, opened 2017-02-23
Adding documentation to process + better panic message

sunjay:
When I was using `process!`, I had no idea that I had to call one of the methods from `grammar!` before I could call a process method. The generated error message on the empty queue was no help at all.

This pull request is my attempt at making the documentation closer to what I wished it had been. I also tried to improve the error message. Hopefully this helps people in the future.

I ran `cargo doc` and this looks good locally. Feedback is welcome!


src/process.rs
```diff
@@ -50,11 +86,20 @@
 ///
 /// let mut parser = Rdp::new(StringInput::new("b"));
 ///
+/// // Notice that you have to call b() here so that there are some tokens in the queue
```
dragostis: I usually use full sentences in comments.

src/process.rs
```diff
@@ -50,11 +86,20 @@
...
 /// parser.b();
 /// parser.ab();
 /// # }
 /// ```
 ///
+/// The reason for using a `panic!` instead of a `Result` is because this error is not common
```
dragostis: I guess you could say here that the `panic!` useful for debugging purposes and I would definitely mention the fact that match mistakes should not be handled at runtime here (with `Result`) because this is what the parser should be doing.

sunjay:
I addressed your comments! :+1: 

If there is nothing else, feel free to merge. :) I also suggest publishing a new version of the docs to the gh-pages branch once this is merged.

Issue #103, opened 2017-02-21
add some basic types to pest so it improves grammar readability


lwandrebeck:
I think having some already defined types would help to enhance readability.
alpha,
decdigit,
hexdigit,
octdigit,
alphanumeric,
space,
signedint,
unsignedint,
signedfloat,
unsignedfloat,
singlequoted,
doublequoted

and maybe a couple others, but these should cover most common needs.

dragostis:
I've assigned this myself for now, but would you be willing to try this out once 1.0 is released?

Keats:
I'll take it if no one gets to it first

lwandrebeck:
`impl_rdp! {
    grammar! {
        plus   = { ["+"] }
        minus  = { ["-"] }
        sign = { minus | plus }
        alpha = { i['a'..'z'] }
        minalpha = { ['a'..'z'] }
        capalpha = { ['A'..'Z'] }
        digit = { ['0'..'9'] }
        alphanum = { alpha | digit }
        minalphanum = { minalpha | digit }
        capalphanum = { capalpha | digit }
        octdigit = { ['0'..'7'] }
        hexdigit = { ['0'..'9'] | i['a'..'f'] }
        uint = { ['0'] | ( ['1'..'9'] ~ digit* ) }
        sint = { sign ~ uint }
        exp = { i['e'] }
        ufloat = { uint ~ ['.'] ~ digit* ~ ( exp ~ ( uint | sint ))? }
        sfloat = { sign ~ ufloat }
        octnum = { ['0'] ~ octdigit+ }
        hexnum = { ["0x"] ~ hexdigit+ }
        dquoted = { !["\""] ~ any }
        squoted = { !['\''] ~ any }
        btquoted = { ![`\``] ~ any }
    }
}
`

Here is a quick shot at it. I’d be very willing to test it once 1.0 is released, as I need it for my pet project.
@Keats feel free to get to it :)

dragostis:
It would probably be a good idea to have pest default rules look differently compared with your own rules. This would definitely make writing grammars a bit more predictable.

lwandrebeck:
Feel free to pick up some, all, rename rules. I guess this would have to be coded in « pure » rust instead of already defined pest grammar ?

dragostis:
@lwandrebeck Not necessarily. But there should be an obvious difference between them, I'd say.

Keats:
Maybe we can use `::` as delimiter, like `:alpha:`, `:int:` etc?

[...]

PR #108, opened 2017-03-14
Added Position and Span.

dragostis:
Added `Position` and `Span` for a leaner way to deal with `Input`s.


pest/src/lib.rs
```diff
@@ -5,23 +5,38 @@
...
+pub mod inputs {
+    pub use super::inputs_private::{Input, Position, Span, StringInput};
 }
 
+/// A `mod` containing iterators and constructs to aid in parser output manipulation.
+pub mod iterators {
+    pub use super::iterators_private::{Pair, Pairs, TokenIterator};
+}
+
+/// A `trait` which parser rules must implement.
```
sunjay: Note that I added some documentation here to clarify *why* this trait works. Let me know if you have any feedback.

pest/examples/parens.rs
```diff
@@ -0,0 +1,105 @@
+extern crate pest;
```
sunjay: Is this an example of what the generated code will be like? If so, some of my comments about this may not apply.

pest/examples/parens.rs
```diff
@@ -0,0 +1,105 @@
...
+    expr,
+    paren,
+    paren_end
+}
+
+struct ParenParser;
+
+impl Parser<Rule> for ParenParser {
+    fn parse<I: Input>(rule: Rule, input: Rc<I>) -> Result<Pairs<Rule, I>, Error<Rule, I>> {
+        fn expr<I: Input>(
```
sunjay: Why are these functions defined inside `parse()`? Would it be possible to avoid the extra nesting and move them out? They don't seem to be benefiting from the nesting. Though they are hidden from outside the function, there are other ways to accomplish that without making this function so large.

pest/examples/parens.rs
sunjay: In particular, I'm worried about the deeply nested functions. I wouldn't wish debugging that on anyone. If it's auto generated then it's probably fine because we'll be the only ones having to debug the generated output.

pest/src/inputs_private/input.rs
```diff
@@ -0,0 +1,83 @@
...
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::fmt::Debug;
+use std::ops::Range;
+
+/// A `trait` that defines an input for a `Parser`. It should be implemented by custom input sources
+/// as minimally and as efficiently possible. `unsafe` methods should not be called directly; in
```
dragostis: I've added a comment clarifying the motivation behind. I'm not completely convinced that this is the best method to handle things, however.

pest/src/inputs_private/string_input.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// assert_eq!(input.len(), 3);
+    /// ```
+    pub fn new(string: String) -> StringInput {
+        StringInput {
+            string: string
+        }
+    }
+}
+
+impl Input for StringInput {
```
dragostis: The inlining is only a hint and it's vital for performance. The tag is used here to let the compiler know that I'd like to have it consider all this API for inlining, even though we're currently at link-time. (since pest will be used as a separate crate which *will* call `Position` and `ParserState` APIs)

pest/examples/parens.rs
dragostis: It will be auto-generated by the derive crate.

pest/examples/parens.rs
dragostis: I totally agree, but I suppose it will be easier to generate like this from the derive. The point is that the user should be able to implement whatever methods he wants on his derived `struct` and defining them outside of this function will flood *some* kind of namespace whichever way you take it.

dragostis:
@sunjay Thanks a lot for the review and awesome job with the CI which was a particular pain point for me. Let me know if you find any better solution for the `Input` issue. I've been considering all kinds of solutions, including making `String` an `Input` itself. My current issues with the current design:

  * There is a ton of unsafe that needs to be there, in the `trait`. Sure, this makes way for the safe and elegant `Position` API, but like this you need to deal with unsafe both in the implementation of a `T: Input` and in `Position`. (which to some degree makes sense, since Position has different safety considerations)
  * It would be nice to have a way to define `parse_to_asyncs` that only takes `StringInput`s, but I'm not completely sure whether everything can be made explicit all the way down the pipeline.

pest/src/lib.rs
dragostis: This is really beneficial. Another change that I'm waiting for is [`pub(restricted)`](https://github.com/rust-lang/rfcs/blob/master/text/1422-pub-restricted.md) which will land in 1.17 in order to get rid of the ugly `*_private` directories.

pest/src/inputs_private/string_input.rs
dragostis: I haven't documented this yet since I'm not so sure how to approach this. On one hand, I'd really like this to be as concise and short as possible and put this as a comment in the `inputs` `mod`. On the other hand, I'd like this to be really easy to figure out if they're hacking on pest and wondering what all these inlines are all about, so a comment documenting each one is also something that I would consider.

pest/src/inputs_private/position.rs
```diff
@@ -0,0 +1,520 @@
...
+
+impl<I: Input> Clone for Position<I> {
+    fn clone(&self) -> Position<I> {
+        new(self.input.clone(), self.pos)
+    }
+}
+
+impl<I: Input> PartialEq for Position<I> {
+    fn eq(&self, other: &Position<I>) -> bool {
+        &*self.input as *const I == &*other.input as *const I &&
```
dragostis: Use `Rc::ptr_eq` instead of raw equality.

pest/src/error.rs
```diff
@@ -5,115 +5,266 @@
...
+                (false, false) => {
+                    format!(
+                        "unexpected {}; expected {}",
+                        enumerate(negatives),
+                        enumerate(positives)
+                    )
+                }
+                (false, true) => format!("unexpected {}", enumerate(negatives)),
+                (true, false) => format!("expected {}", enumerate(positives)),
+                (true, true) => "inexplicit parsing error".to_owned()
```
Keats: inexplicit -> unknown?

pest/src/inputs_private/position.rs
```diff
@@ -0,0 +1,520 @@
...
+    ///
+    /// assert_eq!(span.start(), 0);
+    /// assert_eq!(span.end(), 2);
+    /// ```
+    #[inline]
+    pub fn span(self, other: Position<I>) -> span::Span<I> {
+        if &*self.input as *const I == &*other.input as *const I {
+            span::new(self.input, self.pos, other.pos)
+        } else {
+            panic!("Span created from positions from different inputs")
```
Keats: no way to avoid that `panic`? Is this a possible outcome?

pest/src/inputs_private/string_input.rs
```diff
@@ -0,0 +1,305 @@
...
+                Some('\r') => {
+                    if let Some(&'\n') = chars.peek() {
+                        chars.next();
+
+                        if pos == 1 {
+                            pos -= 1;
+                            line_col = (line_col.0 + 1, 1);
+                        } else {
+                            pos -= 2;
+                            line_col = (line_col.0 + 1, 1);
```
Keats: you can move that line below the if/else since it's the same in both branches

pest/src/iterators_private/mod.rs
```diff
@@ -5,10 +5,12 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
```
Keats: I think you can start using `pub(restricted)`, it should be stable soon

pest/src/iterators_private/pairs.rs
```diff
@@ -0,0 +1,108 @@
...
+    queue: Rc<Vec<QueueableToken<R>>>,
+    input: Rc<I>,
+    start: usize,
+    end: usize
+) -> Pairs<R, I> {
+    Pairs {
+        queue: queue,
+        input: input,
+        start: start,
+        end: end
```
Keats: I believe field init shorthand was released in 1.17 so you can do `Pairs { queue, input, start, end }`

pest/src/iterators_private/token_iterator.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+impl<R: RuleType, I: Input> Iterator for TokenIterator<R, I> {
+    type Item = Token<R, I>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index == self.end {
+            None
+        } else {
```
Keats: you can remove the `else` indentation by returning the `None` above

pest/src/iterators_private/pairs.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl<R: RuleType, I: Input> Iterator for Pairs<R, I> {
+    type Item = Pair<R, I>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.start >= self.end {
+            None
+        } else {
```
Keats: you can remove the else indentation by returning the None above

pest/src/error.rs
dragostis: I guess "unknown" is a bit clearer as a general error. The reason I used "inexplicit" was to encourage shaping the grammar in a way that would make the error explicit.

pest/src/inputs_private/position.rs
dragostis: I would argue that having multiple `Input`s and actually hitting this panic is a rare event which is why I'd keep the syntax simpler here and not wrap in an `Option` or `Result`. What do you think?

pest/src/inputs_private/string_input.rs
dragostis: The whole method is a bit ugly. I've invested some time into rethinking it, but gave up. I'm open to other designs here.

dragostis:
@Keats Thank you so much for the review! 😃 You can try the example to get an idea of how things will run once this is merged. Still gotta port the precedence climber from 0.4 and then continue work on the derive.

pest/src/inputs_private/position.rs
Keats: Yeah, my point was more on how can you end up calling that method with 2 different inputs, I'm all for not having a `Result/Option`!

Keats:
Let me know when #103 can be done!

PR #114, opened 2017-05-12
Fixed links to Roadmap issues

emoon:
* Fixed links to issues to point to the main repo (these will be wrong if one forks the repo and they don't work as is also)
* Removed a trailing white space on one line.


dragostis:
Thank you. :)

PR #117, opened 2017-07-15
Add a parse_str function to the Parser trait

Keats:
And use it in `parses_to!`

### DIFF
--- a/pest/src/parser.rs
+++ b/pest/src/parser.rs
@@ -8,7 +8,7 @@
 use std::rc::Rc;
 
 use super::error::Error;
-use super::inputs::Input;
+use super::inputs::{Input, StringInput};
 use super::iterators::Pairs;
 use super::RuleType;
 
@@ -16,4 +16,6 @@ use super::RuleType;
 pub trait Parser<R: RuleType> {
     /// Parses `input` starting from `rule`.
     fn parse<I: Input>(rule: R, input: Rc<I>) -> Result<Pairs<R, I>, Error<R, I>>;
+    /// Parses an `input` string starting from `rule`.
+    fn parse_str(rule: R, input: &str) -> Result<Pairs<R, StringInput>, Error<R, StringInput>>;
 }
--- a/pest/src/parser.rs
+++ b/pest/src/parser.rs
@@ -16,6 +16,8 @@ use super::RuleType;
 pub trait Parser<R: RuleType> {
     /// Parses `input` starting from `rule`.
     fn parse<I: Input>(rule: R, input: Rc<I>) -> Result<Pairs<R, I>, Error<R, I>>;
-    /// Parses an `input` string starting from `rule`.
-    fn parse_str(rule: R, input: &str) -> Result<Pairs<R, StringInput>, Error<R, StringInput>>;
+    /// Parses an `input` &str starting from `rule`.
+    fn parse_str(rule: R, input: &str) -> Result<Pairs<R, StringInput>, Error<R, StringInput>> {
+        Self::parse(rule, Rc::new(StringInput::new(input.to_owned())))
+    }
 }
--- a/pest_derive/Cargo.toml
+++ b/pest_derive/Cargo.toml
@@ -17,7 +17,8 @@ proc-macro = true
 [dependencies]
 quote = "^0.3"
 syn = "^0.10"
-pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
+# pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
+pest = { path = "../pest" }
 
 [badges]
 travis-ci = { repository = "pest-parser/pest" }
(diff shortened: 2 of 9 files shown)


pest/src/parser.rs
```diff
@@ -8,12 +8,14 @@
...
+use super::inputs::{Input, StringInput};
 use super::iterators::Pairs;
 use super::RuleType;
 
 /// A `trait` that defines a `Parser`.
 pub trait Parser<R: RuleType> {
     /// Parses `input` starting from `rule`.
     fn parse<I: Input>(rule: R, input: Rc<I>) -> Result<Pairs<R, I>, Error<R, I>>;
+    /// Parses an `input` string starting from `rule`.
+    fn parse_str(rule: R, input: &str) -> Result<Pairs<R, StringInput>, Error<R, StringInput>>;
```
dragostis: I would add the full implementation in `Parser` instead of keeping it abstract. This means that all manual implementations can be removed.

pest_derive/Cargo.toml
```diff
@@ -17,7 +17,8 @@ proc-macro = true
 [dependencies]
 quote = "^0.3"
 syn = "^0.10"
-pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
+# pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
```
dragostis: The git dependencies should be kept in order make this easy to depend on. Paths can be use when working locally until we publish on crates.io.

pest_grammars/Cargo.toml
```diff
@@ -11,8 +11,10 @@ license = "MPL-2.0"
 readme = "../README.md"
 
 [dependencies]
-pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
-pest_derive = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
+#pest = { git = "https://github.com/pest-parser/pest", branch = "procedural" }
```
dragostis: Same here.

pest/src/parser.rs
```diff
@@ -8,12 +8,14 @@
...
 
 use super::error::Error;
-use super::inputs::Input;
+use super::inputs::{Input, StringInput};
 use super::iterators::Pairs;
 use super::RuleType;
 
 /// A `trait` that defines a `Parser`.
 pub trait Parser<R: RuleType> {
     /// Parses `input` starting from `rule`.
```
dragostis: This should probably say ``Parses `Input` ``.

pest/src/parser.rs
```diff
@@ -8,12 +8,14 @@
...
     fn parse<I: Input>(rule: R, input: Rc<I>) -> Result<Pairs<R, I>, Error<R, I>>;
+    /// Parses an `input` string starting from `rule`.
```
dragostis: This should specify that we're parsing a `&str`. It should also mention that it's a convenience method.

pest/src/parser.rs
Keats: Would it show on the docs though?

pest_derive/Cargo.toml
Keats: I wouldn't be able to build it if I wasn't using local paths though

pest/src/parser.rs
Keats: @dragostis if I implement it in the macro directly, this wouldn't show at all on pest docs would it?

dragostis: If you implement it in the trait, it will show up.

pest_derive/Cargo.toml
dragostis: Of course. Just make sure you keep the local paths local only without pushing it to the branch since then it would make depending on the git branch in Cargo impossible. (For those hacking on the alpha)

Keats: It wouldn't build on Travis as well too though, is that ok? I guess it's the issue of depending on a git branch

pest/src/parser.rs
Keats: You can't implement it in the trait if it's not part of the trait though. Do you mean a `impl MyParser { pub fn parse_str(..) {...} }` in generator.rs, below the `impl pest::Parser`?

dragostis: It should be part of the trait.

pest_derive/Cargo.toml
dragostis: @Keats Oh, true. I think it's fine to merge on procedural if fails on Travis and then see if it passes there.

pest/src/parser.rs
Keats: So I guess I don't understand what you mean by "I would add the full implementation in Parser instead of keeping it abstract." if it should be part of the trait.

dragostis: @Keats Here's an [example](https://gist.github.com/anonymous/b0d510950de46676980059cac5596d9a#file-playground-rs-L31).

Keats: And all this time I somehow thought traits definition couldn't have actual implementation for some reason :O

Keats:
@dragostis updated and removed the local deps in Cargo.toml
I would keep them commented out though as you need to use them whenever you develop anyway so that saves some time

PR #127, opened 2017-08-19
Clarify the example in the README.

bheisler:
I've reworded some parts of the README and expanded the example to clarify what a token pair is and how they can be used.

This is inspired by [this reddit comment](https://www.reddit.com/r/rust/comments/6uho7c/pest_10_beta_procedural_macros_simplified_api/dltdzzt/) asking for a clarification on this. I had the same confusion when I first saw this project, so hopefully this helps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ pest is a [PEG](https://en.wikipedia.org/wiki/Parsing_expression_grammar) parser
 
 Defining a grammar for a list of alpha-numeric identifiers that do not start with a digit is a straight-forward as:
 
-```
+```rust
 alpha = { 'a'..'z' | 'A'..'Z' }
 digit = { '0'..'9' }
 
@@ -47,16 +47,16 @@ use pest::Parser;
 struct IdentParser;
 
 fn main() {
-    let pairs = IdentParser::parse_str(Rule::ident_list, "abc123 def456").unwrap_or_else(|e| panic!("{}", e));
+    let pairs = IdentParser::parse_str(Rule::ident_list, "a1 b2").unwrap_or_else(|e| panic!("{}", e));
 
-    //Because ident_list is silent, the iterator will contain idents
+    // Because ident_list is silent, the iterator will contain idents
     for pair in pairs {
-        //A pair is a combination of the rule which matched and a span of input
+        // A pair is a combination of the rule which matched and a span of input
         println!("Rule: {:?}", pair.as_rule());
         println!("Span: {:?}", pair.clone().into_span());
         println!("Text: {}", pair.clone().into_span().as_str());
 
-        //A pair can be converted to an iterator of the tokens which make it up:
+        // A pair can be converted to an iterator of the tokens which make it up:
         for inner_pair in pair.into_inner() {
             match inner_pair.as_rule() {
                 Rule::alpha => println!("letter: {}", inner_pair.into_span().as_str()),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,24 +15,25 @@ pest is a [PEG](https://en.wikipedia.org/wiki/Parsing_expression_grammar) parser
 
 ## Elegant grammar
 
-Defining a grammar for an alpha-numeric identifier that does not start with a digit is a straight-forward as:
+Defining a grammar for a list of alpha-numeric identifiers that do not start with a digit is a straight-forward as:
 
-```rust
+```
```
dragostis: I would leave the `rust` tag here in order to have *some* highlighting, even though it's not technically correct.

README.md
```diff
@@ -46,29 +47,60 @@ use pest::Parser;
 struct IdentParser;
 
 fn main() {
-    let pairs = IdentParser::parse_str(Rule::ident, "abc123").unwrap_or_else(|e| panic!("{}", e));
+    let pairs = IdentParser::parse_str(Rule::ident_list, "abc123 def456").unwrap_or_else(|e| panic!("{}", e));
 
+    //Because ident_list is silent, the iterator will contain idents
```
dragostis: Papercut here, but I think a space after the `//` make the comment more elegant.

README.md
```diff
@@ -46,29 +47,60 @@ use pest::Parser;
...
     for pair in pairs {
-        match pair.as_rule() {
-            Rule::alpha => println!("letter: {}", pair.into_span().as_str()),
-            Rule::digit => println!("digit: {}", pair.into_span().as_str()),
-            _ => unreachable!() // ident rule is silent and cannot be reached
-        };
+        //A pair is a combination of the rule which matched and a span of input
```
dragostis: Same here.

README.md
```diff
@@ -46,29 +47,60 @@ use pest::Parser;
...
+        println!("Rule: {:?}", pair.as_rule());
+        println!("Span: {:?}", pair.clone().into_span());
+        println!("Text: {}", pair.clone().into_span().as_str());
+
+        //A pair can be converted to an iterator of the tokens which make it up:
```
dragostis: And here.

README.md
```diff
@@ -46,29 +47,60 @@ use pest::Parser;
...
+                _ => unreachable!()
+            };
+        }
     }
 }
 ```
 
+This produces the following output:
+```
+Rule: ident
```
dragostis: This is a bit long. How about making the input `"a1 b2"`?

README.md
```diff
@@ -46,29 +47,60 @@ use pest::Parser;
...
+letter: f
+digit: 4
+digit: 5
+digit: 6
+```
+
 ## Meaningful error reporting
 
-Parsing `"123"` instead of `"abc123"` in the code above will result in the following panic:
+Parsing `"123"` instead of `"abc123 def456"` in the code above will result in the following panic:
```
dragostis: This will also need to be changed to `"a1 b2"`.

dragostis: This particular example is a little unfortunate because it no longer highlights the "unexpected" error report. Maybe it would be more interesting if only the first ident was not allowed to start with a number? Then we can have one example where we parse `"123"` which errors with "unexpected digit" and one with `"abc *"` which errors with `"expected ident"`.

bheisler: I did some experimenting with the grammar and test string and wasn't able to cause it to produce the errors you describe. With a separate rule for the first and other idents, the string "123" fails with "expected first_ident", while "abc *" parses out "abc" and then stops.

dragostis: This seems to work just fine for me:

```
alpha = { 'a'..'z' | 'A'..'Z' }
digit = { '0'..'9' }

ident = { (alpha | digit)+ }

ident_list = _{ !digit ~ ident ~ ( " " ~ ident )+ }
          // ^
          // ident_list rule is silent which means it produces no tokens
```

```
thread 'main' panicked at ' --> 1:1
  |
1 | 1234
  | ^---
  |
  = unexpected digit', src/main.rs:12:84
```

```
thread 'main' panicked at ' --> 1:4
  |
1 | ab *
  |    ^---
  |
  = expected ident', src/main.rs:12:84
```

README.md
```diff
@@ -15,24 +15,25 @@ pest is a [PEG](https://en.wikipedia.org/wiki/Parsing_expression_grammar) parser
...
+```
 alpha = { 'a'..'z' | 'A'..'Z' }
 digit = { '0'..'9' }
 
-ident = _{ !digit ~ (alpha | digit)+ }
-     // ^
-     // ident rule is silent which means it produces no tokens
+ident = { !digit ~ (alpha | digit)+ }
+
+ident_list = _{ ident ~ ( " " ~ ident )+ }
```
dragostis: I wouldn't use spaces around the interior of enclosing parens.

dragostis:
I decided to merge this PR as is and to do the small changes myself in the interest of time, since there's still some traffic going on today.

PR #143, opened 2017-08-28
2x smaller SVG logo

RazrFalcon:
Mostly lossless SVG optimization. 10.9KiB -> 4.7KiB

Diff:

diff


dragostis:
Sweet! Thanks for the PR.

PR #146, opened 2017-09-02
Allow StringInput from borrowed str, to avoid copy

### LINKED ISSUE #141: Lifetime issue with span.as_str()
I have a function that takes a `&str` and I would like to get a vector of parsed str segments, say `Vec<&str>`. 

I'm using pest parser API `parse_str` to get `Pairs`, convert it to `FlatPairs` and filter those rules to get ones I'm interested in. To get the actual string slice, I convert `Pair` to `Span` by `into_span()` and then push `span.as_str()` to the output vector.

The issue is `&str` generated from `span.as_str()` cannot live out of the scope of `span`.  So it's impossible to be returned from the function. 

From my understanding, `as_str()` should return a slice that shares same lifetime with the `&str` passed into `parsed_str`. However, it has the same lifetime with `Span` currently.

(For now we can use the start/end position to get slice from original `&str` though)

sunng87:
~~Fixes 141~~ 

Added support for `StringInput` to hold a borrowed `str`, as well as `String`. This is to avoid String copy in functions like `parse_str`.

### DIFF
--- a/pest/src/inputs/span.rs
+++ b/pest/src/inputs/span.rs
@@ -19,19 +19,15 @@ use super::position;
 pub struct Span<I: Input> {
     input: Rc<I>,
     start: usize,
-    end: usize
+    end: usize,
 }
 
 #[inline]
 pub fn new<I: Input>(input: Rc<I>, start: usize, end: usize) -> Span<I> {
-    Span {
-        input,
-        start,
-        end
-    }
+    Span { input, start, end }
 }
 
-impl<I: Input> Span<I> {
+impl<'a, I: Input> Span<I> {
     /// Returns the `Span`'s start position as a `usize`.
     ///
     /// # Examples
@@ -148,7 +144,7 @@ impl<I: Input> Span<I> {
     /// assert_eq!(span.as_str(), "b");
     /// ```
     #[inline]
-    pub fn as_str(&self) -> &str {
+    pub fn as_str(&'a self) -> &'a str {
         unsafe { self.input.slice(self.start, self.end) }
     }
 }
@@ -169,9 +165,7 @@ impl<I: Input> Clone for Span<I> {
 
 impl<I: Input> PartialEq for Span<I> {
     fn eq(&self, other: &Span<I>) -> bool {
-        Rc::ptr_eq(&self.input, &other.input) &&
-        self.start == other.start &&
-        self.end == other.end
+        Rc::ptr_eq(&self.input, &other.input) && self.start == other.start && self.end == other.end
     }
 }
 
@@ -182,7 +176,7 @@ impl<I: Input> PartialOrd for Span<I> {
         if Rc::ptr_eq(&self.input, &other.input) {
             match self.start.partial_cmp(&other.start) {
                 Some(Ordering::Equal) => self.end.partial_cmp(&other.end),
-                ordering => ordering
+                ordering => ordering,
             }
         } else {
             None
@@ -192,7 +186,9 @@ impl<I: Input> PartialOrd for Span<I> {
 
 impl<I: Input> Ord for Span<I> {
     fn cmp(&self, other: &Span<I>) -> Ordering {
-        self.partial_cmp(other).expect("cannot compare spans from different inputs")
+        self.partial_cmp(other).expect(
+            "cannot compare spans from different inputs",
+        )
     }
 }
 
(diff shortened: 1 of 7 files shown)


dragostis:
The problem with this is that it renders the `Rc` used everywhere in the API pretty much useless. (in `Position`, `Span`, `Pair` and `Pairs`) Maybe the right approach would be to replace the `Rc` with lifetimes? The only reason why `Rc`s were used in the first place was to keep the API simple.

sunng87:
Yes, for correction, this patch doesn't fix issue mentioned in #141 . It's just to avoid allocating string when we use `parse_str`.

I had an attempt to replace `Rc<Input>` to `&'a Input`, it requires huge amount of keyboard work 😂 (I have got pest module compiles, but it still requires work for test and derive module)
For now, I'm OK with those Rc clones.

dragostis:
Then I propose a 2 part solution:

* implement `StrInput` which works with `&str`
* abstract over `StringInput` because they have the same implementation

After this is done, it may be worth researching whether replacing `Rc`s with borrows does not make the API extremely hard to use.

sunng87:
Just added `StrInput` and removed unneeded lifetime indicators in `StringInput` and `FileInput`.

By the way, I'm using rustfmt with my editor so some files are formatted when I'm saving them. I recommend you to use rustfmt to keep the format consistent too. :-)

dragostis:
@sunng87 rustfmt is still not mature enough. I have a local `rustfmt.toml` file, but there are a few pending bugs in rustfmt that keep this from going anywhere.

pest/src/inputs/string_input.rs
```diff
@@ -143,7 +142,7 @@ unsafe fn line_of(source: &str, mut pos: usize) -> &str {
             |&(_,
```
dragostis: This is one of the issues with rustfmt. 😛

pest/src/inputs/string_input.rs
```diff
@@ -30,10 +36,182 @@ impl StringInput {
...
+            }
+            Some('\n') => {
+                pos -= 1;
+                line_col = (line_col.0 + 1, 1);
+            }
+            Some(c) => {
+                pos -= c.len_utf8();
+                line_col = (line_col.0, line_col.1 + 1);
+            }
+            None => unreachable!(),
```
dragostis: There should be no comma here. Hm, newest version ofrustfmt should have dealt with this particular case.

pest/src/inputs/string_input.rs
```diff
@@ -30,10 +36,182 @@ impl StringInput {
...
+        if end > 0 && source.slice_unchecked(end - 1, end) == "\n" {
+            end -= 1;
+        }
+        if end > 0 && source.slice_unchecked(end - 1, end) == "\r" {
+            end -= 1;
+        }
+
+        end
+    } else {
+        let end = source.char_indices().skip_while(|&(i, _)| i < pos).find(
```
dragostis: This should be formatted like the one above.

sunng87: @dragostis I'll suggest you to put rustfmt.toml under the root directory of this project. So contributors get same rustfmt configuration and we can keep the code consistent. 

And for these formation issues, we can leave it for rustfmt iterations in future

dragostis: I really considered that, but formatting can become really poor in some particular cases, like the one above, and it's not even consistent with the prior example. Putting the file there pretty much means keeping it. Right now, using rustfmt on the whole project makes the formatter hang because of an issues with nested lambdas. I am very much in favor of going down this road, but right now I would say it's partly too much of a hassle, and partly making the code messy like the example above.

sunng87: I see. I had issue with saving `pest/test/json.rs` and thought it was issue with Emacs' rust-mode.

pest/src/inputs/span.rs
```diff
@@ -182,7 +176,7 @@ impl<I: Input> PartialOrd for Span<I> {
         if Rc::ptr_eq(&self.input, &other.input) {
             match self.start.partial_cmp(&other.start) {
                 Some(Ordering::Equal) => self.end.partial_cmp(&other.end),
-                ordering => ordering
+                ordering => ordering,
```
dragostis: One little comma left. (updating rustfmt to latest master should fix it)

sunng87:
I was using rustftm from crate.io, will update to master. 
Disabled format-on-save in emacs and manually formatted those lines.

Issue #158, opened 2017-09-24
Factor out `.pest` parser for use by other tools


Victor-Savu:
I was wondering if it would make sense to factor out the part of the code which (bootstrappingly?) parses the `.pest` file so that other tools such as syntax highlighters and completion engines could use it.

dragostis:
I very much agree. When I first started working on this, I was not aware that `proc_macro` crates had some limitations. In order to get highlighting and completion, one can use `intellij-rust`, but clippy seems to completely ignore this, so it makes sense to do this.

Would you be willing to give this a try?

dragostis:
Actually, factoring this out in a separate crate would also mean making a lot of APIs which are not yet documented public, which means that there will be a bit more work to do.

Victor-Savu:
Ohh right that kind of completion  :) I was referring to to highlighting and completion of the grammar specification language used in the `.pest` file.

Victor-Savu:
This should be a really nice self-contained project, and I'd be happy to have a go at it.

dragostis:
Oh, oh! That too, of course. Hm. It would be great if there was a way to bootstrap pest in order to have the grammar written in pest form, but I don't think that's actually possible.

Yes, I think it can work as a separate project and I'm willing to mentor and help out. :)

Victor-Savu:
Yep, the bootstrapping is that I am working on now :) The past 3 hours have shown that it is not exactly straightforward, but bootstrapping never is. I'll just try and see what comes out.

Victor-Savu:
[WIP comment - I just don't want the art work below to be accidentally lost]

This is great! I would really like to learn the code, philosophy and in general what your vision is regarding the roles of the different modules before committing to a particular architecture. I hope you don't mind if I ask some silly questions along the way.

From your suggestion, I understand that `pest_derive::{parser, ast, validator, optimizer, generate}` all deal with "compiling" the `pest` language:
```
      [FileInput]
           |
           V
(parser + ast + validator) --> [panic!]
           |
    [Vec<ast::Rule>]
           |
           V
      (optimizer)
           |
           V
    [Vec<ast::Rule>]
           |
           V
      (generator)
           |
    [quote::Tokens]
```
If my ASCII doodle is accurate, my initial thoughts would be:
- The `parser + ast + validator` bundle hold the most value for external use. One immediate usability improvement would be if the `validator` returned the errors instead of panicking, leaving the external application to decide on the appropriate reaction to the error.
[...]

Issue #148, opened 2017-09-03
Add rustfmt.toml


dragostis:
Add [this rustfmt.toml](https://gist.github.com/dragostis/476ab6501423d807e4c58a2b01badc21) to the project. Currently blocked by https://github.com/rust-lang-nursery/rustfmt/issues/1265 and https://github.com/rust-lang-nursery/rustfmt/issues/1762.

PR #159, opened 2017-10-08
Extract the `parser` and `validator` from the `pest_derive` crate into `pest_meta`

Victor-Savu:
This is the first step towards #158 . It provides the new crate `pest_meta`, which grants access to the implementations of the parser and validator used by `pest_derive`. However, in order to declare them usable, there needs to be some experimentation outside their original intended use.

This PR is a WIP mainly to get early feedback and to prevent working inside the proverbial cave for too long.

Open questions:
- [x] I am not sure how versioning should work. I just put the same version for `pest_meta` as for all the other crates in the `pest` workspace but I am not sure if that's what I was supposed to do :smiley: 

Once the design & implementation kinks are ironed out and any new ideas are implemented (or postponed, dropped), this PR still needs:
- [ ] documentation for `pest_meta` targeted at people who want to use it for building tools for `pest` (syntax highlighting, linting, completion, etc.)


Victor-Savu:
Just a note on bootstrapping:

Pest itself uses a language for defining the grammar rules for generating a parser. A parser for said language could even be generated by Pest from a description of the language, written in the Pest language itself. That parser could, in turn, be used to implement a new version of Pest. This process is called bootstrapping.

The problem is that the new version  of Pest would depend on the old version of Pest, both of which would need to be compiled. Assuming this worked (which is still unclear at the moment), the users of the pest library would have to incur at least the extra compilation time required for the bootstrapping operation. Unless there can be a way around this (e.g. cargo starts shipping binary artifacts), the benefits of bootstrapping would need to outweigh the compilation time costs.

Another problem is that the parser for the new version of the language would be generated from a description written in an older version of the language. This means that the language itself would need versioning in order to avoid confusion and incompatibilities.

In general, the benefits of bootstrapping the Pest language are:
- The language proves it is usable in non-trivial scenarios. [There are arguably more complex languages which can be used to prove the prowess of a grammar definition language]
- Contributors to the language are likely users of the language, so they may be more familiar with the language.

dragostis:
I guess bootstrapping can currently stay in its current form. The reason bootstrapping might make more sense in the future would if pest will gain error-flexible parsing capabilities.

pest_meta/src/validator.rs
```diff
@@ -0,0 +1,535 @@
...
+
+    let definitions: HashSet<_> = definitions.iter().map(|span| span.as_str()).collect();
+    let called_rules: HashSet<_> = called_rules.iter().map(|span| span.as_str()).collect();
+
+    let defaults = called_rules.difference(&definitions);
+
+    Ok(defaults.into_iter().map(|string| Ident::new(*string)).collect())
+}
+
+pub fn validate_rust_keywords<'a, I: Input+'a, Definitions: IntoIterator<Item=&'a Span<I>>>(
```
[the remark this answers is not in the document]
Victor-Savu: Sure thing! I tried [rustfmt](https://github.com/rust-lang-nursery/rustfmt), but had to revert because it had some very different opinions :smile: and I really wanted to reproduce the style of the other crates as much as possible. Is there any plan to use it at some point when it is stable (and consistent) enough?

pest_derive/src/lib.rs
```diff
@@ -268,39 +267,46 @@ pub fn derive_parser(input: TokenStream) -> TokenStream {
...
+                PestRule::single_quote => "`'`".to_owned(),
                 other_rule => format!("{:?}", other_rule)
             }
         }))
     };
 
-    let (ast, defaults) = parser::consume_rules(pairs);
+    fn format_errors<Errors: IntoIterator>(errors: Errors) -> String where Errors::Item: Display {
+        errors.into_iter()
+            .map(|error| { format!("{}", error) })
```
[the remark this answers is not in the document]
Victor-Savu: Yep, I tried to do that as well, but I think I missed this one :) I really need to get Vim to do it for me :)

pest_meta/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+repository = "https://github.com/pest-parser/pest"
+documentation = "https://docs.rs/pest"
+keywords = ["pest", "parser", "peg", "grammar"]
+categories = ["parsing"]
+license = "MPL-2.0"
+readme = "_README.md"
+
+[dependencies]
+quote = "^0.3"
+maplit = "^0.1.5"
```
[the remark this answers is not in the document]
Victor-Savu: `maplit` [is just a bunch of macros](https://github.com/bluss/maplit/blob/master/src/lib.rs), so it doesn't really take any compilation time at all aside from downloading it (<200 lines of code including tests and documentation).

I thought you meant "quote", which is quite a bit of an annoyance since we only use one struct: `quote::Ident`. Of course, `quote` is already a dependency of `pest_derive`, so for that use case there's no added cost, but for projects who don't care much about producing a rust plugin it may be annoying.

pest_meta/src/validator.rs
```diff
@@ -0,0 +1,535 @@
...
+use pest::Error;
+use pest::inputs::{Input, Span};
+use pest::iterators::Pairs;
+
+use quote::Ident;
+
+use super::parser::{PestRule, ParserExpr, ParserNode, ParserRule};
+
+pub fn validate_pairs<I: Input>(pairs: Pairs<PestRule, I>) -> Result<Vec<Ident>, Vec<Error<PestRule, I>>> {
+    let rust_keywords = hashset!{
```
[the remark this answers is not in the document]
Victor-Savu: Sure! But I would really hope you gave [maplit](https://github.com/bluss/maplit/blob/master/src/lib.rs) another thought. The code for this macro (16 lines) is more than 25% of the library code (16 + 16 + 13 + 14 lines).

pest_meta/src/parser.rs
```diff
@@ -57,19 +55,19 @@ pub enum GrammarRule {
...
         input: Rc<I>
-    ) -> Result<Pairs<GrammarRule, I>, Error<GrammarRule, I>> {
+    ) -> Result<Pairs<PestRule, I>, Error<PestRule, I>> {
         fn grammar_rules<I: Input>(
             pos: Position<I>,
-            state: &mut ParserState<GrammarRule, I>
+            state: &mut ParserState<PestRule, I>
         ) -> Result<Position<I>, Position<I>> {
             pos.sequence(|pos| {
-                soi(pos, state).and_then(|pos| {
```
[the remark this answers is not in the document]
Victor-Savu: I was actually going to ask about this one, because I wasn't sure if it was there for historical reasons or if it was written in expectation of an upcoming feature. Is it OK that I took it out?

pest_meta/Cargo.toml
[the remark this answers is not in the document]
dragostis: I meant both, actually. The only reason I'd give up on `maplit` is because it's easy to replace and one request less to worry about when compiling big projects with a lot of dependencies.

About `quote`, it may be a smart idea to give up on using `quote::Ident` in `pest_meta` since it doesn't make sense at all.

pest_meta/src/parser.rs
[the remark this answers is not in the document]
dragostis: Yup!

pest_meta/src/validator.rs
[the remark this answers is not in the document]
dragostis: Sure. I guess this is a perfect use case and there is pretty much no other dependency.

pest_meta/src/validator.rs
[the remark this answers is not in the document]
dragostis: The issue with `rustfmt` is an [old one](https://github.com/pest-parser/pest/issues/148).

pest_meta/src/parser.rs
```diff
@@ -794,44 +778,42 @@ impl Parser<GrammarRule> for GrammarParser {
...
-                GrammarRule::compound_atomic_modifier => compound_atomic_modifier(pos, &mut state),
-                GrammarRule::non_atomic_modifier => non_atomic_modifier(pos, &mut state),
-                GrammarRule::opening_brace => opening_brace(pos, &mut state),
-                GrammarRule::closing_brace => closing_brace(pos, &mut state),
-                GrammarRule::opening_paren => opening_paren(pos, &mut state),
-                GrammarRule::closing_paren => closing_paren(pos, &mut state),
-                GrammarRule::expression => expression(pos, &mut state),
-                GrammarRule::term => term(pos, &mut state),
-                GrammarRule::positive_predicate_operator => {
+                PestRule::grammar_rules => grammar_rules(pos, &mut state),
```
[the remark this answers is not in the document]
Victor-Savu: Yeah, that is a lot of repetition a lot of repetition.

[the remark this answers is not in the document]
dragostis: This comment is so meta. haha.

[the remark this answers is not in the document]
Victor-Savu: New macro pushed :)

pest_meta/src/validator.rs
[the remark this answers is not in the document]
Victor-Savu: Ah, whoops!  it was right there! Good thing rustfmt is open source.

pest_meta/src/parser.rs
```diff
@@ -776,44 +776,48 @@ impl Parser<PestRule> for PestParser {
...
+        macro_rules! parse_rule {
+            ($pos:ident , $state:ident for $rule:ident in { $( $case:ident ),* }) => {
+                match $rule {
+                    $(PestRule::$case => $case($pos, &mut $state)),*
+                }
+            }
+        }
+
+        pest::state(input, move |mut state, pos| parse_rule!{
+            pos, state for rule in {
```
dragostis: Nice, creative syntax! 😃

pest_meta/src/validator.rs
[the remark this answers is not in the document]
dragostis: Yes, but, unfortunately, the hang bug seems to be pretty non-trivial.

pest_meta/Cargo.toml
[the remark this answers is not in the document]
dragostis: I would add the circumvention of the `quote` crate as a second to do.

pest_meta/src/parser.rs
Victor-Savu: Yeah, I like rust macros! They can allow for some really intuitive and readable code.

Victor-Savu: Come to think about it, maybe something with `match` in the syntax might be more appropriate, as the rule is only matched to exactly one of the variants.

dragostis:
The `match` syntax does look a bit better.

Victor-Savu:
@dragostis any idea what I did to bring the code coverage down so much? I think it might be due to all of the text I removed when I made the validator return Result instead of panicking in a2a2980. I guess those cases tested both the validator and the parser at once.

dragostis:
@Victor-Savu It seems to be the case, yes. The integration tests in `pest_derive` don't seem to produce coverage for `pest_meta`. The ideal solution would be to have a way to get the coverage back instead of writing extra tests. Maybe there's a way to fix it in `travis.yaml`.

dragostis:
@Victor-Savu [tarpaulin](https://github.com/xd009642/tarpaulin) might help. I don't have a machine to test it on right now, so maybe you could try `cargo tarpaulin -v -p pest pest_derive pest_grammars` on your banch?

dragostis:
So, tarpaulin seems to work, but there's an issue with file paths that causes some tests to fail. I've already opened an [issue](https://github.com/xd009642/tarpaulin/issues/57).

Victor-Savu:
Wow! Tarpaulin is awesome!

Victor-Savu:
Does the example tick the documentation box?

Victor-Savu:
Hi, @dragostis! I am not sure what would be left to do for this PR and I'd like to do hear your thoughts.

pest_meta/Cargo.toml
```diff
@@ -12,3 +12,9 @@ readme = "_README.md"
 [dependencies]
 maplit = "^0.1.5"
 pest = { path = "../pest", version = "^1.0.0-beta" }
+
+[[example]]
+name="colorize"
```
dragostis: I'm not completely sure, but the example entry seems to be redundant.

pest_meta/examples/colorize.rs
```diff
@@ -0,0 +1,76 @@
...
+    // If we limit our tool to not do any nested styles (e.g. no italicized word within a bold
+    // sentence), then we can simply select a subsequence of non-overlapping elements from the
+    // flattened AST and print their content with the desired style. In between, we print the
+    // remaining text (straight from the input, without any special styling).
+    
+    // We keep track of the number of code points (characters) from the input that have been
+    // processed (styled), in order to:
+    // - skip sub-trees which overlap with already styled ones to avoid duplicating the text
+    // - print the input spans between the styled sub-trees
+    // .
```
dragostis: The period here seems a bit too much. 😛

pest_meta/examples/colorize.rs
```diff
@@ -0,0 +1,76 @@
...
+            PestRule::identifier => <&str>::cyan,
+            PestRule::opening_brace | PestRule::closing_brace => <&str>::bold,
+            PestRule::character => <&str>::blue,
+            // and skip over the sub-trees that we don't really care about styling.
+            _ => continue
+        };
+        // Now, we print everything between the last processed code point and the beginning of the
+        // current sub-tree span using the default text output style. Then we print the span of the
+        // sub-tree itself using the selected style.
+        print!("{}{}", unsafe {input.slice(processed, span.start())}, style(span.as_str()));
```
dragostis: The rest of the code leave spaces around the unsafe `{}` braces. (still a muddy affair until rustfmt will work)

pest_meta/examples/colorize.rs
```diff
@@ -0,0 +1,76 @@
...
+        // We update the record of the last code point we processed.
+        processed = span.end();
+    }
+    // To finish up, we print whatever is left from the input using the default style.
+    println!("{}", unsafe {input.slice(processed, input.len())});
```
dragostis: Same here.

pest_meta/examples/colorize.rs
Victor-Savu: We are linguists -- of sorts :)

Victor-Savu:
@dragostis No worries! Thanks a lot for taking the time to review this! I applied the fixes (I hope :crossed_fingers:)

Victor-Savu:
Great! If you think the current interface looks good, I can get to the documentation.

I think documentation should be part of the PR. The example colorizer was the first step in that direction, but I feel there needs to be a way for people to find it. Should I add a small section in the top-level README.md? If that gets too long, maybe a dedicated `README.md` to replace the symbolic link in `pest_meta` and a back reference from the main one could work.

In the meantime, I will write the rust API documentation.

Victor-Savu:
Looks like documenting code makes one reflect on what they did :smile: Here is something I haven't considered before starting to document the code:

The rule struct is defined as:
```rust
pub struct Rule {
    pub name: String,
    pub ty: RuleType,
    pub expr: Expr
}
```
Now, going to the [documentation in pest_derive](https://docs.rs/pest_derive/1.0.0-beta.16/pest_derive/#rules), we find that:

> Rules have the following form:
> ```
> name = optional_modifier { expression }
> ```

This might be confusing for the user of `Rule` since instead of `modifier` they find `ty` which is a `RuleType` and it is not really optional. What they would expect would probably look like:
```rust
pub struct Rule {
    pub name: String,
    pub modifier: Option<Modifier>,
    pub expression: Expr,
}
```

This wasn't important before, but now that users of `pest_meta` can see the struct it becomes problematic.

Another issue (maybe related) is that stabilizing `pest_meta` at the same time as `pest` means that the interfaces would be frozen before people had a chance to try them all out. I now think that `pest_meta` should not be `beta` or `alpha`, but rather `0.1` to allow for non-trivial interface changes even after `pest` itself becomes `1.0`.

dragostis:
I fully agree on launching `pest_meta` as a `0.1` release; stability of the crate should not be a huge issue for now since the number of consumers would be quite limited, I would imagine.

The AST could definitely receive a bit of a cleanup. I agree with you on the `Rule` type and I would also add a point on the map functions on the `Expr` type. A visitor pattern might work a little bit better here. The maps are really nice for a very functionally pure kind of API that I like, but I guess is less elegant as a public API. What do you think?

dragostis:
I've thought a bit more about this, and I think it would be a neat thing to bootstrap pest completely. As such, version `1.0.0` would be published without `pest_meta` and will forever remain a branch on GitHub. As such, `pest_meta` will be able to use `1.0.X` and make use of the derive in order to parse its own grammar.

@Victor-Savu, what do you think about this idea?

Victor-Savu:
@dragostis Sounds good! So how would that look? Pest releases in its current form as `1.0`. `pest_meta` then uses `pest_derive 1.0.x` to implement a parser for the pest grammar definition language that can be used for tooling.

In this case, I would suggest closing this PR, as the implementation for the bootstrap would be pretty different than what it is right now.

Would it make sense then for `pest_meta` to live outside of the `pest` repository (and cargo workspace), the way you suggested in the beginning?

dragostis:
I don't think that closing this PR is a good idea. I like the general and bootstrapping the derive would only mean removing a lot of lines of code from the parser, but keeping everything else hopefully intact. Apart from that, this approach might lead to unacceptable build times, but it's too early to say right now, which I why I want to have some benchmarks running for compile times. With this in place, the compilation trace would be `pest_drive 1.0.* -> pest_meta -> pest_derive *`.

Victor-Savu:
@dragostis Ahh, I think I understand now. I investigated this a while back and ran into a problem: as far as I can tell, it seems that you cannot have `pest_derive` (transitively) depend on an earlier version of itself. This is why I postponed the bootstrapping idea back then.

dragostis:
I've managed to get it working by using some older version. One reason why I think that this might pay off is because it offer one extra incentive to lock pest's syntax.

dragostis:
@Victor-Savu Since 1.0 was just released, I guess it makes sense to focus on this PR next. Would you be willing to rebase and fix the conflicts? I can then help you out with the docs.

Victor-Savu:
Great to hear that! Congratulations!

I started the merge :) it is a bit tough (and I am a bit sleepy) but I am hopeful I will eventually nail it. I do need to be careful because I feel dangerously tempted to play with pest 1.0 instead :D

dragostis:
That's really good to hear! :D

With the advent of wasm, I think it's easier than ever to make the pest grammar inspection website and I have a few neat ideas about it.

dragostis:
A bit of help with the merge:

* `Position` and `Span` have been moved to the top mod
* `Input` has been removed and [replaced by lifetimes everywhere](https://github.com/pest-parser/pest/pull/172)

Victor-Savu:
Thanks! things made more sense after I read about the change `Input -> 'i`

Victor-Savu:
OK, merging this behemoth wasn't so bad after all. Another testament to Rust: if this library were written in C/C++, it would have been next to impossible to merge change which involved the lifetime dependency on the input and still hope the code is still sane.

dragostis:
> Another testament to Rust: if this library were written in C/C++, it would have been next to impossible to merge change which involved the lifetime dependency on the input and still hope the code is still sane.

This is pretty much the reason I decided to merge this significant PR so late in the launch of 1.0.

As a side note, I've managed to get some work done on the [website](https://github.com/pest-parser/site). It's pretty bare-bones right now as I'm trying to nail the design.

Victor-Savu:
I'm looking forward to the website and to the book.

The bootstrapping seems to be working, but I discovered I am not as amazing at pest (the language) as I was hoping :) I cannot get it to define the `comment` correctly. would you mind having a look?

dragostis:
Before continuing on this path, what's your general take in compilation speed? Adding this extra bootstrapping step produces even longer compilation times. It might be worth, for now, to simple annotate each function with its rule definition. Once cargo will actually cache crate binaries, it might be worth to make the complete change.

Victor-Savu:
I would be interested to know how much of a cost this actually is. The only big cost that I would expect (this is before getting a chance to measure) it the downloading & building of two versions of pest instead of just the one. I don't expect any change in compilation speed anywhere else, since this just affects how pest_derive `v1.0.4` would be built (instead of being hand-coded, it uses an older version of pest_derive). This assumes, as well, that plug-ins are not re-compiled every time the client code changes, which I hope is the case now.

This is why to me at least it seems worthwhile to implement it and then measure. That is, of course, if you feel the same. I was thinking about comparing the compilation times of the same "client" crate which in one scenario depends on `pest_derive = "1.0.4"` (the new `pest_derive` introduced by this change) and in the other scenario depends on `pest_derive = "1.0.3"` which is the version that `pest_derive v1.0.4` itself would depend on and is not bootstrapped. I would also compare re-compilation time differences between the two scenarios when code in the "client" crate changes, to simulate the experience of a developer of the "client" crate. What compilation time benchmarks would you be interested in?

dragostis:
I agree that it's worthwhile to give it a go. I just wanted to make sure that you feel the same, given the potential disadvantages. I guess the main concern is the debug build time. I've done some minor experiments and it doesn't seem to be such a drastic change, only 2x time during the derive phase. However, I'm not completely sure what other use cases might be and how it would affect those.

Also, I think that this change should come into play in 1.1. I've been working on some optimizations on `v1.1` for now and also renamed the local crate directories for better auto-completion in shells.

Victor-Savu:
Ah, very good point, the benchmark should include both debug and release builds.

Your point regarding when this change should land brought up a rather scary question to mind: Can the bootstrapped parser benefit from other advances in `pest_derive`, such as improvements in the optimization or code generation phase. In its current design, that would not be the case, so the parsing phase of any future `pest_derive` would only be as good as what `pest_derive v1.0.*` was able to produce. I find this thought disturbing. There is hope if at some point in the future cargo would stop building the entire dependency and would instead rely on pre-built signed versions of those dependencies.

dragostis:
It would be stuck in the 1.0 era, yes. But this might not be such a huge issue since the grammar should be stable from now on. However, this means that 1.0 will have to be supported indefinitely. As for the case of pre-built binaries, I don't think this will hit Cargo anytime soon.

Victor-Savu:
I finally bootstrapped the parser, but I am getting two errors that I don't really know how to fix. I suspect they have similar causes (at least in spirit).
Would you mind having a look?

pest_meta/src/grammar/pest.pest
```diff
@@ -0,0 +1,99 @@
+grammar_rules = _{ soi ~ grammar_rule* ~ eoi }
```
dragostis: This should be `grammar_rules = _{ soi ~ grammar_rule+ ~ eoi }`.

pest_derive/src/lib.rs
```diff
@@ -277,42 +272,42 @@ pub fn derive_parser(input: TokenStream) -> TokenStream {
...
-                GrammarRule::repeat_operator => "`*`".to_owned(),
-                GrammarRule::repeat_once_operator => "`+`".to_owned(),
-                GrammarRule::comma => "`,`".to_owned(),
-                GrammarRule::closing_paren => "`)`".to_owned(),
-                GrammarRule::quote => "`\"`".to_owned(),
-                GrammarRule::insensitive_string => "`^`".to_owned(),
-                GrammarRule::range_operator => "`..`".to_owned(),
-                GrammarRule::single_quote => "`'`".to_owned(),
+            error.renamed_rules(|rule| { match *rule {
+                Rule::grammar_rule => "rule".to_owned(),
```
dragostis: A translation rule for `_push` would also be needed.

pest_derive/src/optimizer.rs
```diff
@@ -137,14 +137,12 @@ pub fn optimize(rules: Vec<Rule>) -> Vec<Rule> {
...
-    use quote::Ident;
-
     #[test]
     fn concat_strings() {
         let rules = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Atomic,
+                name: "rule",
+                ty:   RuleType::Atomic,
```
dragostis: Space here is not needed per rustfmt.

pest_derive/src/optimizer.rs
```diff
@@ -159,8 +157,8 @@ mod tests {
         ];
         let concatenated = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Atomic,
+                name: "rule",
+                ty:   RuleType::Atomic,
```
dragostis: Same here.

pest_derive/src/optimizer.rs
```diff
@@ -285,8 +283,8 @@ mod tests {
     fn concat_insensitive_strings() {
         let rules = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Atomic,
+                name: "rule",
+                ty:   RuleType::Atomic,
```
dragostis: Space.

pest_derive/src/optimizer.rs
```diff
@@ -301,8 +299,8 @@ mod tests {
         ];
         let concatenated = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Atomic,
+                name: "rule",
+                ty:   RuleType::Atomic,
```
dragostis: Space.

pest_derive/src/optimizer.rs
```diff
@@ -314,37 +312,37 @@ mod tests {
     fn long_common_sequence() {
         let rules = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Silent,
+                name: "rule",
+                ty:   RuleType::Silent,
```
dragostis: Space.

pest_derive/src/optimizer.rs
```diff
@@ -314,37 +312,37 @@ mod tests {
...
                     ))
                 )
             },
         ];
         let optimized = vec![
             Rule {
-                name: Ident::new("rule"),
-                ty: RuleType::Silent,
+                name: "rule",
+                ty:   RuleType::Silent,
```
dragostis: Space.

pest_meta/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "pest_meta"
+version = "0.1.0"
+authors = ["Victor-Nicolae Savu <victor.nicolae.savu@gmail.com>"]
+repository = "https://github.com/pest-parser/pest"
+documentation = "https://docs.rs/pest"
+keywords = ["pest", "parser", "peg", "grammar"]
+categories = ["parsing"]
+license = "MPL-2.0"
```
dragostis: Should probably switch to Apache/MIT, like the rest of pest.

pest_meta/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+readme = "_README.md"
+
+[dependencies]
+maplit = "^0.1.5"
+pest_derive = "1.0.3"
+pest = "1.0.3"
+
+[dev_dependencies]
```
dragostis: This should be `dev-dependencies`. Not sure I'd keep this here though. It's kind of pointless to have every debug build download a copy of `colored`.

pest_meta/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+repository = "https://github.com/pest-parser/pest"
+documentation = "https://docs.rs/pest"
+keywords = ["pest", "parser", "peg", "grammar"]
+categories = ["parsing"]
+license = "MPL-2.0"
+readme = "_README.md"
+
+[dependencies]
+maplit = "^0.1.5"
+pest_derive = "1.0.3"
```
dragostis: pest dependencies here should be `^1.0`.

pest_meta/src/ast.rs
```diff
@@ -26,33 +24,33 @@ pub enum RuleType {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub enum Expr {
+pub enum Expr<'i> {
     Str(String),
```
dragostis: Since the lifetime was added, it probably makes sense to make `Str(&'i str)`, `Insens(&'i str)` and `Range(&'i str, &'i str)`.

pest_meta/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub mod parser;
+pub mod ast;
+pub mod validator;
+
+use std::fmt::Display;
+
+pub fn unwrap_or_report<T, E>(res: Result<T, E>) -> T where E: IntoIterator, E::Item: Display {
+    res.unwrap_or_else(|e| panic!("grammar error\n\n".to_owned() + &e.into_iter()
+              .map(|error| { format!("{}", error) })
```
dragostis: Not sure about this formatting. Maybe a newer version of rustfmt needed?

pest_meta/src/validator.rs
```diff
@@ -327,30 +311,30 @@ fn validate_whitespace_comment<'i>(rules: &Vec<ParserRule<'i>>) -> Vec<Error<'i,
...
     hash_map
 }
 
-fn left_recursion<'i>(rules: HashMap<Ident, &ParserNode<'i>>) -> Vec<Error<'i, GrammarRule>> {
-    fn check_expr<'i>(
-        node: &ParserNode<'i>,
-        rules: &HashMap<Ident, &ParserNode<'i>>,
-        trace: &mut Vec<Ident>
-    ) -> Option<Error<'i, GrammarRule>> {
+fn left_recursion<'a, 'i: 'a>(rules: HashMap<&'i str, &'a ParserNode<'i>>) -> Vec<Error<'i, Rule>> {
```
dragostis: Does `'a` need to be introduced here? I would have imagined that using `'i` everywhere would work.

pest_meta/Cargo.toml
Victor-Savu: Yup! I suppose 1.1 would already be available by the time this PR lands, so it might make sense to use that instead and release these changes as v 1.2. (same reasoning if 1.2 is released as well before this PR)

pest_meta/Cargo.toml
Victor-Savu: Any suggestions as to what to do with the example code that depends on `colored`?

pest_meta/src/ast.rs
Victor-Savu: Ahh, yes! :) I was wondering why this wasn't part of the PR which introduced lifetimes to begin with. I will put it in, though it does feel a bit out of scope for this PR.

pest_meta/src/validator.rs
Victor-Savu: The second lifetime is needed because the nodes can't live as long as the slices they reference. If we forced `&'i ParserNode<'i>` the nodes referenced in the HashMap would need to live at least as long as lifetime `'i`, which is also the lifetime of the slice that they reference. This, however, is not possible, since the node cannot live beyond the input as that would invalidate the references that it holds to the input.

pest_meta/src/lib.rs
Victor-Savu: I will use `rustfmt` on this file. Would you recommend using `cargo fmt` to format all 4 crates? That would create a lot of changes, and in some places arguably wrong formatting. I feel like rustfmt is not really ready for prime time yet.

pest_derive/src/lib.rs
Victor-Savu: I suppose `soi` would need one as well?

pest_derive/src/optimizer.rs
Victor-Savu: If I run rustfmt on this file, I get a bunch of changes (many not related to the changes I made) Should I just go ahead and use rustfmt, or should I just remove these spaces? I am using rustfmt 0.8.0

Victor-Savu:
Thanks for the review! Any suggestions regarding the two failing tests? I really don't know why it tries to match those extra rules.

pest_meta/src/lib.rs
dragostis: Yeah, `rustfmt` is not ready for prime-time. :/ I'm still kind of torn between whether or not to use it, but for now, I guess it's fine. I expect formatting to get better in time. Anyway, I definitely don't mind the changes if they're concentrated in a last commit.

dragostis:
> Any suggestions regarding the two failing tests? I really don't know why it tries to match those extra rules.

Yes, it's the first comment. 😛 

> This should be `grammar_rules = _{ soi ~ grammar_rule+ ~ eoi }`.

The error reporting can be a bit flaky, I suppose. The problem is that the logic doesn't expect multiple reports on the same exact position the way it happens here. Anyway, this is a paper cut for a subsequent release.

I feel like it doesn't make sense to release [`v1.1`](https://github.com/pest-parser/pest/milestone/4) without this fix and the interactive website. 

By the way, I really like your work with the PR and I think it makes sense for you to join the team if you want to continue developing pest. I definitely have high hopes for the future of the project!

Victor-Savu:
Thanks! I think I missed that comment. Al green now! 

Thank you very much for the invitation! I'd be happy to contribute more!

pest_meta/src/lib.rs
Victor-Savu: We could do a `cargo fmt` as part of a dedicated PR which only deals with formatting. Merging (non-fast-forward) can be quite difficult after such changes since there would be conflicts on so many lines.

Victor-Savu:
I did some measurements for the build times from scratch of the entire cargo workspace and this change increases the build time by 50% on average both in debug (from 16.3s to 24.6s) as well as in release (from 52.6s to 76.2s).
However, there is no change in the build time of the client crates. For example, the individual build time for the `pest_grammars` crate is still at about 3s in debug and 19s in release before and after this change.

Victor-Savu:
I did some measurements for the build times from scratch of the entire cargo workspace and this change increases the build time by 50% on average both in debug (from 16.3s to 24.6s) as well as in release (from 52.6s to 76.2s).
However, there is no change in the build time of the client crates. For example, the individual build time for the `pest_grammars` crate is still at about 3s in debug and 19s in release before and after this change.

dragostis:
I think that's fine. This number is only problematic when using `cargo install`, since you only build once, with release, and some crates do take a very long time to compile due to a large number of dependencies. For example, depending on Cargo for some very small amount of API used means that your executable will take minutes to compile.

pest_derive/src/optimizer.rs
```diff
@@ -91,23 +91,6 @@ pub fn optimize(rules: Vec<Rule>) -> Vec<Rule> {
                     ty,
                     expr: expr.map_bottom_up(rotate_right)
                         .map_bottom_up(unroll_loops)
-                        .map_bottom_up(|expr| {
```
dragostis: I see the issue here: since strings are not owned anymore, concatenation is not possible. Seems like we'll have to revert this change.

Victor-Savu: Do you happen to have any measurements comparing the string concatenation optimization with the slice optimization? The slice optimization prevents memory allocation, prevents copying and improves referential locality, whereas the string concatenation optimization reduces the number of nodes. On first glance I don't really see a clear cut advantage of one optimization over another. However, they can coexist :)
My proposal would be to keep the string slice references and to postpone the concatenation optimization for a separate PR. The reason for this is that the The concatenated strings could be represented later using another `Rule` variant containing a `String` (or another data structure, such as a `Vec[&'i str]`).

Victor-Savu:
> I'm trying to merge this on v1.1

If there is any way I can help with that, please let me know.

I wanted to ask if `pest` is using a particular branching model. If not, I would highly recommend [GitFlow](http://nvie.com/posts/a-successful-git-branching-model/). Adopting such a well known branching model could benefit the project development in the long term by providing contributors a familiar way of working.

Victor-Savu:
Also, Hipchat or another chat service for development would be useful :) I have really abused this PR thread for discussions which are not related to the changes, but rather to the `pest` development as a whole.

pest_derive/src/optimizer.rs
dragostis: The slice optimization happens to make compilation faster, while the concatenation works on improving parse performance once the actual code is generated which is why I would definitely go for the latter. Apart from that, the actual grammar parsing/optimization/generation step is very small during the whole compilation phase which is why I didn't bother at all to do any optimizations on this side and went for code clarity. (as best as I could while still trying to get features out 😃)

Having another variant for concatenations is not a bad idea, but wouldn't it flood the optimization and generation steps too much? Since this would require you to deal with multiple cases that, in the end, reach the same point. The reason why I find the `&str` to be advantageous is because it is way easier to write tests without all the ugly `.to_owned()` labels everywhere. However, this could be improved by writing some macros for test improvement. Quite frankly, tests are extremely verbose since I haven't had the time to invest in making them a bit nicer. However, I find their explicitness to be rather helpful. (if you kind of know your way around the code)

So, I would definitely go for the owned approach.

dragostis:
> If there is any way I can help with that, please let me know.
>
> I wanted to ask if `pest` is using a particular branching model. If not, I would highly recommend [GitFlow](http://nvie.com/posts/a-successful-git-branching-model/). Adopting such a well known branching model could benefit the project development in the long term by providing contributors a familiar way of working.

I'm aware of GitFlow, but I guess I got sidetracked at one particular point. I'll merge `v1.1` on master, then branch out `v1.0` for backward compatibility.

> Also, Hipchat or another chat service for development would be useful :) I have really abused this PR thread for discussions which are not related to the changes, but rather to the `pest` development as a whole.

The Gitter channel is in active use. :D

PR #161, opened 2017-10-12
Allow trailing comma in macros

dbrgn:
Based on https://elliotekj.com/2017/07/03/handling-optional-trailing-commas-macro_rules/.

Before, this was invalid:

```rust
    #[test]
    fn parses_to() {
        parses_to! {
            parser: AbcParser,
            input: "abcde",
            rule: Rule::a,
            tokens: [
                a(0, 3, [
                    b(1, 2),
                ]),
                c(4, 5)
            ]
        };
    }
```

I think I understand the `parses_to` macro invocation, but I can't really say that about the `consumes_to` macro, so it would be good if you could really double-check the changes :)

### DIFF
--- a/pest/src/macros.rs
+++ b/pest/src/macros.rs
@@ -39,7 +39,7 @@ macro_rules! consumes_to {
         };
     };
     ( $rules:ident, $tokens:expr, [ $name:ident ( $start:expr, $end:expr ),
-                                    $( $names:ident $calls:tt ),* ] ) => {
+                                    $( $names:ident $calls:tt ),* $(,)* ] ) => {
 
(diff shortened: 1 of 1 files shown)


dragostis:
Thanks for the PR. This looks good, but I think it's a good idea to add the trailing comma to `fails_with!` as well.

dragostis:
Any updates on this?

dbrgn:
Sorry, I was a bit busy :)

Where exactly would you add that potential trailing comma? The `fails_with!` macro doesn't have a `tokens` key, or any other "square bracket" list, right?

dragostis:
Good point. I completely forgot that those were actual `vec!` calls.

PR #167, opened 2017-11-12
Use fully-qualified path for Result

jturner314:
This is necessary in case the client code redefines `Result` with a type alias.


dragostis:
Thanks for the PRs and sorry for the late reply. I was away for a few days.

PR #171, opened 2017-11-18
Add minimum and maximum repetition rules

scowcron:
Currently, the .pest grammar files allowed the following repetition patterns:

* `a*` - 0 or more
* `a+` - 1 or more
* `a{n}` - exactly n
* `a{m,n}` - from m to n

This PR adds two more forms:

* `a{,n}` - up to n
* `a{n,}` - at least n

This is largely syntactic sugar. Minimum matching can also be achieved using the pattern `a{n} ~ a*`, and maximum can be achieved using `a{0,n}` or `a?{n}`.

What the new forms bring to the table is a consistent, unified syntax for dealing with counted repetitions greater than 1.


scowcron:
I'll be jumping on this in the next couple of days. Haven't looked too close to the validator code yet, but it definitely needs to address RepMin at a minimum.

PR #177, opened 2017-12-29
Add `input_range` method to `Pair`.

jamesotron:
Initially I was going to make `pos()` public, but then I decided that since `Pair` had a reference to `input` then would be better to just have it expose `line_col` instead, since that was the ultimate goal for me.

### DIFF
--- a/pest/src/iterators/pair.rs
+++ b/pest/src/iterators/pair.rs
@@ -200,6 +200,32 @@ impl<R: RuleType, I: Input> Pair<R, I> {
         )
     }
 
+    /// Return the line and column number of the token in the input stream.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use std::rc::Rc;
+    /// # use pest;
+    /// # use pest::inputs::StringInput;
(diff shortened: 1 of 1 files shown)


dragostis:
This is not quite right since a `Pair` can will usually span over multiple columns, maybe even lines. Do you perhaps want to find the line and column of the starting position of the `Pair`?

jamesotron:
Yes, I was trying to find the starting position, although now that you mention it could be quite useful to get the end position also.  I need this information to show the user non-parsing-related error messages during compilation.

dragostis:
It may be a better idea to expose the start and end position.

jamesotron:
I've exposed the start and end position of the `Pair` in the input, however I'm concerned about https://github.com/jamesotron/pest/commit/c0ed4ddc7999d98170f3974de18de78973bc1d46#diff-bb8e329f6a128880f3a917e70c24ea14R230 - it appears that the end position of each token is advanced by one.

dragostis:
I'm closing this since the functionality is available through `Pair::into_span`.

PR #183, opened 2018-01-21
Trivial improvements

ignatenkobrain:


ignatenkobrain:
It would be awesome if you could release new versions of crates after you merge this. It would help me in updating pest in Fedora. Thanks!

pest/Cargo.toml
```diff
@@ -8,7 +8,7 @@ documentation = "https://docs.rs/pest"
 keywords = ["pest", "parser", "peg", "grammar"]
 categories = ["parsing"]
 license = "MIT/Apache-2.0"
-readme = "_README.md"
```
dragostis: The reason why I named this `_README.md` was because I don't want the same readme to show up in the crates as well. Maybe one day they will have separate readme files?

ignatenkobrain: but how is it different? Right now it is symlink to main readme, so you will change it later? I mean filename itself doesn't matter (but confusing if it has underscore).

dragostis: It doesn't show up on GitHub when you navigate to the `pest/` folder, for example.

dragostis: @ignatenkobrain Any thoughts on the matter?

ignatenkobrain: sorry, missed previous reply. without underscore it makes more sense to me because every other crate does this... But if you don't want this, I can safely drop this commit.  Just tell me what you prefer

dragostis: I would definitely prefer it this way, especially since the path is provided in the manifest.

dragostis:
I've published `1.0.2`.

PR #192, opened 2018-01-29
v1.1

dragostis:


pest/src/span.rs
```diff
@@ -152,7 +154,7 @@ impl<'i> fmt::Debug for Span<'i> {
 
 impl<'i> Clone for Span<'i> {
     fn clone(&self) -> Span<'i> {
-        new(self.input, self.start, self.end)
+        unsafe { new(self.input, self.start, self.end) }
```
sphinxc0re: Why did you put this `unsafe` here?

dragostis: Creating a `Span` from random positions is unsafe, but since it was private, it didn't matter much. Having it explicit should make development/contributing a nicer experience.

sphinxc0re: Sounds reasonable

Issue #176, opened 2017-12-08
Unmatched expressions modify the stack


jturner314:
While working with `pest`, I discovered that expressions that aren't matched still can modify the stack with `push` or `pop`, which seems like a bug. Consider:

`grammar.pest`
```rust
foo = ${ soi ~ push("") ~ (aba | a) ~ eoi }
a = @{ peek ~ "a" }
aba = @{ push("a") ~ "b" ~ pop }
```

`main.rs`
```rust
#[macro_use]
extern crate pest_derive;
extern crate pest;

use pest::Parser;

#[cfg(debug_assertions)]
const _GRAMMAR: &'static str = include_str!("grammar.pest");

#[derive(Parser)]
#[grammar = "grammar.pest"]
pub struct MyParser;

fn parse_string(s: &str) {
    match MyParser::parse_str(Rule::foo, s) {
        Ok(o) => println!("ok: {:?}", o),
        Err(e) => println!("err: {}", e),
    }
}

fn main() {
    parse_string("a");
    parse_string("aa");
}
```
The output when running this is
```
err:  --> 1:1
  |
1 | a
  | ^---
  |
  = expected foo
ok: Pairs { pairs: [Pair { rule: foo, span: Span { start: 0, end: 2 }, inner: Pairs { pairs: [Pair { rule: a, span: Span { start: 0, end: 2 }, inner: Pairs { pairs: [] } }] } }] }
```

I would expect The string `"a"` to parse correctly, and the string `"aa"` to fail parsing, but the parser does the opposite. What I think happens is this:

1. Starting with the `foo` rule, `soi ~ push("")` matches, so the stack is now `[""]`.
2. The next step is `~ (aba | a)`, so the parser tries rule `aba` first.
3. The parser sees that `push("a")` matches, so the stack is now `["", "a"]`.
4. The parser sees that `~ "b"` doesn't match, so `aba` doesn't match, so it goes back up to `~ (aba | a)` and tries `a` next. Note that even though `aba` didn't match, it modified the stack.
5. We're in rule `a` now. Since the top of the stack is `"a"`, `peek ~ "a"` can only match the string `"aa"`, not `"a"`.
6. In the `"aa"` example, the final part is `~ eoi`, which matches.

The surprising part is that even though `aba` didn't match, it still modified the stack. This seems like a bug. I would expect the stack to be left unmodified by expressions that don't match (i.e. restore the stack after the expression fails to match).

[...]

PR #208, opened 2018-03-17
Refactor the `ParserState` transformation API to be driven entirely from `ParserState`

jstnlef:
When looking into https://github.com/pest-parser/pest/issues/176, I noticed that it was going to be quite difficult to ensure consistency when rolling back state on `Err` for choice and optional rules if the user level API allowed for arbitrary transformations of the `Position`. In order to account for this and to simplify the API, I have modified the API to no longer take a `Position` and a mutable reference to a `ParserState` for the methods' closures. Now, all of the transformation methods on `ParserState` will take a `Box<ParserState>` and return a `Result<Box<ParserState>>` and the `ParserState` owns the current `Position` on the input.

The upshot of this is that the API is much simpler (no need to wrap pos.sequence in state.sequence for example) and the performance is equivalent to the original.

However, this PR will not solve https://github.com/pest-parser/pest/issues/176 in its current state. I'm going to be writing a subsequent change to solve that issue.


pest/src/parser_state.rs
```diff
@@ -42,61 +48,101 @@ pub struct ParserState<'i, R: RuleType> {
...
+    /// Allocates a fresh `ParserState` object to the heap and returns the owned `Box`. This `Box`
+    /// will be passed from closure to closure based on the needs of the specified `Parser`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use pest;
+    ///
+    /// let input = "";
+    /// let state: Box<pest::ParserState<&str>> = pest::ParserState::new(input);
```
dragostis: I wouldn't add the type explicitation here.

pest/src/parser_state.rs
```diff
@@ -42,61 +48,101 @@ pub struct ParserState<'i, R: RuleType> {
...
+            pos_attempts: vec![],
+            neg_attempts: vec![],
+            attempt_pos: 0,
+            atomicity: Atomicity::NonAtomic,
+            stack: Stack::new()
+        })
+    }
+
+    /// Returns a reference to the current `Position` of the `ParserState`.
+    ///
```
dragostis: I would add very simple examples to all docs.

pest/src/parser_state.rs
```diff
@@ -234,45 +301,350 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    #[inline]
+    pub fn skip_until(mut self: Box<Self>, string: &str) -> ParseResult<Box<ParserState<'i, R>>> {
+        if self.position.skip_until(string) {
+            Ok(self)
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Asks the `ParserState` to match the start of the `input`. If the match is successful, an
```
dragostis: Since `input` is not an argument, I would remove the back-ticks.

pest/src/parser_state.rs
```diff
@@ -234,45 +301,350 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    #[inline]
+    pub fn start_of_input(self: Box<Self>) -> ParseResult<Box<ParserState<'i, R>>> {
+        if self.position.at_start() {
+            Ok(self)
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Asks the `ParserState` to match the end of the `input`. If the match is successful, an
```
dragostis: Same here.

pest/src/parser_state.rs
```diff
@@ -42,61 +48,101 @@ pub struct ParserState<'i, R: RuleType> {
...
+    /// not be needed as manually manipulating the `Position` can lead to a future inconsistent
+    /// state.
+    ///
+    pub fn clone_position(&self) -> Position<'i> {
+        self.position.clone()
+    }
+
+    /// Given a `Position`, replace the current `Position` of the `ParserState` with the new one.
+    ///
+    pub fn with_updated_position(mut self: Box<Self>, position: Position<'i>) -> Box<Self> {
```
dragostis: I would expect a `with` function to take a closure as an argument. Maybe `updated_position` or simply `update_position` would be enough?

pest/src/stack.rs
```diff
@@ -0,0 +1,177 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. All files in the project carrying such notice may not be copied,
+// modified, or distributed except according to those terms.
+
+/// Implementation of a `Stack` which maintains an log of `StackOp`s in order to rewind the stack
```
dragostis: "a ~~~an~~~ log"

pest/src/stack.rs
```diff
@@ -0,0 +1,177 @@
...
+/// to a previous state.
+#[derive(Debug)]
+pub struct Stack<T: Clone> {
+    ops: Vec<StackOp<T>>,
+    cache: Vec<T>,
+    snapshots: Vec<usize>
```
dragostis: It might be a bit early to say this, but `snapshots` may end up very similar to the `queue.truncate`-ion that we use in `ParserState`. In case it does, it makes sense to have one way to do this. (either on the call stack or in a `Vec`)

pest/src/stack.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /// Returns the top-most `&T` in the `Stack`.
+    pub fn peek(&self) -> Option<&T> {
+        self.cache.last()
+    }
+
+    /// Pushes a `T` onto the `Stack`.
+    pub fn push(&mut self, elem: T) {
+        self.ops.push(StackOp::Push(elem.clone()));
```
dragostis: Cloning here might be costly. Again, hard to say before we see exactly how this will be implemented.

pest/src/stack.rs
```diff
@@ -0,0 +1,177 @@
...
+        match self.snapshots.pop() {
+            Some(index) => {
+                self.ops.truncate(index);
+                self.recache();
+            }
+            None => {}
+        }
+    }
+
+    // Rebuilds the internal cache based on the previous operations stored.
```
dragostis: It might make sense to have an implementation that takes better advantage of the cache by not clearing it, instead making snapshots of the stack content every now and then. (when I say content, I mean top and concatenated content, in which case a less general `&str` instead of `T` implementation would make sense)

pest/src/parser_state.rs
jstnlef: This type annotation is required in order for this to compile since rust need to know the explicit type of `R` in `pest::ParserState<R>`

pest/src/parser_state.rs
jstnlef: That's interesting. I wouldn't have necessarily thought a function prefixed with `with` takes a closure. The rationale is that I'm asking the `Box<Self>` to give me a `Box<Self>` with an updated position. `update_position`, at least to me, connotes mutation. Willing to change it though.

pest/src/stack.rs
jstnlef: Would you prefer that I remove this from this PR or just leave it with the expectation that it's going to get changed significantly when addressing #176?

pest/src/parser_state.rs
jstnlef: Will do 👍

pest/src/parser_state.rs
dragostis: Oh, of course. 😄

pest/src/stack.rs
dragostis: I'm actually exploring the case where `Stack` is added in the following PR that solves #176. Like that it will be easier to see exactly what kind of implications it will have.

derive/src/lib.rs
```diff
@@ -164,6 +164,13 @@
 //!
 //!     where `e`, `e1`, and `e2` are expressions.
 //!
+//! Expressions can modify the stack only if they match the input. For example,
+//! if `e1` in the compound expression `e1 | e2` does not match the input, then
+//! it does not modify the stack, so `e2` sees the stack in the same state as
+//! `e1` did. Repetitions and optionals (`e*`, `e+`, `e{, n}`, `e{n,}`,
+//! `e{m,n}`, `e?`) can modify the stack each time `e` matches. The `!e`
```
dragostis: I would also add the positive lookahead case here. (`&e`)

meta/Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "_README.md"
 
 [dependencies]
 maplit = "^0.1.5"
-pest = { path = "../pest", version = "^1.0" }
+pest = "^1.0"
```
dragostis: Why not have the path dependency here? It makes development much easier.

meta/src/ast.rs
```diff
@@ -25,22 +25,39 @@ pub enum RuleType {
...
     RepOnce(Box<Expr>),
+    /// Matches an expression an exact number of times, e.g. `e{n}`
     RepExact(Box<Expr>, u32),
+    /// Matches an expression at least a number of times, e.g. `e{n,}`
     RepMin(Box<Expr>, u32),
+    /// Matches an expression at most a number of times, e.g. `e{,n}`
     RepMax(Box<Expr>, u32),
+    /// Matches an expression at range of times, e.g. `e{m, n}`
     RepMinMax(Box<Expr>, u32, u32),
+    /// Matches any expression that does not contain the strings in the Vec.
```
dragostis: I would change this to "Matches anything until the strings in the `Vec` are found.".

pest/src/lib.rs
```diff
@@ -58,8 +58,14 @@
 
 extern crate memchr;
 
+pub use error::Error;
```
dragostis: I would generally publish these after `mod` definitions. Was this generated by `rustfmt`?

pest/src/parser_state.rs
```diff
@@ -42,61 +48,151 @@ pub struct ParserState<'i, R: RuleType> {
...
-        Err(Error::ParsingError {
-            positives: state.pos_attempts,
-            negatives: state.neg_attempts,
-            // All attempted positions were legal.
-            pos: unsafe { position::new(input.as_bytes(), state.attempt_pos) }
-        })
+    let state = ParserState::new(input);
+
+    match f(state) {
+        Ok(s) => {
```
dragostis: I would call this `state` instead of `s`, just to keep everything consistent, regardless of the shadowing.

pest/src/parser_state.rs
```diff
@@ -42,61 +48,151 @@ pub struct ParserState<'i, R: RuleType> {
...
+    /// let input = "ab";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let position = state.get_position();
+    /// assert_eq!(position.pos(), 0);
+    /// ```
+    pub fn get_position(&self) -> &Position<'i> {
+        &self.position
+    }
+
+    /// Returns a clone of the current `Position` of the `ParserState`. Ideally, this will likely
```
dragostis: Not completely sure, but maybe a better API here would be to have a more functional approach. Instead of imperatively cloning and reassigning the position, we could use a `map_position<F>(f: F) where F: FnOnce(Position) -> Position`.

pest/src/parser_state.rs
```diff
@@ -234,45 +351,350 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    {
+        let result = f(self);
+
+        match result {
+            Ok(state) | Err(state) => Ok(state)
+        }
+    }
+
+    /// Asks the `ParserState` to match the given `string`. If the match is successful, the
+    /// `position` will be updated and an `Ok` `Result` is returned. If failed, an `Err` is
```
dragostis: I would explain here what the `Result` contains, much like in the above doc.

pest/src/lib.rs
jstnlef: This was formatted with `rustfmt`. I can put it back if you want.

meta/Cargo.toml
jstnlef: This has to be changed because `pest_meta` depends on both `pest_derive` and `pest` and with the significant changes made to `pest`, `pest_meta` won't compile. Also, in general, both `pest` and `pest_derive` should point to the most recently released versions of those libraries since this is the library that `pest_derive` depends on.

pest/src/stack.rs
jstnlef: Originally, the stack holds `Span` structs. I expected that that wouldn't necessarily have to change. `Span`s are pretty cheap to clone.

pest/src/stack.rs
jstnlef: Not quite sure I follow there. When you keep popping off values from the stack, you need to know where the boundaries of the individual strings are no? A top and concatenated properties would lose context no?

derive/src/lib.rs
```diff
@@ -168,7 +168,7 @@
 //! if `e1` in the compound expression `e1 | e2` does not match the input, then
 //! it does not modify the stack, so `e2` sees the stack in the same state as
 //! `e1` did. Repetitions and optionals (`e*`, `e+`, `e{, n}`, `e{n,}`,
-//! `e{m,n}`, `e?`) can modify the stack each time `e` matches. The `!e`
+//! `e{m,n}`, `e?`, `&e`) can modify the stack each time `e` matches. The `!e`
```
dragostis: `&e` should be next to `!e`, not in the parens, since it's lookahead as well.

meta/src/ast.rs
```diff
@@ -124,7 +124,7 @@ impl Expr {
                     let mapped = Box::new(map_internal(*expr, f));
                     Expr::Push(mapped)
                 }
-                expr => expr
+                expr => expr,
```
dragostis: Has `rustfmt` added these commas? The `rustfmt.toml` explicitly bans them.

pest/src/parser_state.rs
dragostis: Any thoughts on this one?

derive/src/lib.rs
```diff
@@ -312,9 +312,9 @@ pub fn derive_parser(input: TokenStream) -> TokenStream {
                 Rule::insensitive_string => "`^`".to_owned(),
                 Rule::range_operator => "`..`".to_owned(),
                 Rule::single_quote => "`'`".to_owned(),
-                other_rule => format!("{:?}", other_rule),
-            })
-        ),
+                other_rule => format!("{:?}", other_rule)
+            },)
```
dragostis: Seems to still have an issue.

derive/src/lib.rs
```diff
@@ -301,7 +313,7 @@ pub fn derive_parser(input: TokenStream) -> TokenStream {
                 Rule::range_operator => "`..`".to_owned(),
                 Rule::single_quote => "`'`".to_owned(),
                 other_rule => format!("{:?}", other_rule)
-            })
+            },)
```
dragostis: Trailing comma here.

pest/src/parser_state.rs
```diff
@@ -351,10 +813,152 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+            Ok(mut state) => {
+                state.stack.pop();
+                Ok(state)
+            }
+            Err(state) => Err(state)
+        }
+    }
+
+    /// Drops the top of the stack and returns `Ok(Box<ParserState>)` if there was a value to
+    /// drop. Otherwise, it returns `Err<Box<ParserState>)`.
```
dragostis: This should be `Err(Box<ParserState>)`.

pest/src/position.rs
```diff
@@ -91,17 +89,6 @@ impl<'i> Position<'i> {
     }
 
     /// Returns the line - and column number pair of the current `Position`.
-    ///
```
dragostis: Why remove the example here? `Position::line_col` is useful to the user.

pest/src/position.rs
```diff
@@ -148,18 +135,7 @@ impl<'i> Position<'i> {
         line_col
     }
 
-    /// Returns the actual line of the current `Position`.
```
dragostis: Same here.

pest/src/position.rs
```diff
@@ -216,67 +192,22 @@ impl<'i> Position<'i> {
...
-    /// ```
+    /// Returns `true` when the `Position` points to the start of the input `&str`.
     #[inline]
-    pub fn at_start(self) -> Result<Position<'i>, Position<'i>> {
-        if self.pos == 0 {
-            Ok(self)
-        } else {
-            Err(self)
-        }
+    pub fn at_start(&self) -> bool {
```
dragostis: Shouldn't this be `pub(crate)`?

pest/src/position.rs
```diff
@@ -216,67 +192,22 @@ impl<'i> Position<'i> {
...
-    /// ```
+    /// Returns `true` when the `Position` points to the end of the input `&str`.
     #[inline]
-    pub fn at_end(self) -> Result<Position<'i>, Position<'i>> {
-        if self.pos == self.input.len() {
-            Ok(self)
-        } else {
-            Err(self)
-        }
+    pub fn at_end(&self) -> bool {
```
dragostis: `pub(crate)`?

pest/src/span.rs
```diff
@@ -127,18 +127,6 @@ impl<'i> Span<'i> {
     }
 
     /// Captures a slice from the `&str` defined by the `Span`.
-    ///
```
dragostis: This should probably also not be deleted since it's user-facing.

pest/src/position.rs
jstnlef: I'll come up with a better example.

pest/src/span.rs
jstnlef: I'll come up with a better example here as well

pest/src/parser_state.rs
```diff
@@ -141,38 +198,60 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
 
-                self.queue.push(QueueableToken::End {
-                    rule,
-                    pos: pos.pos()
-                });
-            } else {
-                self.queue.truncate(index);
+        match result {
+            Ok(mut new_state) => {
+                if new_state.lookahead == Lookahead::Negative {
```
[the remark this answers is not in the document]
jstnlef: I'll take another look at this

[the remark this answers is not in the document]
jstnlef: So, with the new API I'm not sure it's feasible to to keep the old logic. The trouble is that you really want to call `track` on the resulting `new_state` based on the new_state's lookahead, but if you try to unwrap `result` before the match, the value is moved.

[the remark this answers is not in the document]
dragostis: I see. It's fine then.

PR #212, opened 2018-03-25
Bump the version to 2.0.0

jstnlef:


meta/Cargo.toml
```diff
@@ -12,6 +12,7 @@ readme = "_README.md"
 
 [dependencies]
 maplit = "^0.1.5"
+# TODO: When we get 2.0.0 out, this will have to be bumped as well.
```
dragostis: I wouldn't bump this, since it will cause a ripple compilation effect: meta 2.0 will need meta 1.0 which will need derive 1.0. The more you bump, the longer the chain. This will not be a problem once cargo will come with prebuilt binaries.

jstnlef: This would be bumped some time after 2.0.0 gets released to crates.io. This is the bottom of the dependency tree, so once there are 2.0.0 crates, we can have meta depend on the most recently released ones and everything will work out fine.

Issue #203, opened 2018-03-09
syn 0.12 / quote 0.4


ignatenkobrain:
We have updated syn/quote/synstructure recently in Fedora which broke pest... It would be very nice if you could update dependencies of pest_derive. Thank you in advance!

dragostis:
Fixing this on `1.0.*` will be a tad involved. The plan was to have the new versions in `1.1`. Would that be okay? I backporting could also be an option, but it would be a lot of work.

cuviper:
It should be fine for us to update Fedora to pest 1.1 too.

PR #213, opened 2018-03-28
Upgrading to syn 0.12 and quote 0.4

jstnlef:
This is to address https://github.com/pest-parser/pest/issues/203

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -256,8 +256,7 @@ use std::io::{self, Read};
 use std::path::Path;
 
 use proc_macro::TokenStream;
-use quote::Ident;
-use syn::{Attribute, Lit, MetaItem};
+use syn::{Attribute, DeriveInput, Ident, Lit, Meta};
 
 mod generator;
 mod optimizer;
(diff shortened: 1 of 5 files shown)


jstnlef:
hrm. Looks like there are failing tests.
```
---- tests::derive_ok stdout ----
	thread 'tests::derive_ok' panicked at 'proc_macro::__internal::with_sess() called before set_parse_sess()!', /checkout/src/libproc_macro/lib.rs:822:9
```
I'm really not sure what to make of this. The code runs fine when not run in a testing context but I can't seem to figure out how to get these to work. Any thoughts? Specifically, they fail when trying to create the TokenStream object from the definition string.
```
let tokens = TokenStream::from_str(definition).unwrap();
```

dragostis:
This seems to be a bug in the `proc_macro` crate. I would imagine that `TokenStream::from_str` should not panic.

cuviper:
You could let `parse_derive` and those tests work directly with `DeriveInput` instead:
https://paste.fedoraproject.org/paste/3SaTTY1DfyDCf2UoeP8EGg

jstnlef:
The works for me. I'll make the change tonight.

derive/src/lib.rs
```diff
@@ -332,20 +330,19 @@ fn read_file<P: AsRef<Path>>(path: P) -> io::Result<String> {
...
-            MetaItem::NameValue(ref ident, _) => format!("{}", ident) == "grammar",
+        .filter(|attr| match attr.interpret_meta() {
+            Some(Meta::NameValue(name_value)) => name_value.ident.to_string() == "grammar",
             _ => false
         })
         .collect();
 
     let filename = match grammar.len() {
-        0 => panic!("a grammar file needs to be provided with the #[grammar(\"...\")] attribute"),
+        0 => panic!("a grammar file needs to be provided with the #[grammar = \"...\"] attribute"),
```
dragostis: Ouch. This is very sad leftover. Glad you found it!

PR #214, opened 2018-03-29
Improve the performance of skip_until_any

jstnlef:
I was looking to see why there was such a performance dip on these [benchmarks](https://github.com/Geal/parser_benchmarks/tree/master/json) after commit c6e6c649f740307ce903699f1cb7ddcf87b1a10c. Originally I thought it might have to do with repeated calls to libc's `memchr` function, essentially eating a cost of calling a foreign library. However, this was definitely not the case as I tried to use OSX's `memmem` libc function to achieve the same results without multiple calls out from the rust code and achieved the exact same results.
Then, after a bit more digging, I realized that the entire issue was around the skip optimizer itself. To prove this, I commented out the optimizer and ended up achieving similar results to before the commit so I knew I was onto something.  

Following this path, it led me to the implementation of `skip_until_any`, the new function I made as part of the 2.0 API refactor (initially in generator.rs) when it hit me. The way that the original function worked was that it skipped ahead in the input until it found (or didn't) the first character for the first string, and then each subsequent string. This means that in the worst possible case, you could be iterating the entire input "number of strings" times and still come up empty. Instead, for the case where I'm looking for a number of strings, at the current byte position, I'm going to check all of the strings here for a match. If none are found, I'm going to bump a byte and check again. The expectation is that I'll either find what I'm looking for sooner rather than later or I iterate through the entire input only once and error. Also, I only have to eat the cost of converting the strings to bytes once.

The results speak for themselves.
Before:
```
running 4 tests
test apache ... bench:   5,794,511 ns/iter (+/- 585,674) = 21 MB/s
test basic  ... bench:       1,602 ns/iter (+/- 949) = 47 MB/s
test canada ... bench:  31,945,573 ns/iter (+/- 2,851,110) = 70 MB/s
test data   ... bench:      63,938 ns/iter (+/- 11,822) = 144 MB/s
```
After:
```
running 4 tests
test apache ... bench:     666,037 ns/iter (+/- 121,526) = 191 MB/s
test basic  ... bench:       1,485 ns/iter (+/- 607) = 51 MB/s
test canada ... bench:  32,129,002 ns/iter (+/- 3,136,427) = 70 MB/s
test data   ... bench:      55,436 ns/iter (+/- 32,818) = 166 MB/s
```


pest/src/position.rs
```diff
@@ -282,6 +282,25 @@ impl<'i> Position<'i> {
         false
     }
 
+    /// Skips until the one of the given `strings` is found. If none of the `strings` can be found,
+    /// this function will return `false` and its `pos` will not be updated.
+    #[inline]
+    pub(crate) fn skip_until_any(&mut self, strings: &[&str]) -> bool {
```
dragostis: We should probably only keep `skip_until_any` and remove `skip_until` with its depedency.

pest/src/position.rs
```diff
@@ -282,6 +282,25 @@ impl<'i> Position<'i> {
...
+        let mut current_pos = self.pos;
+
+        while current_pos < self.input.len() {
+            for slice in strings.iter().map(|s| s.as_bytes()) {
+                if slice == &self.input[current_pos..current_pos + slice.len()] {
+                    self.pos = current_pos;
+                    return true;
+                }
+            }
+            current_pos += 1;
```
dragostis: I've tried replacing this `while` with SIMD-friendly [jetscii](https://github.com/shepmaster/jetscii/), but, quite interestigly, it did not improve performance in any of the benchmarks. Instead, performance was a bit lower. Maybe it's still worth keeping an open mind about `jestscii` for future releases.

jstnlef:
Here are the new numbers when testing with the requested changes and rust 1.25
```
running 4 tests
test apache ... bench:     598,335 ns/iter (+/- 297,090) = 212 MB/s
test basic  ... bench:       1,377 ns/iter (+/- 710) = 55 MB/s
test canada ... bench:  29,317,882 ns/iter (+/- 3,546,155) = 76 MB/s
test data   ... bench:      48,559 ns/iter (+/- 24,376) = 190 MB/s
```

pest/src/position.rs
jstnlef: I think I'll just have the latter delegate to the former as an API convenience

pest/src/position.rs
```diff
@@ -262,21 +260,23 @@ impl<'i> Position<'i> {
     /// function will return `false` and its `pos` will not be updated.
     #[inline]
     pub(crate) fn skip_until(&mut self, string: &str) -> bool {
```
dragostis: `Position::skip_until` and `ParserState::skip_until` are not used any more, as far as I can tell.

jstnlef: I guess the question is whether or not we should keep it as a separate function. Another thing I can do is remove this and rename `skip_until_any` to `skip_until` and call it a day.

jstnlef: It might be nice to keep both for people using `pest` without `pest_derive`

dragostis: We could leave them both, sure.

jstnlef:
You mean specifically the one benchmark in `pest` right? Not the ones in `pest_grammar`

jstnlef:
Deleted the benchmarks

PR #219, opened 2018-03-31
remove #[inline] on lambdas

### LINKED ISSUE #218: FTBFS on 1.26.0-nightly (80785a547 2018-03-30)
Hey there,

`meta` currently fails to build from source on nightly:

```console
$ cargo build
   Compiling pest_meta v2.0.0 (file:///Users/kivikakk/Code/pest-parser/pest/meta)
error[E0518]: attribute should be applied to function
  --> meta/src/parser.rs:21:10
   |
21 | #[derive(Parser)]
   |          ^^^^^^ not a function

error: aborting due to previous error

For more information about this error, try `rustc --explain E0518`.
error: Could not compile `pest_meta`.

To learn more, run the command again with --verbose.
```

The `pest_derive` from crates.io it's using is 1.0.6.

`E0158` appears to be `#[inline(…)]` applied to a non-function:

```
This error indicates that an `#[inline(..)]` attribute was incorrectly placed
on something other than a function or method.
```

I've tried looking at the generated code with `--pretty=expanded` and … well, there's a lot of inlines, and I can't easily find one that doesn't correspond with a `fn` or lambda.

I can't easily find a change in rustc itself that would've caused this … so I'm going to try manually bisecting by installing nightlies that correspond to changes in the relevant code, i.e. https://github.com/rust-lang/rust/commits/361509320c882ee76f0d1359f842dec4637e6b08/src/librustc/hir/check_attr.rs. Wish me luck!

kivikakk:
Fixes #218.

Per https://github.com/rust-lang/rust/pull/49291#issuecomment-377682736, `#[inline]` has no effect on lambdas, and rustc has stopped allowing this in nightly.

### DIFF
--- a/derive/src/generator.rs
+++ b/derive/src/generator.rs
@@ -31,7 +31,7 @@ pub fn generate(name: Ident, rules: Vec<Rule>, defaults: Vec<&str>) -> Tokens {
         quote! {
             #[inline]
             pub fn eoi(state: Box<::pest::ParserState<Rule>>) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
-                state.rule(Rule::eoi, #[inline(always)] |state| {
+                state.rule(Rule::eoi, |state| {
                     state.end_of_input()
                 })
             }
@@ -171,7 +171,7 @@ fn generate_rule(rule: Rule) -> Tokens {
             let atomic = generate_expr_atomic(rule.expr);
 
             quote! {
-                state.atomic(::pest::Atomicity::Atomic, #[inline(always)] |state| {
+                state.atomic(::pest::Atomicity::Atomic, |state| {
                     #atomic
                 })
             }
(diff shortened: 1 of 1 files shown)


kivikakk:
I'd like to test that this _actually_ fixes the problem locally, but it's unclear to me quite how to do that. (do I set up an alternate registry?? that sounds like Not Fun.)

Kroisse:
I think it can be proven by compiling https://github.com/pest-parser/pest/blob/master/grammars/src/lib.rs

kivikakk:
I'm not sure how that helps: on rust nightly, it attempts to build `meta`, which uses `pest_derive` 1.0.6, which doesn't have this change, so the resulting generated code fails.

Kroisse:
I've tried that by adding `[patch.crates-io]` section to use the local version of `pest_derive`, but failed because `pest_derive` is depends on `pest_meta` so they make a circular dependency.

I think that is the only solution releasing 1.0.7 with this patch. It's possible because it still can be compiled on rustc 1.25 stable.

Kroisse:
@kivikakk I've tried to fix on `v1.0.6` instead `master` and it works.

PR #222, opened 2018-04-05
Fix conversion from Pair to Pairs

wirelyre:
This matters when using `Tokens`. Without this fix, `pair.into_inner().tokens()` will not iterate the final `End` token.


dragostis:
Thanks for the PR!

I'm not sure I understand where this goes wrong. Could you please write a very simple test case?

wirelyre:
The `Pairs` and `Tokens` iterators basically iterate over `start <= index < end` in the queue. We want `end` to point to the token *after* the last token.

But `self.pair()` points to the matching end token directly, which is already the token after the last inner token. No need to subtract one.

PR #224, opened 2018-04-10
Created as_span method

### LINKED ISSUE #223: Unable to get Span without consuming Pair
If getting a Span consumes a Pair, I have to choose between getting that Span and getting its inner Pairs. This is counter intuitive given that the two don't really seem related. Sometimes I want to look further into the contents of a pair and also get its position. Can we add a `span(&self)` method (or something) to get the Span of a Pair and still allow you to get its inner Pairs afterwards?

I would be happy to create a PR for this. :smile:

sunjay:
Fixes #223.

I left `into_span` so that this change is backwards compatible. `into_span` is now implemented in terms of `as_span` because they do the same thing.

Since the two methods are mostly the same, the test is also pretty much exactly the same. I added an extra part at the end to make sure that the pair is not consumed by the method.

### DIFF
--- a/pest/src/iterators/pair.rs
+++ b/pest/src/iterators/pair.rs
@@ -123,6 +123,35 @@ impl<'i, R: RuleType> Pair<'i, R> {
     /// ```
     #[inline]
     pub fn into_span(self) -> Span<'i> {
+        // into_span was here first, so it is left in for backwards compatibility
+        self.as_span()
+    }
+
+    /// Returns the `Span` defined by the `Pair`, **without** consuming it.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::rc::Rc;
+    /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {
+    ///     ab
+    /// }
+    ///
+    /// let input = "ab";
+    /// let pair = pest::state(input, |state| {
+    ///     // generating Token pair with Rule::ab ...
+    /// #     state.rule(Rule::ab, |s| s.match_string("ab"))
+    /// }).unwrap().next().unwrap();
+    ///
+    /// assert_eq!(pair.as_span().as_str(), "ab");
+    /// // pair is still available for use because it was not consumed
+    /// assert_eq!(pair.as_span().start(), 0);
+    /// ```
+    #[inline]
+    pub fn as_span(&self) -> Span<'i> {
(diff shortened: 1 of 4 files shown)


jstnlef:
We're already breaking all sorts of compatibility with 2.0.0 on the horizon. Not sure I would bother too much with maintaining compatibility. Looks good otherwise though 👍

pest/src/iterators/pair.rs
```diff
@@ -123,6 +123,35 @@ impl<'i, R: RuleType> Pair<'i, R> {
     /// ```
     #[inline]
     pub fn into_span(self) -> Span<'i> {
+        // into_span was here first, so it is left in for backwards compatibility
```
dragostis: Is there a way to mark this as deprecated?

pest/src/iterators/pair.rs
```diff
@@ -123,6 +123,35 @@ impl<'i, R: RuleType> Pair<'i, R> {
...
+        self.as_span()
+    }
+
+    /// Returns the `Span` defined by the `Pair`, **without** consuming it.
```
dragostis: I would keep the documentation the same for both. The only difference is that the deprecated one should point here.

dragostis:
@jstnlef I would keep backwards compatibility here since this is a pretty popular method and it's pretty simple to deprecate.

pest/src/iterators/pair.rs
dragostis: I'm afraid there [might be](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md). Seems to be stabilized too.

sunjay: Ah! Yes you are right. I was thinking of the [stability attributes](https://github.com/rust-lang/rfcs/issues/1491).

sunjay:
I deprecated `into_span` and replaced its occurrences with `as_span`. No more need for an explicit clone. :)

The docs for `as_span` are now exactly the same as the original documentation for `into_span`. The documentation for `into_span` contains a note that the method is deprecated. This is in addition to the one that will be generated by the Rust compiler whenever that method is used.

pest/src/iterators/pair.rs
```diff
@@ -122,7 +124,34 @@ impl<'i, R: RuleType> Pair<'i, R> {
     /// assert_eq!(pair.into_span().as_str(), "ab");
     /// ```
     #[inline]
+    #[deprecated(since="1.0.7", note="Please use `as_span` instead")]
```
dragostis: Please change to `2.0.0`.

sunjay:
Done. :)

PR #226, opened 2018-04-14
Fix custom derive for structs with type parameters

### LINKED ISSUE #215: Custom derive fails on structs with type parameters
## How to reproduce

```rust
#[derive(Parser)]
pub struct Parser<'a, T>;
```

generates the compile-time errors `error[E0106]: missing lifetime specifier` and `error[E0243]: wrong number of type arguments`, but confusingly points to the identifier `Parser` as the source of the errors.

## Expected behavior

The custom derivation would work on any struct, regardless of type parameters, or else fail with a descriptive error message.

wirelyre:
Closes #215.

I'll need some help writing a test for this. It seems that making a [`Generics`](https://docs.rs/syn/0.13/syn/struct.Generics.html) directly is quite awkward.

### DIFF
--- a/derive/src/generator.rs
+++ b/derive/src/generator.rs
@@ -10,11 +10,11 @@
 use std::collections::HashMap;
 
 use quote::Tokens;
-use syn::Ident;
+use syn::{Generics, Ident};
 
 use pest_meta::ast::*;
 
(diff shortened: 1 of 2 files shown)


wirelyre:
Fair enough! It's basically lifted straight from the `syn` documentation anyway.

PR #227, opened 2018-04-15
[WIP] Fix unmatched stack ops

jstnlef:
Attempts to fix: https://github.com/pest-parser/pest/issues/176
I wanted to post this PR to get some feedback. Basically what I've done is to make sure that the code generator only checkpoints the state of the parser on logical branches that could possibly manipulate the state of the stack.


derive/src/optimizer/mod.rs
```diff
@@ -15,17 +16,30 @@ mod rotater;
...
         .map(skipper::skip)
         .map(unroller::unroll)
         .map(concatenator::concatenate)
         .map(factorizer::factor)
+        .map(|rule| restorer::restore_on_err(rule, &rules_to_exprs))
         .collect()
 }
 
+fn populate_rules_to_exprs(rules: &[Rule]) -> HashMap<String, Expr> {
+    let mut rules_to_exprs = HashMap::new();
```
dragostis: There's a [nicer way](https://github.com/pest-parser/pest/blob/master/meta/src/validator.rs#L373) to write this. 😉 Some of that code could actually be extracted and made more abstract so that it can be used in both cases.

derive/src/optimizer/restorer.rs
```diff
@@ -0,0 +1,77 @@
...
+            Rule {
+                name,
+                ty,
+                expr
+            }
+        }
+    }
+}
+
+fn child_modifies_state(expr: &Expr, rules_to_exprs: &HashMap<String, Expr>) -> bool {
```
dragostis: Why not put `wrap_branch_exprs` before?

meta/src/ast.rs
```diff
@@ -58,15 +58,25 @@ pub enum Expr {
...
     Skip(Vec<String>),
     /// Matches an expression and pushes it to the stack, e.g. `push(e)`
-    Push(Box<Expr>)
+    Push(Box<Expr>),
+
+    /// Internal Exprs
+
+    /// If the wrapped expression errors, return the parser state to where it was prior to
+    /// running the closure
+    RestoreOnErr(Box<Expr>)
```
dragostis: I'm still not quite sure how the AST nodes should be dealt with. Since there are two types, I don't know if the internal compiler types (skip, restore, etc.) should be visible here or what we should do with them.

meta/src/ast.rs
```diff
@@ -201,10 +211,86 @@ impl Expr {
     }
 }
 
+pub struct ExprTopDownIterator {
```
dragostis: This is a nice way to solve this. I've been lazy with the optimizer implementations and went for hacky, functional-jerk type of implementation.

meta/src/ast.rs
```diff
@@ -201,10 +211,86 @@ impl Expr {
...
+    current: Option<Expr>,
+    next: Option<Expr>,
+    right_branches: Vec<Expr>
```
dragostis: Of course, this makes the implementation more complicated than some quick-and-dirty DFS that you can implement recursively.

jstnlef: You're not wrong :P. Took me a bit to get this working correctly.

meta/src/ast.rs
jstnlef: Using the build in iterator trait has a lot of benefits and would definitely prefer to use it where appropriate.

meta/src/ast.rs
jstnlef: This is a great question. I'm not entirely sure myself. I'm of 2 minds. On the one hand, it's incredibly useful to have the internal compiler types here since they will necessarily be a product of the optimizer. On the other hand, I'm not entirely sure they should be exposed directly since I don't think we'll have any plans of making any explicit syntax for restore.

jstnlef:
I'm not entirely sure what we'd like to do with this one. At this point, it is working as intended and the checkpoints are only inserted when they are needed. We could just merge this one with possible follow up work of optimizing the stack implementation and/or fixing up the optimizer API. What do you think?

dragostis:
Sorry for the late reply. Somehow, I must have missed this particular notification. 😞 Anyway, I agree with merging this as is, provided that performance is going well. Refactoring the optimizer API can be done later, maybe post 2.0.

jstnlef:
I can confirm that there is no hit in performance to grammars which do not use the stack mutation rules. However, I haven't tested a sufficiently complex grammar which has rules that would require the state checkpointing. Though, I argue, in those specific situations it makes sense for us to take the performance hit in favor of correctness.

dragostis:
I completely agree. I think the path that's worth following right now is to have a YAML indentation grammar implemented and see what performance looks like. Then we can optimise for what would be the normal use case.

dragostis:
Before merging this, I would also add some conformance tests in `grammar.pest` and see them pass both in `derive` and in `vm`.

jstnlef:
Completely agree. Was thinking of doing that just this morning. I'll do that tonight.

PR #232, opened 2018-04-28
Add drop to be validated as a keyword and predefined rule

jstnlef:


dragostis:
Maybe add a little notice in `generator.rs`:
```
 // Note that predefined rules should be pest keywords in meta/src/validator.rs.
```

jstnlef:
Factored out the predefined_rules generation and added the comment to that method

PR #233, opened 2018-04-29
Added a docstring and example to restore_on_err

jstnlef:
I actually also discovered a subtle bug. Originally, the stack would never be restored to an empty state since snapshotting requires that the ops vec was modified at least once prior to snapshot. To fix this, I pre-seed the snapshots with 0

### DIFF
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -916,8 +916,8 @@ impl<'i, R: RuleType> ParserState<'i, R> {
     ///
     /// let input = "ab";
     /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
-    /// let mut result = state.restore_on_err( |state| state.stack_push( |state|
-    ///     state.match_string("a")).and_then( |state| state.match_string("a"))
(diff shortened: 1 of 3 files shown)


pest/src/parser_state.rs
```diff
@@ -903,7 +903,29 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    ///
+    /// ```
+    /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "ab";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let mut result = state.restore_on_err( |state| state.stack_push( |state|
```
dragostis: No space before the closures.

jstnlef:
Fixed the formatting, ran a cargo fmt, and added a test to ensure appropriate stack behavior when restoring to empty

PR #235, opened 2018-04-30
Add some basic types to pest

jstnlef:
Issue: https://github.com/pest-parser/pest/issues/103

Added a bunch of builtin rules to pest. I decided to go with all caps for the syntax.

Let the bikeshedding of syntax commence!


dragostis:
I've thought a bit about this and I think we should go ahead with this. Breaking compatibility can be done smartly by releasing a beta and encouraging people to try it. Based on that feedback, we can decide whether or not this was a good call.

derive/src/generator.rs
```diff
@@ -69,66 +69,55 @@ pub fn generate(
 }
 
 // Note: All predefined rules should be validated as pest keywords in meta/src/validator.rs.
```
dragostis: Rename this to `builtin`.

derive/src/lib.rs
```diff
@@ -175,19 +175,30 @@
...
-//! * `comment` - gets run between rules and sub-rules
-//! * `any` - matches exactly one `char`
-//! * `soi` - (start-of-input) matches only when a `Parser` is still at the starting position
-//! * `eoi` - (end-of-input) matches only when a `Parser` has reached its end
-//! * `pop` - pops a string from the stack and matches it
-//! * `peek` - peeks a string from the stack and matches it
-//! * `drop` - drops the top of the stack (fails to match if the stack is empty)
-//!
-//! `whitespace` and `comment` should be defined manually if needed. All other rules cannot be
+//! * `WHITESPACE` - runs between rules and sub-rules
```
dragostis: I would convert this into two tables. First, the special rules, then the shorthands. This can be done in a separate PR, however.

derive/src/macros.rs
```diff
@@ -0,0 +1,46 @@
+// pest. The Elegant Parser
```
dragostis: Thank you for bringing sanity into this world! ❤️

jstnlef:
Addressed the review feedback

Issue #231, opened 2018-04-28
Add a `stack` rule


jstnlef:
We should have a `stack` rule which allows the parser to match the whole stack.

PR #237, opened 2018-05-08
Adding the PEEK_ALL and POP_ALL rules

jstnlef:
Addresses issue https://github.com/pest-parser/pest/issues/231.

### DIFF
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -885,12 +885,19 @@ impl<'i, R: RuleType> ParserState<'i, R> {
     /// assert_eq!(result.unwrap().position().pos(), 4);
     /// ```
     #[inline]
-    pub fn stack_match_peek(self: Box<Self>) -> ParseResult<Box<Self>> {
-        let string = self.stack.iter()
-            .map(|s| s.as_str())
-            .collect::<Vec<&str>>()
-            .concat();
-        self.match_string(&string)
+    pub fn stack_match_peek(mut self: Box<Self>) -> ParseResult<Box<Self>> {
+        let mut position = self.position.clone();
+        let result = self.stack.iter().all(|span| {
+            position.match_string(span.as_str())
+        });
+
+        self.position = position;
+
+        if result {
+            Ok(self)
+        } else {
+            Err(self)
+        }
     }
 
     /// Matches the full state of the stack. This method will clear the stack as it evaluates.
@@ -913,11 +920,23 @@ impl<'i, R: RuleType> ParserState<'i, R> {
     /// ```
     #[inline]
     pub fn stack_match_pop(mut self: Box<Self>) -> ParseResult<Box<Self>> {
-        let mut strings = Vec::new();
+        let mut position = self.position.clone();
+        let mut result = true;
         while self.stack.peek().is_some() {
-            strings.push(self.stack.pop().unwrap().as_str())
+            let span = self.stack.pop().unwrap();
+            result = position.match_string(span.as_str());
+            if !result {
+                break;
+            }
         }
-        self.match_string(&strings.concat())
+
+        self.position = position;
+
+        if result {
+            Ok(self)
+        } else {
+            Err(self)
+        }
     }
 
     /// Drops the top of the stack and returns `Ok(Box<ParserState>)` if there was a value to
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -891,9 +891,8 @@ impl<'i, R: RuleType> ParserState<'i, R> {
             position.match_string(span.as_str())
         });
 
-        self.position = position;
-
         if result {
+            self.position = position;
             Ok(self)
         } else {
             Err(self)
@@ -930,9 +929,8 @@ impl<'i, R: RuleType> ParserState<'i, R> {
             }
         }
 
-        self.position = position;
-
         if result {
+            self.position = position;
             Ok(self)
         } else {
             Err(self)
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -785,7 +785,7 @@ impl<'i, R: RuleType> ParserState<'i, R> {
     ///
     /// let input = "ab";
     /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
-    /// let mut result = state.stack_push( |state| state.match_string("a"));
+    /// let mut result = state.stack_push(|state| state.match_string("a"));
     /// assert!(result.is_ok());
     /// assert_eq!(result.unwrap().position().pos(), 1);
     /// ```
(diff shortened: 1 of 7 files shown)


pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    /// let mut result = state.stack_push( |state| state.match_string("a")).and_then(|state| {
+    ///     state.stack_push( |state| state.match_string("a"))
+    /// }).and_then(|state| state.stack_match_peek());
+    /// assert!(result.is_ok());
+    /// assert_eq!(result.unwrap().position().pos(), 4);
+    /// ```
+    #[inline]
+    pub fn stack_match_peek(self: Box<Self>) -> ParseResult<Box<Self>> {
+        let spans = self.stack.get_state();
+        let string = spans.iter()
```
dragostis: Since you're iterating anyway, why not use [`Iterator::all`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all)?

pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    /// ```
+    /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "aaaa";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let mut result = state.stack_push( |state| state.match_string("a")).and_then(|state| {
+    ///     state.stack_push( |state| state.match_string("a"))
```
dragostis: Extra space here.

pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    ///
+    /// ```
+    /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "aaaa";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let mut result = state.stack_push( |state| state.match_string("a")).and_then(|state| {
```
dragostis: Extra space.

pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    /// ```
+    /// /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "aaaa";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let mut result = state.stack_push( |state| state.match_string("a")).and_then(|state| {
+    ///     state.stack_push( |state| state.match_string("a"))
```
dragostis: Extra space.

pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    ///
+    /// ```
+    /// /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "aaaa";
+    /// let mut state: Box<pest::ParserState<Rule>> = pest::ParserState::new(input);
+    /// let mut result = state.stack_push( |state| state.match_string("a")).and_then(|state| {
```
dragostis: Extra space. Might be in a couple more places. :D

pest/src/stack.rs
```diff
@@ -54,6 +54,13 @@ impl<T: Clone> Stack<T> {
         popped
     }
 
+    /// Dump the full state of the stack
+    pub fn get_state(&self) -> Vec<T> {
+        let mut stack = self.cache.clone();
```
dragostis: Calling this too often is expensive. Why not return a `&Vec<T>` here and `iter().rev()` later?

pest/src/parser_state.rs
```diff
@@ -858,23 +858,67 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
+    ///     state.stack_push( |state| state.match_string("a"))
+    /// }).and_then(|state| state.stack_match_peek());
+    /// assert!(result.is_ok());
+    /// assert_eq!(result.unwrap().position().pos(), 4);
+    /// ```
+    #[inline]
+    pub fn stack_match_pop(mut self: Box<Self>) -> ParseResult<Box<Self>> {
+        let mut strings = Vec::new();
```
dragostis: Maybe there one could define `Stack::get_state_mut`, then apply [`Vec::drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain), then `Iterator::all`?

jstnlef: Drain won't work because we need to keep track of all the mutations to the stack which we get through the `pop` call.

pest/src/parser_state.rs
jstnlef: I guess I could run `self.match_string()` multiple times. I'm not sure if that's better or worse than producing a string and matching against that.

dragostis: Joining the strings call O(N) memcpy and O(1) memcmp. Iterating should be cheaper because you run memcmp O(N) times and that's it.

jstnlef: So the more I think about it, the more I like my initial solution. Mostly because repeatedly running `self.match_string()` involves wrapping and unwrapping subsequent `ParseResult`s. Also `match_string()` is consuming so I'm not entirely sure how to get `Iterator::fold` to work with it since the `self` is borrowed while iterating `self.stack`

dragostis: I agree with you. However, since this is implemented in `parser_state`, why not use `Position::match_string`? Then you could simply fold on a `bool`.

dragostis: Generally, I think that the solution is fine, but allocating a lot of small `Vec`s on the heap for every `peek_all` will slow down heap allocation after a while.

jstnlef:
Made some more changes to address review comments

pest/src/parser_state.rs
```diff
@@ -885,12 +885,19 @@ impl<'i, R: RuleType> ParserState<'i, R> {
...
-            .collect::<Vec<&str>>()
-            .concat();
-        self.match_string(&string)
+    pub fn stack_match_peek(mut self: Box<Self>) -> ParseResult<Box<Self>> {
+        let mut position = self.position.clone();
+        let result = self.stack.iter().all(|span| {
+            position.match_string(span.as_str())
+        });
+
+        self.position = position;
```
dragostis: I believe this needs some kind of mechanism to roll-back to the initial position if all the matches weren't successfull. This line should be in the if clause. Same with pop all.

jstnlef: Good point. I've made the change 👍

PR #243, opened 2018-05-26
[WIP] No std

### LINKED ISSUE #240: No std?
I assume this doesn't support no std. What work needs to be done to make it support it? I would be happy to do it with direction.

Restioson:
Fixes #240. 

TODO:
- [ ] Test for whether it works in no std

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -24,3 +24,6 @@ syn = "^0.13"
 codecov = { repository = "pest-parser/pest" }
 maintenance = { status = "actively-developed" }
 travis-ci = { repository = "pest-parser/pest" }
+
+[features]
+no_std = []
--- a/derive/src/macros.rs
+++ b/derive/src/macros.rs
@@ -19,26 +19,64 @@ macro_rules! insert_public_builtin {
     };
 }
 
+#[cfg(not(feature = "no_std"))]
 macro_rules! generate_rule {
     ($name: ident, $pattern: expr) => {
         quote! {
             #[inline]
             #[allow(dead_code, non_snake_case, unused_variables)]
-            fn $name(state: Box<::pest::ParserState<Rule>>) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
+            fn $name(
+                state: Box<::pest::ParserState<Rule>>
+            ) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
                 $pattern
             }
         }
     }
 }
 
-macro_rules! generate_public_rule {
+#[cfg(feature = "no_std")]
+macro_rules! generate_rule {
     ($name: ident, $pattern: expr) => {
         quote! {
             #[inline]
             #[allow(dead_code, non_snake_case, unused_variables)]
-            pub fn $name(state: Box<::pest::ParserState<Rule>>) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
+            fn $name(
+                state: alloc::boxed::Box<::pest::ParserState<Rule>>
+            ) -> ::pest::ParseResult<alloc::boxed::Box<::pest::ParserState<Rule>>> {
                 $pattern
             }
         }
     }
 }
+
+
+#[cfg(not(feature = "no_std"))]
+macro_rules! generate_public_rule {
+    ($name: ident, $pattern: expr) => {
+        quote! {
+            #[inline]
+            #[allow(dead_code, non_snake_case, unused_variables)]
+            pub fn $name(
+                state: Box<::pest::ParserState<Rule>>
+            )  -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
+                $pattern
+            }
+        }
+    }
+}
+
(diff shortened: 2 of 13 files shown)


pest/src/prec_climber.rs
```diff
@@ -83,7 +85,7 @@ impl<R: RuleType> BitOr for Operator<R> {
 /// [1]: https://en.wikipedia.org/wiki/Operator-precedence_parser#Precedence_climbing_method
 #[derive(Debug)]
 pub struct PrecClimber<R: RuleType> {
-    ops: HashMap<R, (u32, Assoc)>
+    ops: BTreeMap<R, (u32, Assoc)>
```
jstnlef: I'm curious about this change. It seems like this isn't relevant for no_std. Why the change?

Restioson: It is relevant, HashMap isn't in `alloc` (it uses rand or something).

Restioson: Maybe it could be `cfg`'d though, if it were any slower (is it? benching required)

dragostis: I think that the best approach here is to limit the `Vec` in the contractor to 256. (assert + document with `# Panics` section) Then use a `Box<[Assoc; 256]>` to store the `ops`.

derive/src/macros.rs
```diff
@@ -19,26 +19,64 @@ macro_rules! insert_public_builtin {
...
 
+#[cfg(not(feature = "no_std"))]
 macro_rules! generate_rule {
     ($name: ident, $pattern: expr) => {
         quote! {
             #[inline]
             #[allow(dead_code, non_snake_case, unused_variables)]
-            fn $name(state: Box<::pest::ParserState<Rule>>) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
+            fn $name(
+                state: Box<::pest::ParserState<Rule>>
```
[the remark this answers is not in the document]
Restioson: Yeah perhaps

derive/Cargo.toml
```diff
@@ -24,3 +24,6 @@ syn = "^0.13"
 codecov = { repository = "pest-parser/pest" }
 maintenance = { status = "actively-developed" }
 travis-ci = { repository = "pest-parser/pest" }
+
+[features]
+no_std = []
```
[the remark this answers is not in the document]
Restioson: Actually I probably can make them depend on eachother. I'll try that.

pest/src/prec_climber.rs
jstnlef: We'll definitely want to test to see if there's a notable difference. I suspect for most of our use-cases it'll be similar enough.

jstnlef:
Also looks like the build is broken. Seems like some changes need to happen to pest_vm.

Restioson:
Yeah, that's a bit weird. Strangely it Works On My Machine (as you can see from the screenshot), but I'll look at that soonish. Unfortunately I am writing exams right now and for the next few weeks so I may not be able to get to it that soon, but after exams I  will.

pest/src/lib.rs
```diff
@@ -56,6 +56,34 @@
...
+#[cfg_attr(feature = "no_std", macro_use)]
+extern crate alloc;
+
+#[cfg(feature = "no_std")]
+mod std {
+    pub use alloc::*;
+    pub use core::*;
+    pub use core::{fmt, str};
+
+    /// Cheap-ish knockoff of std::erorr. Just does what we need it to, nothing more.
```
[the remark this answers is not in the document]
Restioson: Hmm... How could we test?

pest/src/lib.rs
```diff
@@ -56,6 +56,34 @@
 
 #![doc(html_root_url = "https://docs.rs/pest")]
 
+#![cfg_attr(feature = "no_std", no_std)]
+#![cfg_attr(feature = "no_std", feature(alloc))]
+#![cfg_attr(feature = "no_std", feature(slice_concat_ext))]
+
+#![cfg(feature = "no_std")]
```
wirelyre: Looks like this exclamation mark should be removed.

wirelyre:
@dragostis Does Travis need to be updated to test both regular and `no_std` configurations?

dragostis:
@wirelyre Yes, but that shouldn't be a huge issue. I would simply add another `cargo build`/`cargo test` run for the `no_std` case. This should be part of this same PR, I'd say.

@Restioson Travis issue might stem from the fact that it's running the tests on stable and you might have a different configuration.

Restioson:
OK - I'm done exams so hopefully I'll be able to look at this soon.

sjmackenzie:
How's the status of this PR? Would be fantastic to have it in.

Restioson:
Hi! Unfortunately, it's been years since I've worked on this, and I imagine that both the rust no_std ecosystem as well as pest itself has evolved beyond this change. Therefore, I'll close this now as it'd probably be easier to implement this from scratch from HEAD than to update it from my 4 year old fork 😅 If someone else wants to, feel free to reuse any code from this or the approach, but personally I don't have the bandwidth to see this through right now.

PR #246, opened 2018-06-01
[WIP] Add regex terminal

### LINKED ISSUE #175: using Unicode Category
I want to do lexical analysis by the Unicode category.
Is there any way to get enough performance?

```
id          = id1 idn*
id1         = [\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}]
idn         = [\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\p{Cf}]
```

CAD97:
This is a bit messy, and probably needs more tests. However, `pest_derive/tests/grammar::regex` does prove that the feature is functional. Closes #175. The given grammar would be written:

```pest
id          = { id1 ~ idn* }
id1         = { r"[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}]" }
idn         = { r"[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\p{Cf}]" }
```

Note that the author recommends the following [UAX31-R1](https://www.unicode.org/reports/tr31/#R1)-compliant grammar for identifiers instead:

```pest
identifier = { r"\p{XID_START}\p{XID_CONTINUE}*" }
```

A more complete modification-resiliant [UAX31-D1](https://www.unicode.org/reports/tr31/#D1) grammar is left as an exercise to the reader.

If everything is working as intended, publicly facing changes should be as follows:

- public dependency on `regex:^1` added
- when using a regex terminal,
  - peer dependency on `regex:^1` added
  - peer dependency on `lazy_static:^1` added
  - lazy static called `REGEXES` is added during the derive
- regex terminals are now available!
  - syntax: raw string like Rust's
    - that is, `"r" ~ push("#"*) ~ (!("\""~peek) ~ any)* ~ "\"" ~ pop`
  - the regex crate is used to test the given regex against the input
  - provides access to Unicode Properties [as per the regex crate documentation](https://github.com/rust-lang/regex/blob/master/UNICODE.md#rl12-properties)
- added `match_regex(Box<self>, &Regex)` method to `pest::ParserState`
  - (that's where the public dependency on `regex` comes in)

Any further publicly-visible changes are probably a bug in my implementation. Now accepting pointers on where I should add tests, any functionality I've missed (do I need to do something for `pest_vm`?), or stupid (or style) errors.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -21,6 +21,10 @@ proc-macro2 = "^0.4.4"
 quote = "^0.6.3"
 syn = "^0.14.1"
 
+[dev-dependencies]
+regex = "1"
+lazy_static = "1"
+
 [badges]
 codecov = { repository = "pest-parser/pest" }
 maintenance = { status = "actively-developed" }
--- a/derive/src/generator.rs
+++ b/derive/src/generator.rs
@@ -22,8 +22,12 @@ pub fn generate(
     defaults: Vec<&str>
 ) -> TokenStream {
     let uses_eoi = defaults.iter().any(|name| *name == "EOI");
+    let uses_regex = rules.iter().any(|rule|
+        if let OptimizedExpr::Regex(_) = rule.expr { true } else { false }
+    );
 
     let builtins = generate_builtin_rules();
+    let regexes = generate_regexes(uses_regex);
     let rule_enum = generate_enum(&rules, uses_eoi);
     let patterns = generate_patterns(&rules, uses_eoi);
     let skip = generate_skip(&rules);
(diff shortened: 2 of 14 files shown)


CAD97:
For the curious:

- The test identifier is the one given in [UAX31-A1](https://www.unicode.org/reports/tr31/#A1) as an example without ZWNJ of why accepting ZWNJ in some cases may be desirable.

Further optimization steps that may be desirable:

- transform simple regexps into regexless pest
  - transform literal regexps (those without regex control characters) into pest literal strings
  - note that `+` and `*` regex repeaters _cannot_ be translated into pest due to pest's greedy behavior where regex takes all available paths

CAD97:
Closing in favor of #247

PR #247, opened 2018-06-02
Add Unicode Property builtins

### LINKED ISSUE #175: using Unicode Category

CAD97:
This adds a large number of builtin rules to the pest parser that can be used without definition. The full list is visible in `pest_meta::UNICODE_PROPERTY_NAMES`. These rules are only added to your parser when used, and thus the tables should be optimized out as dead code when not required.

Serves as an alternative to #246 and closes #175. The example given in that issue would now be written as:

```pest
id  = @{ id1 ~ idn* }
id1 = { UPPERCASE_LETTER | LOWERCASE_LETTER | TITLECASE_LETTER | MODIFIER_LETTER | OTHER_LETTER | LETTER_NUMBER }
idn = { id1 | NONSPACING_MARK | SPACING_MARK | DECIMAL_NUMBER | CONNECTOR_PUNCTUATION | FORMAT }
```

Note, however, that the author of this PR recommends the following [UAX31-R1](https://www.unicode.org/reports/tr31/#R1)-compliant identifier grammar instead:

```pest
identifier = @{ XID_START ~ XID_CONTINUE* }
```

A full [UAX31-D1](https://www.unicode.org/reports/tr31/#D1) identifier grammar is left as an exercise to the reader.

Please note that all `#[doc(hidden)]` public items are hidden because they are not considered part of the public surface area of the pest crates, and may change at any time.

As a summary of the publicly facing changes:

- Added a large number of new builtin rules for binary Unicode Properties and General_Category membership.
- Added `ParserState::match_char_by(Box<self>, impl FnOnce(char)->bool)` to match character-wise via a filter function.
- Added a large number of hidden items to thread info about the unicode properties through the `pest` workspace without duplicating it. _These are not public stable API._


CAD97:
Note that 8_387 of the added loc are generated tables. They are committed to avoid requiring users to download and unzip the UCD and have `ucd-generate` locally.

CAD97:
I realize coverage is important, but codecov, please. The file causing the lower coverage is [`pest::unicode`](https://codecov.io/gh/pest-parser/pest/src/eba9ea04d3da0adbd6bbf0146cdf397b0fe7dedb/pest/src/unicode/mod.rs). That's really just a lot of boilerplate wrapping up `ucd-trie` for exposure. I could add unit tests but they seem unnecessary, and tests at the pest->parser level would just distract from the not-just-the-same-thing-over-and-over.

CAD97:
Also, since it was brought up on the issue (but with the caveat that CI times are flaky):

- This PR: 9min 40s
- master: 10min 58s

I think that shows that this doesn't noticeably affect the compile times of this crate.

CAD97:
Observation: All preexisting (2.0) builtins (modulo `skip`) are `SHOUT_CASE`. Currently, the Unicode Properties are added as `Title_Case`, which is the way that they are presented by the Unicode Consortium (though they recommend loose matching which includes case insensitivity). We don't have any real reason to support more than one way to refer to the builtin rules, so we can chose whatever casing we want.

It would be trivial to change the case from `Title_Case` to `SHOUT_CASE` if that is desired for uniformity with other builtins. Personally, I'd prefer sticking to the source `Title_Case`, but either is a valid choice

derive/src/generator.rs
```diff
@@ -119,6 +120,18 @@ fn generate_builtin_rules() -> HashMap<&'static str, TokenStream> {
             .or_else(|state| state.match_string("\r"))
     );
 
+    for property in ::pest_meta::UNICODE_PROPERTY_NAMES {
```
dragostis: The leading `::` in `pest_meta` and `syn` should not be necessary.

meta/src/validator.rs
```diff
@@ -102,6 +102,7 @@ pub fn validate_pairs<'i>(pairs: Pairs<'i, Rule>) -> Result<Vec<&'i str>, Vec<Er
     builtins.insert("ALPHANUMERIC");
     builtins.insert("ASCII");
     builtins.insert("NEWLINE");
+    builtins.extend(::UNICODE_PROPERTY_NAMES);
```
dragostis: Are the `::` necessary here? I think `UNICODE_PROPERTY_NAMES` can simply be `use`d.

pest/src/lib.rs
```diff
@@ -74,6 +76,8 @@ pub mod prec_climber;
 mod span;
 mod stack;
 mod token;
+#[doc(hidden)]
+pub mod unicode;
```
dragostis: This path should be added to [codecov.yml](https://docs.codecov.io/docs/ignoring-paths).

meta/src/validator.rs
CAD97: Sorry, that's habit; I've been writing too many macros recently where you can't really import names.

pest/src/lib.rs
CAD97: 👍 It's been done

pest/src/position.rs
```diff
@@ -274,6 +274,25 @@ impl<'i> Position<'i> {
         false
     }
 
+    /// Matches the char at the `Position` against a filter function and returns `true` if a match
+    /// was made. If no match was made, returns `false` and `pos` will not be updated.
+    pub(crate) fn match_char_by<F>(&mut self, f: F) -> bool
```
jstnlef: Might be worth annotating `#[inline]` for consistency. That said, I would have to do some profiling to see if there is a significant performance difference.

pest/src/parser_state.rs
```diff
@@ -433,6 +433,19 @@ impl<'i, R: RuleType> ParserState<'i, R> {
         }
     }
 
+    /// Asks the `ParserState` to match a single character based on a filter function.
+    /// If the match is successful, this will return an `Ok` with the updated `Box<ParserState>`.
+    /// If failed, an `Err` with the updated `Box<ParserState>` is returned.
```
jstnlef: Should probably include an example here.

pest/src/position.rs
CAD97: I _think_ `#[inline]` doesn't have any effect for a `pub(crate) fn`. If I recall correctly, all `#[inline]` does is provide the required metadata cross-crate to inline the function. And as it's generic, that already happens anyway to provide monomorphization. It _might_ also hint LLVM that this is likely a good target for inlining, but LLVM is able to do whatever it pleases with that hint.

That said, consistency is a good thing, and the hint doesn't hurt, so I've gone ahead and added it.

pest/src/parser_state.rs
CAD97: I've added an example using `char::is_ascii`. It could be implemented better as a builtin by just checking the first byte, but as an example it's easy to understand.

wirelyre:
I really like this change. Seems right in line with the project goals.

---

The huge list of built-in rules makes me wonder if they should be namespaced somehow — the example `XID_Start` and `XID_Continue` are easy to distinguish from user-defined rules because of their prefixes.

I think it would be valuable to have a simple, noisy visual indication of a built-in rule. Especially since there are now so many. For that reason, I weakly support one of
- prefix, *e.g.* `UC_Uppercase_Letter`
- shouting, *e.g.* `UPPERCASE_LETTER`
- special syntax, *e.g.* `UC(Uppercase_Letter)` ([in the spirit of `\p{…}` syntax](https://unicode.org/reports/tr18/#Categories))

Nevertheless, `Title_Case` seems like the objectively correct syntax, and no one appears to be using capitalized rules in the wild anyway. This is a bikeshed.

pest/src/unicode/mod.rs
```diff
@@ -0,0 +1,284 @@
+//! Character inclusion in binary or General_Category value Unicode sets.
+//!
+//! We rely on dead code elimination to remove the tables that aren't needed.
+
+#![allow(bad_style)]
+
+#[allow(unused)]
+mod binary;
+pub fn ASCII_Hex_Digit(c: char) -> bool {
```
wirelyre: This module might be cleaner to read and update as a macro? Something like:

```rust
char_properties! {
    self::binary::ASCII_HEX_DIGIT => ASCII_Hex_Digit,
    self::binary::ALPHABETIC => Alphabetic,
    …
    self::category::UPPERCASE_LETTER => Uppercase_Letter,
}
```

CAD97: That makes a lot of sense. I originally did it by copy/pasting the generated BY_NAME constants and doing regex replace in-editor, but doing the transformation via macro makes for a better presentation. 👍

I'd like to make the macro also spit out `UNICODE_PROPERTY_NAMES`, but that constant needs to be available to meta and derive, whereas the functions are required by the runtime.

CAD97: I plan to write up a script to update the Unicode tries once I figure out the best way to use `ucd-generate` as a library that doesn't add to much maintenance burden to BurntSushi. (Likely it will be caving and just saying "install it".) Hopefully I can also spit out the mod.rs as well.

CAD97:
If we go for demarcated syntax a la `\p{PROPERTY}`, I'd expect the contents of the demarcation to adhear to [UAX44-LM3](https://unicode.org/reports/tr44/#UAX44-LM3) and accept the short aliases.

> Property aliases and property value aliases are symbolic values. When comparing them, use loose matching rule UAX44-LM3.
> 
> UAX44-LM3. Ignore case, whitespace, underscore ('_'), hyphens, and any initial prefix string "is".
> 
> - "linebreak" is equivalent to "Line_Break" or "Line-break"
> - "lb=BA" is equivalent to "lb=ba" or "LB=BA"
> - "Script=Greek" is equivalent to "Script=isGreek" or "Script=Is_Greek"

This is probably a lot more work that desired, though [ucd-util](https://github.com/BurntSushi/ucd-generate/tree/master/ucd-util) (sibling to ucd-generate and ucd-trie) would make the job of canonicalizing a bit simpler.

Though more immediately, this would also require adding a new rule class, rather than just piggybacking on top of the existing builtin rule plumbing.

----

For further consideration, here's a list of all builtins:

<details><summary>pest builtins</summary>

Keywords

- ANY
- DROP
- EOI
- PEEK
- PEEK_ALL
- POP
- POP_ALL
- PUSH
- skip
- SOI

Non-keyword Builtin

- DIGIT
- NONZERO_DIGIT
- BIN_DIGIT
- OCT_DIGIT
- HEX_DIGIT
- ALPHA_LOWER
- ALPHA_UPPER
- ALPHA
- ALPHANUMERIC
- ASCII
- NEWLINE

Magic Symbols

- WHITESPACE
- COMMENT

Binary Unicode Properties

 - ASCII_Hex_Digit
 - Alphabetic
 - Bidi_Control
 - Case_Ignorable
 - Cased
 - Changes_When_Casefolded
 - Changes_When_Casemapped
 - Changes_When_Lowercased
 - Changes_When_Titlecased
 - Changes_When_Uppercased
 - Dash
 - Default_Ignorable_Code_Point
 - Deprecated
 - Diacritic
 - Extender
 - Grapheme_Base
 - Grapheme_Extend
 - Grapheme_Link
 - Hex_Digit
 - Hyphen
 - IDS_Binary_Operator
 - IDS_Trinary_Operator
 - ID_Continue
 - ID_Start
 - Ideographic
 - Join_Control
 - Logical_Order_Exception
 - Lowercase
 - Math
 - Noncharacter_Code_Point
 - Other_Alphabetic
 - Other_Default_Ignorable_Code_Point
 - Other_Grapheme_Extend
 - Other_ID_Continue
 - Other_ID_Start
 - Other_Lowercase
 - Other_Math
 - Other_Uppercase
 - Pattern_Syntax
 - Pattern_White_Space
 - Prepended_Concatenation_Mark
 - Quotation_Mark
 - Radical
 - Regional_Indicator
 - Sentence_Terminal
 - Soft_Dotted
 - Terminal_Punctuation
 - Unified_Ideograph
 - Uppercase
 - Variation_Selector
 - White_Space
 - XID_Continue
 - XID_Start

Unicode General Category Membership

 - Cased_Letter
 - Close_Punctuation
 - Connector_Punctuation
 - Control
 - Currency_Symbol
 - Dash_Punctuation
 - Decimal_Number
 - Enclosing_Mark
 - Final_Punctuation
 - Format
 - Initial_Punctuation
 - Letter
 - Letter_Number
 - Line_Separator
 - Lowercase_Letter
 - Mark
 - Math_Symbol
 - Modifier_Letter
 - Modifier_Symbol
 - Nonspacing_Mark
 - Number
 - Open_Punctuation
 - Other
 - Other_Letter
 - Other_Number
 - Other_Punctuation
 - Other_Symbol
 - Paragraph_Separator
 - Private_Use
 - Punctuation
 - Separator
 - Space_Separator
 - Spacing_Mark
 - Surrogate
 - Symbol
 - Titlecase_Letter
 - Unassigned
 - Uppercase_Letter

</details><br/>

There's only one full duplicate between preexisting builtins and the Unicode-defined ones: HEX_DIGIT and ASCII_Hex_Digit. The existing ones should probably be marked as ASCII-only in a perfect world -- and sometimes that's desired (e.g. `u*::parse` only works on `/[0-9]+/`, not `/\p{Decimal_Number}+/`).

CAD97:
For a dead simple rule for builtin-vs-user-defined, we could say "all builtins will start with an uppercase letter" and "user defined rules are suggested to start with a lowercase letter to avoid confusion with builtins". An aggressive stance would be to enforce that.

But 🚲🏠 is fun and easy. As the used forms in Title_Case are the "canonical" form published by the Unicode Consortium, I'm in favor of keeping them as just the proper Title_Case or using a simple demarcation similar to \p{Title_Case}, but if we go the route of demarcation, we should support the short aliases.

jstnlef:
Personally, I think I prefer all-caps for all the built in rules (looks like we missed skip which should also be all-caps). To that end, we can demarcate the unicode categories as `UC(Title_Case)` or `U(Title_Case)`. This will allow the properties to be represented as defined in the spec (no surprises). Also, we wouldn't necessarily need to support the short aliases immediately. That can be left as future work if necessary. I am also for properly prefixing the names of the ASCII only builtins with `ASCII` to be more descriptive of which characters those rules will match against.

jstnlef:
I'm also not against using ucd-util to make the process easier. Seems like a pretty small and useful library.

CAD97:
I'm not sure exactly how I'd go about adding a demarcated syntax. Obviously I'd need to edit the grammar, but every attempt I've made so far has dominoed into further-reaching changes than I'm really comfortable with.

So the current status as I see it:

- If the Unicode Properties are demarcated with "start" and "stop" symbols (such as `U(` and `)`), then we should canonicalize the names.
  - If someone can add "namespacing" support (minimally just for builtin rules), I can do the Unicode-specific part.
- If the Unicode Properties do not have special announcement tokens, they should still be separated from userland rules visually.
  - This is partially achieved by `Title_Case` or `SHOUT_CASE`.
    - Prior 2.0 builtins are all `SHOUT_CASE`
      - Except `skip` which should be `SKIP` anyway
  - Should they be separated from the other builtins?
    - Prefacing the ASCII-specific rules with `ASCII_` is probably the right move here.
    - Unicode Property builtins could be prefixed with `U_` or `UNICODE_` for maximum verbosity.
      - Or, it could be decided that Unicode Properties are `Title_Case`, other builtins `SHOUT_CASE` and user rules `snake_case`.
  - All other things being equal, `Title_Case` would be objectively better than `SHOUT_CASE` for the Unicode Properties, as that is their canonical format.

dragostis:
I think that a particularity of pest is that it is a biased implementation of a parser library. And to continue on the same path, it makes sense to build on the same foundation, with the ultimate goal of being as concise and intuitive as possible. I think that the current implementation is the most straight-forward and that adhering to Unicode's property naming scheme is not at all necessary as long as the functionality is usable.

Ultimately, prepending some `UNICODE_` (or `PROP_`) so `SHOUT_CASE` builtins sounds like the most obvious choice. @jstnlef, @wirelyre, any thoughts on this approach?

CAD97:
I think my personal preference (assuming free reign) would be:

- SHOUT_CASE the Unicode builtin rules
  - (Remove ASCII_HEX_DIGIT -- it's really wasteful of space)
- Rename existing ASCII builtins to `ASCII_<name>`
- Un-shout `WHITESPACE` and `COMMENT`
  - These aren't builtins, they're special rules with special semantics
  - Maybe `Whitespace` and `Comment`? I'm not sure.

The idea is to make using the Unicode rules easier by default, but still provide the ASCII rules for when they're desired.

However, I'm still open to prefixing the Unicode rules as well. Maybe we can have the following builtins for clarity and consistency:

<details><summary>Builtins</summary>

Keywords:  
- `ANY`/`SKIP`
- `SOI`/`EOI`
- `PUSH`/`POP`/`POP_ALL`/`PEEK`/`PEEK_ALL`/`DROP`/`DROP_ALL`

ASCII Builtin Rules:
- `ASCII_DIGIT`/`ASCII_NONZERO_DIGIT`/`ASCII_BIN_DIGIT`/`ASCII_OCT_DIGIT`/`ASCII_HEX_DIGIT`
- `ASCII_ALPHA`/`ASCII_UPPERCASE`/`ASCII_LOWERCASE`
- `ASCII_ALPHANUMERIC`
- `ASCII_SYMBOL`
- `ASCII_NEWLINE`

Rules with Special Semantics:
- `WHITESPACE`
- `COMMENT`

Generated Unicode Property
- `PROP_ALPHABETIC`
- `PROP_BIDI_CONTROL`
- `PROP_CASE_IGNORABLE`
- (etc)

Generated Unicode Category
- `CAT_CASED_LETTER`
- `CAT_CLOSE_PUNCTUATION`
- `CAT_CONNECTOR_PUNCTUATION`
- (etc)

</details><br/>

(Obviously subject to a vote on the appropriate prefixes for each category)

CAD97:
Just created and used an update script for the tables. It doesn't touch `pest::unicode::mod` or `pest_meta::UNICODE_PROPERTY_NAMES`, but the actual tables can be regenerated from the most recent UCD data with `update_unicode.sh` from $CARGO_WORKSPACE_DIR.

dragostis:
@CAD97 Sorry for the late reply. Yes, I agree with this approach. However, I see whitespace/comment rules as being empty but overridable which is why to me it makes sense for them to be shout case.

PR #255, opened 2018-08-15
pest_derive does not work with quote = "0.3.14" so bump dep

Eh2406:
In testing [`criterion` with Cargos `-Z minimal-versions`](https://github.com/japaric/criterion.rs/issues/183) I notestd that `pest_derive` does not work with quote = "<=0.3.14". This PR bumps the minimal acceptable versions in Cargo.toml to 0.3.15, so as to match what you are really compatible with.

This gets `pest.*` working with Cargos -Z minimal-versions. This is part of the process of seeing how hard minimal-versions is for crates to use in preparation for getting it stabilized for use in CI. It is easy to use if all of your dependencies support it, but much harder if trying to impose it on them.

### DIFF
--- a/pest_derive/Cargo.toml
+++ b/pest_derive/Cargo.toml
@@ -15,9 +15,9 @@ name = "pest_derive"
 proc-macro = true
 
 [dependencies]
-quote = "^0.3"
-syn = "^0.11"
-pest = { path = "../pest", version = "^1.0" }
+quote = "0.3.15"
+syn = "0.11"
+pest = { path = "../pest", version = "1.0" }
 
 [badges]
 codecov = { repository = "pest-parser/pest" }
--- a/README.md
+++ b/README.md
@@ -146,8 +146,8 @@ pest requires [Cargo and Rust 1.23](https://www.rust-lang.org/en-US/downloads.ht
 Add the following to `Cargo.toml`:
 
 ```toml
-pest = "^1.0"
-pest_derive = "^1.0"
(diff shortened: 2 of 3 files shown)


pest_derive/Cargo.toml
```diff
@@ -15,9 +15,9 @@ name = "pest_derive"
 proc-macro = true
 
 [dependencies]
-quote = "^0.3"
-syn = "^0.11"
-pest = { path = "../pest", version = "^1.0" }
+quote = "0.3.15"
```
dragostis: Why not `"^0.3.15"`?

Eh2406: Cargo default to using `^`. So both formulations are equivalent. If you would prefer the other form I can change it.

dragostis: Oh, I see. Then, could you please change all of them?

Eh2406: I removed all the usages of `^` where they are redundant. It is possible that I misunderstood you, and you wanted me to add them in where they are implied. If so I am sorry, and I can fix it the other way if you ask.

dragostis: There are two other Cargo files that have dependencies in meta and vm.

Eh2406: This PR is against the 1.0 branch which does not have those folders.

dragostis: My bad! 😅

PR #269, opened 2018-08-31
Added preliminary SIMD optimization.

dragostis:
Currently only kicks in in skip_until sequences, but this opens doors to a series of similar approaches.

Implementation relies on packed_simd which is why it is hidden behind a feature flag. Once this is stabilized and implemented in std, this feature can be removed.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -14,6 +14,9 @@ readme = "_README.md"
 name = "pest_derive"
 proc-macro = true
 
+[features]
+no_simd = []
+
 [dependencies]
 pest = { path = "../pest", version = "2.0" }
 pest_meta = { path = "../meta", version = "2.0" }
--- a/derive/src/generator.rs
+++ b/derive/src/generator.rs
@@ -553,6 +553,26 @@ fn generate_expr_atomic(expr: OptimizedExpr) -> TokenStream {
             }
         }
         OptimizedExpr::Skip(strings) => {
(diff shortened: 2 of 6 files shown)


CAD97:
Shouldn't the feature be additive?

```toml
[features]
default = ["simd"]
simd = ["packed_simd"]
```

Then s/feature = "no_simd"/not(feature = "simd")/g and s/not(feature = "no_simd")/feature = "simd"/g

dragostis:
I feel like it would be a good idea to have it opt-out, rather than opt-in. The only downsides are slightly slower compile time until `std::simd` is stable and the fact that it *might* not compile on every conceivable platform right now.

jstnlef:
Seems to be having an issue building on stable in travis.

CAD97:
That still is opt-out, as the feature is on by default.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section

> With the exception of the `default` feature, all features are opt-in. To opt out of the default feature, use `default-features = false` and cherry-pick individual features.

CAD97:
But yes, `packed_simd` is still nightly-only, so it needs to be opt-in if we continue supporting stable (which we should!)

dragostis:
Oh, I see. I was running nightly locally—was actually pretty impressed that this was building on stable. Seems like I'll have to do opt-out until it's stable.

dragostis:
Actually, this can probably live on this branch a little while longer; having nightly-only features is additional maintenance that we don't need right now.

PR #277, opened 2018-09-15
Added initial debugger implementation.

### LINKED ISSUE #98: Grammar backtrace
Currently, if you want to figure out what the parser was expecting when it failed, you call the parser's expected() method which tells you which rules could have been next and provides a position which can be referenced in the source. 

Unfortunately, when debugging a grammar, it can be difficult to use that information alone to figure out what went wrong. Especially in large grammars, knowing the next expected token doesn't always give you enough of the picture to figure out what went wrong. The expected token may be used in many places. 

You can make an educated guess from the given source position, but it is hard to know for sure if your assumptions can be trusted. 

It would be convenient if, when returning an error, the parser could generate a "backtrace" through the grammar which would inform us of exactly which path through the rules failed. 

This is not a cheap calculation, so it should be saved for error cases and even possibly only done behind a feature flag which doesn't even compile that code unless the user adds a features section in their Cargo.toml.

dragostis:
Fixes #98.

This implementation is quite rudimentary:

* relies on threads instead of stack-full generators
* parsing threads that have not reached their end-of-input will not be joined
* commands parsing is rough and brittle; e.g. it doesn't support paths that contain white space
* the readline's formatting could use some improvement
* only shows position from the `ParserState` when it reaches a breakpoint
* no way to run another rule from a breakpoint, only from the start

I'm not completely convinced that we should merge this. But I think it makes for a decent proof-of-concept. Maybe someone would be interested in further work here? @hansihe, this should be good enough as-is for tracing 2.0 grammars.


hansihe:
Thanks, this certainly helps a lot! I'm just going to braindump some more ideas I have

* It would be nice if it could be made to spit out every single rule it enters. Even though this produces a huge amount of output, it's often very useful for finding the right starting point for further investigation.
* For both automation and quickly rerunning commands it would be nice if the point above, along with just general breakpoint debugging could be started directly from the command line. I could imagine something like this `pest_debugger run --rule top --grammar grammar.pest --input indata.txt --breakpoint number --breakpoint digit`. It could then drop you into the interactive shell when it hits a breakpoint. This makes it easy to run things repeatedly, and shortens the debug feedback loop.

I also hit https://github.com/pest-parser/pest/issues/279, which might be related to the debugger.

dragostis:
By the way, if you need these improvements soon, you could fork my own PR and work on a separate PR. I'm happy to close this one in favor of that.

dragostis:
@hansihe, any news regarding your progress?

tomtau:
Made some improvements in this PR: https://github.com/pest-parser/pest/pull/736

PR #283, opened 2018-09-24
Advertise the book/landing page

measlytwerp:
Hope the wording is acceptable.


dragostis:
Thanks for the PR. I'd gladly accept this, but I'm probably going to do a bigger overhaul anyway so it doesn't make sense to merge this.

PR #284, opened 2018-09-27
Clean up documentation

wirelyre:
The most prominent changes are:

* Much briefer docs in the main module (that fit on one screen!)
* Removed item kinds ("this is a `mod` which…")
* Fixed descriptions in `iterators` which referred to old types

Hopefully the reduced noise makes it easier to see what *isn't* documented.

I think that a logical next step is to document `iterators/mod.rs` with an explanation similar to [the "Parser API" chapter of the book](https://github.com/pest-parser/book/blob/0cc17d4a7be3d9e3b3aca5125d0facbdf76b2b36/src/parser_api.md) (except denser and more boring). From there, only `ParserState` remains.

Once that's all done, the landing page for the `pest` crate should link to

1. `pest_derive` docs (which it already does)
2. the `iterators` module
3. the book


pest/src/error.rs
```diff
@@ -18,7 +18,7 @@ use RuleType;
 use position::Position;
 use span::Span;
 
-/// A `struct` defining errors.
+/// The error type for a failed parse.
```
dragostis: I would use "parse-related" in order to include custom error messages in the definition which might be strictly failed parse errors.

pest/src/error.rs
```diff
@@ -173,11 +173,14 @@ impl<R: RuleType> Error<R> {
...
     }
 
-    /// Renames all `Rule`s from a `ParsingError` `variant`. It does nothing when called on
-    /// `CustomError` `varriant`.
+    /// Renames all `Rule`s if this is a [`ParsingError`]. It does nothing when called on a
+    /// [`CustomError`].
     ///
     /// Useful in order to rename verbose rules or have detailed per-`Rule` formatting.
     ///
+    /// [`ParsingError`]: enum.ErrorVariant.html#variant.ParsingError
```
dragostis: I wonder if intra-rustdoc links were stabilized?

pest/src/position.rs
```diff
@@ -16,8 +16,7 @@ use std::str;
 
 use span;
 
-/// A `struct` containing a position that is tied to a `&str` which provides useful methods to
-/// manually parse it.
+/// A character position in a `&str` which provides useful methods to manually parse that string.
```
dragostis: I guess this should explain the fact that the `Position` is actually a pointer that points *between* two characters and not a character itself.

dragostis:
Also, I don't think that `ParserState` should be a priority for the 2.0. Not more than precedence climbing anyway.

pest/src/error.rs
CAD97: I believe you can link between symbols in scope with ``[`symbol`]`` barring a few strange corner cases.

wirelyre: That's what I thought, but I can't make it work on my machine. Maybe some part of my toolchain is outdated?

wirelyre: Nope, looks like it works on nightly only. Guess we're stuck with explicit URLs for now.

PR #289, opened 2018-10-01
Generator crate to vendor bootstrap

CAD97:
`pest_generator` is the previous `pest_derive`, with `proc_macro2::TokenStream` instead of `proc_macro::TokenStream`, and as a normal library. `pest_derive` is a trivial `proc_macro` wrapper around `pest_generator`.

This should be able to break the bootstrap cycle by "precompiling" the meta pest file into a Rust file, so that pest is no longer required to build pest. ([Proof of concept](https://github.com/CAD97/pest-vendor))


generator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "pest_generator"
+description = "pest's derive macro"
```
dragostis: This should be updated to something like 'pest code generator'.

generator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version = "2.0.0"
+authors = ["Dragoș Tiselice <dragostiselice@gmail.com>"]
+homepage = "https://pest-parser.github.io/"
+repository = "https://github.com/pest-parser/pest"
+documentation = "https://docs.rs/pest"
+keywords = ["pest", "parser", "peg", "grammar"]
```
dragostis: Should probably be `["pest", "generator"]`.

generator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+categories = ["parsing"]
+license = "MIT/Apache-2.0"
+readme = "_README.md"
+
+[lib]
```
dragostis: Does it need name here?

generator/src/lib.rs
```diff
@@ -0,0 +1,431 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. All files in the project carrying such notice may not be copied,
+// modified, or distributed except according to those terms.
+
+//! # pest. The Elegant Parser
```
dragostis: I wouldn't add this documentation here. Can be empty until we get to document all internal crates.

generator/Cargo.toml
CAD97: I just included it since I was modifying `pest_derive` which has a lib.name set. I'll remove it here since it's redundant though.

generator/src/lib.rs
CAD97: Alright, I've removed it for now, it was just duplicated from the derive crate.

generator/Cargo.toml
dragostis: Looks like I did a poor job of keeping Cargo files clean. 😓

CAD97:
The `pest-vendor` POC is _very_ rough. I'm uncertain if it even needs to exist; we can instead have a simple script to do the vendoring, and that'd be part of the update to actually break the bootstrap cycle anyway.

I doubt anyone else needs to vendor, and everyone definitely should prefer to use the derive and incremental compilation.

dragostis:
I agree that the likelihood of reusing this externally is quite small. Sure. We can do something similar in a small bootstrap crate.

PR #291, opened 2018-10-01
Generator explicit `include_str!("grammar.pest")` flag

CAD97:
Merge and then `pest_generator` should be good to publish for the bootstrap script


generator/src/generator.rs
```diff
@@ -22,12 +22,17 @@ pub fn generate(
     generics: &Generics,
     path: &Path,
     rules: Vec<OptimizedRule>,
-    defaults: Vec<&str>
+    defaults: Vec<&str>,
+    include_grammar: bool,
```
dragostis: Would normally not use trailing commas.

generator/src/generator.rs
```diff
@@ -22,12 +22,17 @@ pub fn generate(
...
 ) -> TokenStream {
     let uses_eoi = defaults.iter().any(|name| *name == "EOI");
 
     let builtins = generate_builtin_rules();
-    let include_fix = generate_include(&name, &path.to_str().expect("non-Unicode path"));
+    let include_fix = if include_grammar {
+        generate_include(&name, &path.to_str().expect("non-Unicode path"))
+    } else {
+        quote!();
```
dragostis: Should this be a statement?

generator/src/generator.rs
CAD97: Habit from rustfmt style; I'll stick to the file's

generator/src/generator.rs
CAD97: Yes it should 😅

PR #293, opened 2018-10-02
Added small clarifications to unsafe use.

dragostis:


dragostis:
I've already benchmarked this when I first started. UTF-8 boundary checks are expensive, but slice boundary checks are not. In small benchmarks, boundary checks are noticeable (~5% improvement on my machine), but `&str` slicing has `O(s.len())` complexity instead of `O(1)`.

CAD97:
`&str` slicing is _not_ `O(n)`, it's still `O(1)`, just with a higher constant.

A key feature of UTF8 is that you can tell if you're on a character boundary by looking at the single next byte. The algorithm is, [basically](https://github.com/rust-lang/rust/blob/8b7f164eab538025a29488e2e47ae7b4bbee2ad9/src/libcore/str/mod.rs#L488):

```rust
fn is_utf8_head_byte(s: &u8) -> bool {
    (*s & 0b11_000_000) != 0b10_000_000
}
```

This does require loading the contents as opposed to a slice that doesn't check, however, which might be where you're seeing a larger difference. (Still, I'd like to see benchmarks present to back this unsafe up.)

dragostis:
Closing this in favor of #303.

Issue #301, opened 2018-10-03
Reinforce invariants on Position


dragostis:
`Position` currently uses `unsafe` in quite a few places. Maybe there is a better way to model this in order to minimize the use of `unsafe` and strengthen its invariants.

wisn:
Is this still going, @dragostis? I'm asking because there is a merged PR already

dragostis:
@wisn, thanks for the ping. I think this is fine. If @CAD97 is fine with the current state of affairs, we can close this.

CAD97:
#383 is a case where one of the expensive correctness checks (gated on debug assertions) is causing unacceptable performance degredation in debug mode (therefore making it useless). (When I added it, I thought/hoped it wasn't on a hot path.)

I think we need to make that invariant obvious on the type so that this sanity check can be removed, and then we can consider this closed.

All that needs to happen performance wise is an `unsafe` path that's used when the invariant is guaranteed (on pain of nasal demons) to hold. A type combining the queue Rc and the input ref to guarantee that the queue refers to that input seems the way to go.

PR #303, opened 2018-10-03
Removed position::new and replaced it with position::Position::new,

jkarns275:
Addresses issue #301 . 
Replaced unsafe code with Options and unwrapping where appropriate (which is everywhere unsafe code was being used). 

I've also changed the position::new function to position::Position::new, since it is now safe. Moreover, I've encountered an issue in my own parser project - I wanted to use the position API to find the line + column of an error but couldn't because I can't create my own position object. This pull request should fix that.


pest/src/position.rs
```diff
@@ -17,16 +17,33 @@ use std::str;
 use span;
 
 /// A cursor position in a `&str` which provides useful methods to manually parse that string.
+#[derive(Clone)]
 pub struct Position<'i> {
     input: &'i [u8],
     pos: usize
 }
 
-pub unsafe fn new(input: &[u8], pos: usize) -> Position {
```
CAD97: Just want to note that removing this fn is a breaking change. 2.0 would have to keep it around in `#[deprecated]` format.

jkarns275: Where else does it get used? I thought that because it is a private API removing it would be alright (as long as all references to it in the library itself were changed, which I think is the case).

CAD97: My bad I thought it was exported. (Can we please `pub(crate)` some things as we move to Edition2018)

CAD97:
I think this is the wrong approach -- you're taking `&[u8]` and immediately trying to turn it into `&str`. Just take `&str`!

As I mentioned in the linked issue's linked PR, I think the "correct" representation is to encode the UTF8 guarantee by storing `&str` and then the unsafety is on those slices which benchmarking shows benefit from it.

And then we have something to show the informal wg that's aiming to eliminate the need to write unsafe code for performance and not ffi.

pieguy275:
Most places where that use Position::new / position::new are working with byte slices. Do you think that is also something that should be changed?

jkarns275:
Do you mean that only Position should take &str, or that the other places where Position::new is called e.g. Tokens should also use &str?

Also, could something like this work instead of `str::from_utf8(&input[pos..]).is_err()`:

`input[pos] & 0b11_000_000 != 0b10_000_000`

I think it should, unless my interpretation of your message [here](https://github.com/pest-parser/pest/pull/293#issuecomment-426147271) is incorrect.

CAD97:
If you have a `&str` (that is, a `&[u8]` that you know is valid UTF8`), you can slice it by just checking the borders of where you're slicing for being a UTF8 leading byte. This is what the `&str` slicing does for you already.

You shouldn't need to implement this manually, just storing `&str` and slicing it does it for you.

If you already know you're on a character boundary, then you can slice without checking (or have the optimizer realize that you don't need to check, but that's one of the... more difficult ones to provide, and why the `Chars` iterator offers a conversion back into `&str`).

CAD97:
https://github.com/rust-lang/rust/blob/4cf11765dc98536c6eedf33f2df7f72f6e161263/src/libcore/str/mod.rs#L1860-L1868

```rust
        fn get(self, slice: &str) -> Option<&Self::Output> {
            if self.start <= self.end &&
               slice.is_char_boundary(self.start) &&
               slice.is_char_boundary(self.end) {
                Some(unsafe { self.get_unchecked(slice) })
            } else {
                None
            }
        }
```

jkarns275:
Oh I see, that makes sense. Still the trouble is, most (if not all) code that uses Position is also using a byte slice. Do you think it would be worth changing this?

pest/src/position.rs
```diff
@@ -17,16 +17,34 @@ use std::str;
...
+    /// # Examples
+    /// ```
+    /// # use pest::Position;
+    /// let cheart = '💖';
+    /// let heart = "💖";
+    /// assert_eq!(Position::new(heart, 1), None);
+    /// assert_ne!(Position::new(heart, cheart.len_utf8()), None);
+    /// ```
+    pub fn new(input: &str, pos: usize) -> Option<Position> {
+        if input.len() < pos || input.get(pos..).is_none() {
```
CAD97: This is now redundant. You can just do

```rust
input.get(pos..).map(|input| Position { input, pos })
```

because if `pos > input.len()`, `input.get(pos..)` is going to return `None`.

jkarns275: Alright I removed the redundant code. Still experiencing some bug(s?) that is causing the calculator tests to fail, even though unit tests pass.

CAD97:
```
error: proc-macro derive panicked
  --> grammars/src/lib.rs:22:14
   |
22 |     #[derive(Parser)]
   |              ^^^^^^
   |
   = help: message: byte index 56 is not a char boundary; it is inside 'ș' (bytes 55..57) of `// pest. The Elegant Parser
```

This is... not good. I don't spot any algorithm changes (though I may have missed something), so that may mean we're slicing into characters in the current algorithm. That scares me, because several places don't check that the backing buffer is valid when called safely.

All the more reason to explore changing as many backing buffers to `&str` as possible, and maybe even avoiding unsafe slices/conversion in debug builds.

jkarns275:
Yes, I've been making sure the algorithms are identical for several hours now to fix this bug (and related bugs). If you run the tests in the pest/pest crate the calculator example fails, and it is simple enough such that the bug may be easy to track down so thats where I'm looking now.

jkarns275:
Is there a way to see a backtrace in a proc-macro panic? I couldn't find anything but it sure would be useful here.

dragostis:
Take into account that slicing in `Position` is not always at UTF-8 borders. When matching a string, for example, it's perfectly safe to slice at non-UTF-8 borders for the purpose of comparing to a guaranteed UTF-8 string.

jkarns275:
The bug was in Parser::match_range, but in the process of fixing it I replaced `&[u8]` with `&str` everywhere in the pest/pest crate

dragostis:
How does this impact performance? There's a benchmark in grammars that you can run.

jkarns275:
Unsurprisingly it tanked performance:

`18,067 ns/iter (+/- 1,598)` vs `24,590 ns/iter (+/- 1,729)`

I'll look into profiling the code tomorrow. I feel like similar performance is possible

CAD97:
> it's perfectly safe to slice at non-UTF-8 borders for the purpose of comparing to a guaranteed UTF-8 string

I assume you're talking about where the code was

```rust
        let matched = {
            // Matching is safe since, even if the string does not fall on UTF-8 borders, that
            // particular slice is only used for comparison which will be handled correctly.
            let slice = unsafe { str::from_utf8_unchecked(&self.input[self.pos..]) };
            if slice.is_char_boundary(string.len()) {
                let slice = unsafe { slice.get_unchecked(0..string.len()) };
                slice.eq_ignore_ascii_case(string)
            } else {
                false
             }
        };
```

If `&self.input[self.pos..]` here is not valid UTF8, this is _still Very Bad_! This is benign in that it's only UB in theory; every function we call on the invalid string (`str::is_char_boundary` and `str::eq_ignore_ascii_case`) _is currently implemented_ by immediately swapping back to viewing it as a byte slice, and completely ignores the front of the string. For that reason though, it would be better to _just do the operation completely within bytes land ourselves._ That way we're not at the mercy of safe functions continuing to (unguaranteed) ignore that they're requiring UTF8 well-formedness.

To use some young terminology from internals, this str is in a _valid_ but _unsafe_ state. As such, we are only guaranteed to be able to rely on _documented guarantees_ of code.

pest/src/position.rs
```diff
@@ -321,12 +332,8 @@ impl<'i> Position<'i> {
     #[inline]
     pub(crate) fn match_insensitive(&mut self, string: &str) -> bool {
         let matched = {
-            // Matching is safe since, even if the string does not fall on UTF-8 borders, that
-            // particular slice is only used for comparison which will be handled correctly.
-            let slice = unsafe { str::from_utf8_unchecked(&self.input[self.pos..]) };
-
-            if slice.is_char_boundary(string.len()) {
-                let slice = unsafe { slice.get_unchecked(0..string.len()) };
+            let slice = &self.input[self.pos..];
```
CAD97: This whole function should (as noted above) work in bytes. The whole thing should probably have a go-over to see where our "cursor" is currently pointing inside of a Unicode codepoint to determine if there's a different way to handle it. (And benchmark along the way.)

```rust
fn match_insensitive(&mut self, string: &str) -> bool {
    let matched = {
        let slice = &self.input.as_bytes()[self.pos..];
        if let Some(slice) = slice.get(0..string.len()) {
            slice.eq_ignore_ascii_case(&string.as_bytes())
        } else {
            false
        }
    };
    if matched {
        self.pos += string.len();
    }
    matched

**Comment by CAD97** [comment]
Another scary bit:

```rust
            // Cannot actually cause undefined behavior.
            let slice = unsafe { str::from_utf8_unchecked(&self.input[self.pos..]) };
            if let Some(c) = slice.chars().next() {
```

This one is _certified Very Bad™️_. If `self.pos` isn't on a character boundary here, bad things can happen _with today's implementation_!

```
impl<'a> Iterator for Chars<'a> {
    fn next(&mut self) -> Option<char> {
        next_code_point(&mut self.iter).map(|ch| {
            // str invariant says `ch` is a valid Unicode Scalar Value
            unsafe {
                char::from_u32_unchecked(ch)
            }
        })
    }
```

I hope you see the problem here. For a _very abusive, one off_ benchmark comparison I relied on `chars()` to not overrun a random byte buffer if I terminated it with four null bytes and ignored the output. I realize now that even then I was pushing it (though at least it was before the niche filling optimization), as _this could return invalid `char` values_ if you give it invariant-breaking `&str` input.

If we aren't guaranteed that `self.pos` is on a character boundary, _this function as is defined on master is latent UB_. I'm being overly agressive, sure, but that's because breaking guarantees of standard library types _is a big deal_ and can break things in _very unrelated ways_.

@jkarns275, after you feel this is in a decent place, I'll go over and attempt a full audit of where slicing is done and what guarantees we're keeping. It may mean (re)introducing some `unsafe` for `slice_unchecked` and friends to match performance, but strong documentation of our use of `unsafe` and a greater certainty of correctness is definitely worth it.

jkarns275:
@CAD97 Though I'm not done optimizing, I've managed to improve the performance - it actually runs faster than the current master branch does (for me at least).

jkarns275:
Okay, @CAD97 I'm all set looking for optimizations and bugs. 

There are some unsafe blocks that could be replaced by unwraps, not sure what the performance impact is though.

CAD97:
Actually, hold that thought.

```
error: proc-macro derive panicked
  --> grammars\src\lib.rs:22:14
   |
22 |     #[derive(Parser)]
   |              ^^^^^^
   |
   = help: message: 💥 UNSAFE `Pair` CREATED 💥
```

😬

CAD97:
<details><summary>Benchmark comparisons</summary>

### CAD97/unsafe-audit

```
D:\usr\Documents\Code\Rust\pest>cargo +nightly bench --package=pest_grammars
    Finished release [optimized] target(s) in 0.11s
     Running target\release\deps\pest_grammars-84bc40a485e21838.exe

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target\release\deps\json-62e2e79657492ab0.exe

running 1 test
test data ... bench:      14,691 ns/iter (+/- 576)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured; 0 filtered out
```

### pest-parser/master

```
D:\usr\Documents\Code\Rust\pest>cargo +nightly bench --package=pest_grammars
    Finished release [optimized] target(s) in 0.12s
     Running target\release\deps\pest_grammars-84bc40a485e21838.exe

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target\release\deps\json-62e2e79657492ab0.exe

running 1 test
test data ... bench:      16,786 ns/iter (+/- 644)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured; 0 filtered out
```

</details><br/>

Current master consistently benchmarks >16k ns/iter, my branch consistently benchmarks at <15k ns/iter, so it's improving performance! 🎉

CAD97:
The final question is whether we want `Span::new`/`Position::new` to return `Option<Self>` or panic. Either are a valid choice.

wirelyre:
The convention in the standard library seems to be that constructors should not panic.

See [`NonZeroU8`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html), which has a safe `fn new(...) -> Option<Self>` and `unsafe fn new_unchecked(...)`.

Most constructors that can fail are not named `new`, however.

dragostis:
@wirelyre, I agree. I think `Option<Self>` constructors should work great.

pest/src/position.rs
```diff
@@ -410,7 +409,7 @@ impl<'i> Ord for Position<'i> {
 
 impl<'i> Hash for Position<'i> {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        (self.input as *const [u8]).hash(state);
+        (self.input.as_bytes() as *const [u8]).hash(state);
```
dragostis: I believe this approach only hashes the first hash of the `&str`/fat pointer.

```suggestion
        (self.input as *const str).hash(state);
```

pest/src/span.rs
```diff
@@ -176,7 +201,7 @@ impl<'i> Eq for Span<'i> {}
 
 impl<'i> Hash for Span<'i> {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        (self.input as *const [u8]).hash(state);
+        (self.input.as_bytes() as *const [u8]).hash(state);
```
dragostis: I believe this approach only hashes the first hash of the `&str`/fat pointer.

```suggestion
        (self.input as *const str).hash(state);
```

pest/src/span.rs
```diff
@@ -18,18 +18,49 @@ use position;
...
+    ///
+    /// # Safety
+    ///
+    /// `input[start..end]` must be a valid subslice; that is, said indexing should not panic.
+    pub(crate) unsafe fn new_unchecked(input: &str, start: usize, end: usize) -> Span {
+        debug_assert!(input.get(start..end).is_some());
+        Span { input, start, end }
+    }
+
+    /// Attempts to create a new span. Will return None if `input[start..end]` is an invalid index
```
dragostis:
```suggestion
    /// Attempts to create a new span. Will return `None` if `input[start..end]` is an invalid index
```

pest/src/position.rs
dragostis: Is this something we want to do before merging this?

pest/src/position.rs
```diff
@@ -260,10 +289,15 @@ impl<'i> Position<'i> {
     #[inline]
     pub(crate) fn skip_until(&mut self, strings: &[&str]) -> bool {
         for from in self.pos..self.input.len() {
-            for slice in strings.iter().map(|s| s.as_bytes()) {
-                let to = from + slice.len();
+            let bytes = if let Some(t) = self.input.get(from..) {
```
dragostis:
```suggestion
            let bytes = if let Some(string) = self.input.get(from..) {
```

pest/src/position.rs
```diff
@@ -260,10 +289,15 @@ impl<'i> Position<'i> {
...
+                t.as_bytes()
```
dragostis:
```suggestion
                string.as_bytes()
```

pest/src/position.rs
```diff
@@ -183,23 +214,23 @@ impl<'i> Position<'i> {
                 None => 0
             }
         };
-
-        let end = if self.input.is_empty() {
+        let input_bytes = self.input.as_bytes();
```
dragostis: Maybe call this `bytes` for consistency?

pest/src/iterators/pairs.rs
```diff
@@ -255,7 +255,7 @@ impl<'i, R: Eq> Eq for Pairs<'i, R> {}
 impl<'i, R: Hash> Hash for Pairs<'i, R> {
     fn hash<H: Hasher>(&self, state: &mut H) {
         (&*self.queue as *const Vec<QueueableToken<R>>).hash(state);
-        (self.input as *const [u8]).hash(state);
+        (self.input.as_bytes() as *const [u8]).hash(state);
```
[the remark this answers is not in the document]
CAD97: (Disclaimer: implementation is a Rust implementation detail)

```rust
    #[stable(feature = "rust1", since = "1.0.0")]
    impl<T: ?Sized> Hash for *const T {
        fn hash<H: Hasher>(&self, state: &mut H) {
            if mem::size_of::<Self>() == mem::size_of::<usize>() {
                // Thin pointer
                state.write_usize(*self as *const () as usize);
            } else {
                // Fat pointer
                let (a, b) = unsafe {
                    *(self as *const Self as *const (usize, usize))
                };
                state.write_usize(a);
                state.write_usize(b);
            }
        }
    }
```

The observed effect is that [either approach will hash both address and length](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=a0eb5d197b361908e36b1243b1c67091). If you just want to hash the address, use str::as_ptr.

Which is the better approach here? I agree that if the same behavior is desired, `as *const str` is better than `.as_bytes() as *const [u8]`, but this is a good time to clarify which is desired.

pest/src/parser_state.rs
```diff
@@ -86,7 +86,8 @@ where
                     positives: state.pos_attempts.clone(),
                     negatives: state.neg_attempts.clone()
                 },
-                unsafe { position::new(input.as_bytes(), state.attempt_pos) }
+                // TODO(performance): Guarantee state.attempt_pos is a valid position
```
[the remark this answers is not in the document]
CAD97: This one is me.

I mean this TODO to be guarantee as in go through and find the manipulators of attempt_pos and make sure that all of them are making sure to set it to a character boundary, which I haven't done. That's why I changed it to the safe implementation rather than the unsafe one for now.

Honestly, this is probably cold enough that just using the unwrap is fine.

pest/src/position.rs
CAD97: No, this implementation is good as is for now.

Future profiling could definitely take a pass through here for what functions are slicing strings and could be slicing bytes instead, though.

CAD97:
Triage since I was going through the issues:

@jkarns275, this is good to merge if you address the inline concerns! I'd love to see this merged; it's a win for safe code and performance!

pest/src/position.rs
jkarns275: Fixed this [here](https://github.com/pest-parser/pest/pull/303/commits/5024c4029391e2eb554f16bb704d4e6fe3e507ff#diff-082ec1879c86e477c78290d490f0e6f6R217

PR #310, opened 2018-10-11
Add clippy into CI

luizdepra:
Hey. This solves #295.

But, clippy is throwing tons of warnings and it breaks the CI build with the gigantic size of the log.
Also, I can change the clippy command to fail the build on the first warning: `cargo clippy -- -D warnings`

Should I fix all warnings in this PR?

Issue #295, opened 2018-10-03
Run clippy on CI


dragostis:
clippy should run on Travis.

PR #312, opened 2018-10-12
Fix clippy errors

CAD97:
Prerequisite to enabling clippy on CI (#295, #310).

cc @luizdepra (feel free to pull this into #310)

### DIFF
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -17,6 +17,7 @@ use pest_meta::ast::*;
 use pest_meta::optimizer::*;
 use pest_meta::UNICODE_PROPERTY_NAMES;
 
+#[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
 pub fn generate(
     name: Ident,
     generics: &Generics,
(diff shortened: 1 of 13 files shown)


CAD97:
Actually, this changes some `pub` fn, so someone needs to go through and check whether these are exported (thus breaking changes to change) and allow for those instead. I just naively did the `&Vec<...>`->`&[...]` transformation for those.

CAD97:
The actually breaking changes (just did a quick check):

- `pest_meta::validator::validate_rust_keywords`
- `pest_meta::validator::validate_pest_keywords`
- `pest_meta::validator::validate_already_defined`
- `pest_meta::validator::validate_undefined`
- `pest_meta::validator::validate_ast`

We have three options: bump the `pest_meta` version, change them back and allow the lint, or change them back, allow the lint, deprecate them, and make new versions that take slices.

dragostis:
Deprecating is a bit unfortunate since you'll have two almost identical versions in the docs, but I guess it makes most sense.

luizdepra:
@CAD97, ok. Waiting the merge.

CAD97:
Wait we don't have bors set up 😆 I'll merge when green manually then

PR #313, opened 2018-10-12
Set up bors-ng 

CAD97:
https://bors.tech/

Requires the bors integration to have access to the organization to do its work; I've requested permission on its behalf. This is merely a suggestion, but I've had great success using bors on personal projects even without multiple PRs in flight at the same time, even just as a merge-when-green bot.

### DIFF
--- a/.bors.toml
+++ b/.bors.toml
@@ -1,0 +1,3 @@
+status = [
+    "continuous-integration/travis-ci/push",
+]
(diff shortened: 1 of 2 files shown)


CAD97:
bors: ping

It's alive!

| Syntax | Description
|--------|------------
| bors r+ | Run the test suite
| bors r=[list] | Same as r+, but the "reviewer" will come from the argument
| bors r- | Cancel an r+ or r=
| bors try | Run the test suite without pushing
| bors delegate+ | Allow the pull request author to r+
| bors delegate=[list] | Allow the listed users to r+
| bors ping | Will respond if bors is set up

Issue #300, opened 2018-10-03
Migrate to 2018 edition


dragostis:
We should probably migrate a bit before it is released on stable.

SaadiSave:
Any update on this?

PR #315, opened 2018-10-13
New Travis matrix

### LINKED ISSUE #294: Enforce rustfmt
rustfmt styling should be enforced on PRs by being run on Travis.

### LINKED ISSUE #295: Run clippy on CI

CAD97:
This PR has been limited to the Travis matrix update, with most jobs allowed to fail temporarily.
I will resubmit the fmt, clippy, and misc. test improvements changes removed from this PR momentarily as a new PR.

-----

Original text:

Closes #294, closes #295, supersedes #310

This has a rather large diff since I did a `cargo fmt` across the whole workspace, but it should enable mindless rustfmt from here on. I can tweak the `rustfmt.toml` as well if there are other [configurations](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md) we want to change; I've deliberately pinned rustfmt to a nightly version.

Clippy is pinned to 1.29.0 as that's the first rust toolchain to contain clippy-preview. As we'd like to take advantage of the new edition eventually, #300, I'm thinking that we won't officially commit to old compiler versions until 1.31 and edition 2018. At that point we can probably pin both rustfmt and clippy to the final 1.31 nightly to reduce the number of build jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,47 @@
 sudo: required
 language: rust
 cache: cargo
+
 os:
   - linux
-rust:
-  - stable
-  - nightly
+
 matrix:
+  include:
+    - name: stable
+      rust: stable
+      before_script: cargo bootstrap
+      script:
+        cargo build --all --verbose &&
+        cargo test --all --verbose &&
+        cargo doc --all --verbose
+    - name: style
+      rust: 1.31.0 # Pinned warnings
+      install:
+      - rustup component add clippy
+      - rustup component add rustfmt
+      before_script: cargo bootstrap
+      script:
+      - cargo clippy --all -- -D warnings
+      - cargo fmt --all -- --check
+    - name: minimal versions
+      rust: nightly-2018-12-08 # Arbitrary nightly for unstable cargo feature -- make this MSRV build later?
+      before_script:
+        - cargo -Z minimal-versions generate-lockfile
+        - cargo bootstrap
+      script: cargo test --all --verbose
+    - name: coverage
+      rust: nightly-2018-12-08 # Update this when updating tarpaulin
+      env: RUSTFLAGS="--cfg procmacro2_semver_exempt"
+      install: cargo install cargo-tarpaulin --version 0.6.10
+      before_script: cargo bootstrap
+      script: cargo tarpaulin --out Xml
+      after_success: bash <(curl -s https://codecov.io/bash)
   allow_failures:
-    - rust: nightly
+    - name: style # TODO(CAD97): Fix this build and require it
+    - name: minimal versions # Remove once it passes :D
+    - name: coverage # So builds don't have to wait on it
   fast_finish: true
+
 addons:
   apt:
     packages:
@@ -26,20 +58,3 @@ branches:
     - master
 notifications:
   email: false
-install:
-  - rustup component add clippy-preview
-script:
-  # Need a custom script because we are using cargo workspaces.
-  - |
-      cargo bootstrap && # dynamic build dependency of pest_meta
-      cargo clippy --all --verbose -- -D clippy &&
-      cargo build --all --verbose &&
-      cargo test --all --verbose &&
-      cargo doc --all --verbose
-after_success:
-  - |
-      if [[ $TRAVIS_RUST_VERSION =~ nightly ]]; then
-        RUSTFLAGS="--cfg procmacro2_semver_exempt" cargo install cargo-tarpaulin
-        cargo tarpaulin --out Xml &&
-        bash <(curl -s https://codecov.io/bash)
-      fi
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,7 +1,1 @@
-unstable_features = true
-condense_wildcard_suffixes = true
-imports_indent = "Block"
-reorder_imported_names = true
-reorder_imports = true
-reorder_imports_in_group = true
-trailing_comma = "Never"
+newline_style = "Unix"


CAD97:
Easy links:

- [The fmt diff](https://github.com/pest-parser/pest/pull/315/commits/bec7a2268a5ffdb740e583755c3c15529f86ec08)
- [Actual enforcement of rustfmt/clippy](https://github.com/pest-parser/pest/pull/315/commits/2c98ad5c43108b66298e3ea7e5649c7f42c2645c)

dragostis:
@CAD97, I think we should definitely bump the version, but I'm not quite sure when. 3.0 should definitely, but I'm not sure whether we want 2.1 to already require 2018 or not.

CAD97:
We can add a minimal-version job to this as well; I'm just choosing the best targets for clippy/rustfmt currently.

CAD97:
bors: try

meta/src/optimizer/restorer.rs
```diff
@@ -133,33 +127,37 @@ mod tests {
...
+                Box::new(Str(String::from("a")))
+            )
+        }];
 
         let restored = OptimizedRule {
             name: "rule".to_owned(),
             ty: RuleType::Normal,
             expr: Choice(
                 Box::new(RestoreOnErr(Box::new(Push(Box::new(Str(String::from(
-                    "a"
```
dragostis: What an eye-sore. Maybe this should be store in a variable just to make things less ugly?

pest/src/iterators/pairs.rs
```diff
@@ -287,12 +290,12 @@ mod tests {
         assert_eq!(
             format!("{:?}", pairs),
             "[\
-                Pair { rule: a, span: Span { str: \"abc\", start: 0, end: 3 }, inner: [\
-                    Pair { rule: b, span: Span { str: \"b\", start: 1, end: 2 }, inner: [] }\
-                ] }, \
-                Pair { rule: c, span: Span { str: \"e\", start: 4, end: 5 }, inner: [] }\
-            ]"
-        .to_owned()
+             Pair { rule: a, span: Span { str: \"abc\", start: 0, end: 3 }, inner: [\
```
dragostis: Maybe this should be `rustfmt::skip`?

CAD97: Definitely agree that the string formatting is useful here.

CAD97:
Just noting here that we'll probably want to bump the nightly version used for rustfmt somewhat regularly. It definitely should be pinned so that the build doesn't break unexpectedly but it also probably shouldn't fall too far behind until rustfmt is actually somewhat stable.

Firing off another try to make sure nothing upstream broke these added checks:

bors: try

meta/src/optimizer/restorer.rs
CAD97: Honestly, it might be worth it to make a very-simple-to-translate local macro_rules to write these test cases in. Honestly, just something to box up everything would be helpful.

I made a limited version of that in 72dc062e793d98c1b3a123d348bd68908fbcf8ec to show it off, this is very much a your call thing though.

CAD97:
Rebased on top of the clippy PR! I also took the liberty of squashing the history, as it wasn't being very useful and I was rebasing anyway.

meta/src/optimizer/restorer.rs
dragostis: Sure, definitely! In pursuit of thoroughness, I often lose track of code quality within tests.

dragostis:
I agree on updating rustfmt. However, this won't need to be done too often as 1.0 should arrive soon enough.

CAD97:
@dragostis Alright, `box_tree!` has cleaned up the optimizer tests.

Now, lets...

bors: try

generator/src/generator.rs
```diff
@@ -679,7 +676,8 @@ mod tests {
                         state.match_string("d")
                     })
                 })
-            }.to_string()
+            }
```
dragostis: Is there a way to refactor this to something more manageable?

meta/src/optimizer/restorer.rs
dragostis: Any update on this?

pest/src/error.rs
```diff
@@ -457,7 +459,8 @@ mod tests {
                 "  |  ^---",
                 "  |",
                 "  = unexpected 4, 5, or 6; expected 1, 2, or 3",
-            ].join("\n")
+            ]
```
dragostis: Same with these. I guess it would be best to use something better than `join`ing here.

meta/src/optimizer/restorer.rs
CAD97: Yes, the [macro exists](https://github.com/pest-parser/pest/pull/315/files#diff-daba655efcad6f906c92c455f29e0e14R14) and is used elsewhere in the optimizer, I just missed this file 😅

I've realized just now that other crates might also want a `box_tree!` for their tests, namely `generator/generator.rs` at least. I can add that in a later PR, I'd like to get this one merged as soon as realistic to avoid growing stale since it (necessarily) touches so much, tbh.

generator/src/generator.rs
CAD97: For the contents of `quote!`? Not really, not while maintaining the "obvious correctness" of what's being tested against. For the fact `to_string` is on its own line? Unless we write a routine to compare `TokenStream` directly, this is the best we'll get. Of course, the great thing about rustfmt is not having to worry about minor formatting things and just asking rustfmt to do it for you.

(It formats within `mac!(..)` and treats it like a normal function call, but within `mac! { .. }` you're on your own for formatting.)

pest/src/error.rs
CAD97: The "other" option:

```rust
        assert_eq!(
            format!("{}", error),
            " --> 2:2
  |
2 | cd
  |  ^---
  |
  = unexpected 4, 5, or 6; expected 1, 2, or 3",
        );
```

dragostis:
Yes. I think it's a good idea to merge this as-is and work from here once we merge #353 as well.

CAD97:
Alright, I'm rebasing and fixing up the travis matrix so there'll be some force pushes and Travis action in the near future.

dragostis:
I guess we can also have clippy and rustfmt run on the stable build now?

CAD97:
@dragostis We're still using unstable rustfmt options (the big one being `trailing_comma = "Never"`), so unless you want to drop those then we need to keep rustfmt on nightly.... but for some reason nightly isn't installing rustfmt/clippy currently.

I also still like to have the style check on a separate job from the tests, since that makes it easier to tell what failed at a glance (and it's not like we're duplicating work this way), and since that allows pinning (since at the very least `-D warnings` is never guaranteed to be stable). But yeah, I'll set those to stable at least for now, we can always switch it later if desired.

CAD97:
bors: r+

I've stripped back this PR to _only_ the Travis changes to unblock other CI-related work immediately, and am resubmitting the repo-wide fmt and style changes in new PRs.

PR #316, opened 2018-10-13
Added Error line/column info.

### LINKED ISSUE #307: Getting position/span from Error
handlebars is using the line/col number from pest `Error` for error reporting. From 2.0, it seems no way to get from the `Error` struct. I suggest to hold the `Position` or `Span` within `InputLocation` and provide getters for them.

dragostis:
Fixes #307. @sunng87, is this a fine solution?

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -25,11 +25,11 @@ pub struct Error<R> {
     pub variant: ErrorVariant<R>,
     /// Location within the input string
     pub location: InputLocation,
+    /// Line/column within the input string
+    pub line_col: LineColLocation,
     path: Option<String>,
     line: String,
     continued_line: Option<String>,
-    start: (usize, usize),
-    end: Option<(usize, usize)>
 }
 
 /// Different kinds of parsing errors.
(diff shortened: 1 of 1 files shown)


pest/src/error.rs
```diff
@@ -230,11 +237,17 @@ impl<R: RuleType> Error<R> {
...
+    }
+
     fn spacing(&self) -> String {
-        let line = if let Some((line, _)) = self.end {
-            cmp::max(self.start.0, line)
-        } else {
-            self.start.0
+        let line = match self.line_col {
+            LineColLocation::Pos((line, _)) => line,
+            LineColLocation::Span(_, (line, _)) => cmp::max(self.start().0, line)
```
CAD97: Wouldn't this be better written as

```rust
LineColLocation::Span((head, _), (tail, _)) => cmp::max(head, tail)
```

dragostis: Definitely. It slipped through. Feel free to push the fix and merge if you have the time. I'm going to be travelling until Wednesday.

dragostis: Probably call them `start_line`, `end_line` though?

sunng87:
Thanks! This looks good for my requirement.

It just seems a little verbose to have both `InputLocation` and `LineColLocation` under `Error`. Semantically I think the function of `LineColLocation` should belong to `InputLocation`.

dragostis:
@sunng87, how would you define the `InputLocation` data structure?

sunng87:
Perhaps we can have a light-weighted `Position` struct that stores `pos`, `line_no` and `col_no`. `InputLocation` is still an enum of `Pos` and `Span`, but each holding the light weight `Position` so we can get enough information . This breaks API compatibility though. 

Anyway, this patch fixes it well while maintained backward compatibility and I'm OK with it overall. We can refactor it during next major release.

dragostis:
bors r=CAD97

sunng87:
Cool! Could you please create a release for this?

dragostis:
Done. :)

PR #318, opened 2018-10-17
Pairs::peek

CAD97:
How does `next_back` work? Would a `peek_back` be useful to have (for completeness?)?

### DIFF
--- a/pest/src/iterators/pairs.rs
+++ b/pest/src/iterators/pairs.rs
@@ -167,6 +167,7 @@ impl<'i, R: RuleType> Pairs<'i, R> {
         tokens::new(self.queue, self.input, self.start, self.end)
     }
 
+    /// Peek at the first inner `Pair` without changing the position of this iterator.
     #[inline]
     pub fn peek(&self) -> Option<Pair<'i, R>> {
         if self.start < self.end {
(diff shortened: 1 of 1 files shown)


pest/src/iterators/pairs.rs
```diff
@@ -167,6 +167,15 @@ impl<'i, R: RuleType> Pairs<'i, R> {
         tokens::new(self.queue, self.input, self.start, self.end)
     }
 
+    #[inline]
+    pub fn peek(&mut self) -> Option<<Self as Iterator>::Item> {
```
dragostis: Why use `<Self as Iterator>::Item` instead of `Pair`?

```suggestion
    pub fn peek(&mut self) -> Option<Pair> {
```

dragostis: Should probably have docs with an example.

CAD97:
bors: r=dragostis

CAD97:
Whoops on the signature 😅 
Blame mobile that I didn't realize that CI had _already_ reported that...

bors: r=dragostis

CAD97:
bors: cancel

Actually, going to add minimal docs real quickly since I've got it pulled up

CAD97:
... And fix a huge oversight that I had made `peek` take `&mut self` instead of `&self`

CAD97:
Alright, this time we should be good. Good enough to merge, we can iterate on the docs from here.

bors: r=dragostis

PR #321, opened 2018-10-25
Generate dummy grammar include in release mode too.

### LINKED ISSUE #320: Changes in grammar file not picked up by cargo for rebuild
This is probably not possible to solve within pest... it would require a build.rs-like API to add dependency files.

But maybe a hint in the documentation may make sense?

birkenfeld:
The conditional was probably added because it was in the original
hint to do this manually in the documentation.  There doesn't seem
to be a downside since the constant will be thrown away in any
case, being unused.

Fixes #320.


birkenfeld:
Done.

PR #331, opened 2018-11-04
Add json5-rs to Projects using pest

callum-oakley:
pest is a real pleasure to work with by the way :)


dragostis:
Thanks for the PR! Excited to see pest enabling your work! :)

bors r+

PR #336, opened 2018-11-10
Add conversion Pair -> Pairs (iter::once)

CAD97:
[Mentioned in the gitter channel](https://gitter.im/pest-parser/pest?at=5be61fe87326df140ee00980), `PrecClimber` hands outs `Pair` (as it should) and `FromPest` (in [pest-parser/ast](https://github.com/pest-parser/ast)'s `from-pest`) takes `&mut Pairs`.

This causes a problem where the current `from-pest` design (taking a mutable iterator cursor) isn't compatible with using a `PrecClimber`.

This patch gives `Pairs` the equivalent of `iter::once`, creating an iterator of a single `Pair` from said `Pair`. This makes it possible to then hand this new iterator to the `FromPest` implementation.

### DIFF
--- a/pest/src/iterators/pair.rs
+++ b/pest/src/iterators/pair.rs
@@ -248,6 +248,14 @@ impl<'i, R: RuleType> Pair<'i, R> {
     }
 }
 
+impl<'i, R: RuleType> Pairs<'i, R> {
+    /// Create a new `Pairs` iterator containing just the single `Pair`.
+    pub fn single(pair: Pair<'i, R>) -> Self {
+        let end = pair.pair();
+        pairs::new(pair.queue, pair.input, pair.start, end)
+    }
+}
+
 impl<'i, R: RuleType> fmt::Debug for Pair<'i, R> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Pair")


pest/src/iterators/pair.rs
```diff
@@ -248,6 +248,14 @@ impl<'i, R: RuleType> Pair<'i, R> {
     }
 }
 
+impl<'i, R: RuleType> Pairs<'i, R> {
```
dragostis: Shouldn't this be part of `iterators/pairs.rs`?

pest/src/iterators/pair.rs
```diff
@@ -248,6 +248,14 @@ impl<'i, R: RuleType> Pair<'i, R> {
...
+    /// Create a new `Pairs` iterator containing just the single `Pair`.
```
dragostis: Should probably contain a doc-test before merging in order to keep at least `pest`'s documentation fully complete.

CAD97:
Oops, misclicked in my IDE and deleted the branch 😅

The magic of GitHub remembering what the head of the branch was since I had the PR open though saves it!

shanemikel:
@CAD97 what's the deal with this PR?

CAD97:
State of the PR is that @dragostis wants a doctest and I've just not had time to think up the contrived example to add to it. It's fairly self-explanatory and we can always add it later so....

bors: r+

shanemikel:
How about this?
```
assert!(pair == Pairs::single(pair).next().unwrap())
```

shanemikel:
Or perhaps this:
```
let pairs = Pairs::single(pair.clone());
assert!(pair == pairs.next().unwrap());
assert!(pairs.next().is_none());
```

PR #341, opened 2018-11-15
Stack fixes

dragostis:
A couple of fixes to the stack infrastructure. These showed up so late partly because of poor testing but also because there was little to no use of the features until now.


dragostis:
@jstnlef, pushing an integer to a `Vec` is very cheap, but, yes, once the rST grammar is done, this should definitely be fine-tuned in order to get great performance.

jstnlef:
I mean it is quite cheap until the `Vec` has to reallocate itself a few times since, if I remember correctly, the way this is currently implemented is that anytime it has to travel down a "branching" path with a stack operation, it will take a snapshot. Restore is only called on errors so the snapshots `Vec` will continue to increase until the program completes.

dragostis:
We could mitigate that by having an `Rc`-like approach, where each `snapshot()` call would return a `Guard` type whose `drop` would automatically pop the snapshot when not used anymore. Restoring would then be possible only with said `Guard`.

jstnlef:
That's fine 👍

PR #344, opened 2018-11-23
Unary prefix/suffix operator support for PrecClimber

segeljakt:
Adds a new `PrattParser`, which works similar to `PrecClimber` but extends it with unary (prefix and suffix) operator handling.

To try it out, checkout this PR and run `cargo run --example=calc` in the terminal.

Given a grammar:

```pest
WHITESPACE   =  _{ " " | "\t" | NEWLINE }

program      =   { SOI ~ expr ~ EOI }
  expr       =   { prefix* ~ primary ~ postfix* ~ (infix ~ prefix* ~ primary ~ postfix* )* }
    infix    =  _{ add | sub | mul | div | pow }
      add    =   { "+" } // Addition
      sub    =   { "-" } // Subtraction
      mul    =   { "*" } // Multiplication
      div    =   { "/" } // Division
      pow    =   { "^" } // Exponentiation
    prefix   =  _{ neg }
      neg    =   { "-" } // Negation
    postfix  =  _{ fac }
      fac    =   { "!" } // Factorial
    primary  =  _{ int | "(" ~ expr ~ ")" }
      int    =  @{ (ASCII_NONZERO_DIGIT ~ ASCII_DIGIT+ | ASCII_DIGIT) }
```

You can now define a `PrattParser`:

```rust
let pratt = PrattParser::new()
      .op(Op::infix(Rule::add, Left) | Op::infix(Rule::sub, Left))
      .op(Op::infix(Rule::mul, Left) | Op::infix(Rule::div, Left))
      .op(Op::infix(Rule::pow, Right))
      .op(Op::postfix(Rule::fac))
      .op(Op::prefix(Rule::neg));
```

And use it like:

```rust
fn parse_to_i32(pairs: Pairs<Rule>, pratt: &PrattParser<Rule>) -> i32 {
  pratt
      .map_primary(|primary| match primary.as_rule() {
          Rule::int  => primary.as_str().parse().unwrap(),
          Rule::expr => parse_to_i32(primary.into_inner(), pratt),
          _          => unreachable!(),
      })
      .map_prefix(|op, rhs| match op.as_rule() {
          Rule::neg  => -rhs,
          _          => unreachable!(),
      })
      .map_postfix(|lhs, op| match op.as_rule() {
          Rule::fac  => (1..lhs+1).product(),
          _          => unreachable!(),
      })
      .map_infix(|lhs, op, rhs| match op.as_rule() {
          Rule::add  => lhs + rhs,
          Rule::sub  => lhs - rhs,
          Rule::mul  => lhs * rhs,
          Rule::div  => lhs / rhs,
          Rule::pow  => (1..rhs+1).map(|_| lhs).product(),
          _          => unreachable!(),
      })
      .parse(pairs)
}
```

Some things to consider:
* Does not support ternary operators.
* Precedence climbing like before uses recursion, though I'm not sure how deep you need to go for this to become a problem.
* The current naming (`prefix`/`suffix`/`infix`) could be misinterpreted as binary operators with `prefix`/`suffix`/`infix` notation. Does it work, or should I name things more explicitly, e.g., `unary_prefix`/`unary_suffix`/`binary_infix`?

### DIFF
--- a/pest/tests/calculator.rs
+++ b/pest/tests/calculator.rs
@@ -13,7 +13,7 @@ extern crate pest;
 use pest::{state, ParseResult, Parser, ParserState};
 use pest::error::Error;
 use pest::iterators::{Pair, Pairs};
-use pest::prec_climber::{Assoc, Operator, PrecClimber};
+use pest::pratt_parser::{Assoc, Op, PrattParser};
 
 #[allow(dead_code, non_camel_case_types)]
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
@@ -109,8 +109,8 @@ impl Parser<Rule> for CalculatorParser {
     }
 }
 
-fn consume<'i>(pair: Pair<'i, Rule>, climber: &PrecClimber<Rule>) -> i32 {
-    let primary = |pair| consume(pair, climber);
+fn consume<'i>(pair: Pair<'i, Rule>, pratt: &PrattParser<Rule>) -> i32 {
+    let primary = |pair| consume(pair, pratt);
     let infix = |lhs: i32, op: Pair<Rule>, rhs: i32| match op.as_rule() {
         Rule::plus => lhs + rhs,
         Rule::minus => lhs - rhs,
@@ -122,7 +122,10 @@ fn consume<'i>(pair: Pair<'i, Rule>, climber: &PrecClimber<Rule>) -> i32 {
     };
 
     match pair.as_rule() {
-        Rule::expression => climber.climb(pair.into_inner(), primary, infix),
+        Rule::expression => pratt
+          .map_primary(primary)
+          .map_infix(infix)
+          .parse(pair.into_inner()),
         Rule::number => pair.as_str().parse().unwrap(),
         _ => unreachable!()
     }
@@ -187,14 +190,15 @@ fn expression() {
 }
 
 #[test]
-fn prec_climb() {
-    let climber = PrecClimber::new(vec![
-        Operator::new(Rule::plus, Assoc::Left) | Operator::new(Rule::minus, Assoc::Left),
-        Operator::new(Rule::times, Assoc::Left) | Operator::new(Rule::divide, Assoc::Left)
-            | Operator::new(Rule::modulus, Assoc::Left),
-        Operator::new(Rule::power, Assoc::Right),
-    ]);
+fn pratt_parse() {
+    let pratt = PrattParser::new()
+        .op(Op::infix(Rule::plus,    Assoc::Left)
+          | Op::infix(Rule::minus,   Assoc::Left))
+        .op(Op::infix(Rule::times,   Assoc::Left)
+          | Op::infix(Rule::divide,  Assoc::Left)
+          | Op::infix(Rule::modulus, Assoc::Left))
+        .op(Op::infix(Rule::power,   Assoc::Right));
 
     let pairs = CalculatorParser::parse(Rule::expression, "-12+3*(4-9)^3^2/9%7381");
-    assert_eq!(-1_525, consume(pairs.unwrap().next().unwrap(), &climber));
+    assert_eq!(-1_525, consume(pairs.unwrap().next().unwrap(), &pratt));
 }
--- a/meta/src/parser.rs
+++ b/meta/src/parser.rs
@@ -11,7 +11,7 @@ use std::char;
 use std::iter::Peekable;
 
 use pest::iterators::{Pair, Pairs};
-use pest::prec_climber::{Assoc, Operator, PrecClimber};
+use pest::pratt_parser::{Assoc, Op, PrattParser};
 use pest::{Span, Parser};
 use pest::error::{Error, ErrorVariant};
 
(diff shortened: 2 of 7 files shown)


pest/src/prec_climber.rs
```diff
@@ -79,137 +91,322 @@ impl<R: RuleType> BitOr for Operator<R> {
...
+        check(has_infix,  self.infix.is_some(),  "infix");
+    }
+
+    /// Start climbing `pairs`.
+    pub fn climb<P: Iterator<Item = Pair<'pest, R>>>(
+        &self,
+        pairs: P
+    ) -> T {
+
+        #[cfg(debug_assertions)]
```
dragostis: Maybe it would be worthwhile to introduce a third type that this one would get compiled to. This would both offer a nice API, i.e. return a `None` if missing `map_*`s, and relive the user from possible headaches in release builds.

pest/src/prec_climber.rs
```diff
@@ -79,137 +91,322 @@ impl<R: RuleType> BitOr for Operator<R> {
...
+        let mut iter = Some(op);
+        while let Some( Op { rule, kind, next } ) = iter.take() {
+            self.ops.insert(rule, (kind, self.prec));
+            iter = next.map(|op| *op);
+        }
+        self
+    }
+
+    /// Maps primary expressions with a closure `primary`.
+    pub fn map_primary<'climber, 'pest, X, T>(&'climber self, primary: X)
```
dragostis: `'i` in pest stands for `input`. `'pest` lifetime is too ambiguous.

```suggestion
    pub fn map_primary<'climber, 'i, X, T>(&'climber self, primary: X)
```

pest/src/prec_climber.rs
```diff
@@ -7,66 +7,78 @@
...
+    /// Defines `rule` as a prefix unary operator.
+    pub fn prefix(rule: R) -> Self {
+        Self {
+            rule,
+            kind: OpKind::Prefix,
+            next: None
+        }
+    }
+
+    /// Defines `rule` as a suffix unary operator.
```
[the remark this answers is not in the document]
segeljakt: done

pest/src/prec_climber.rs
```diff
@@ -79,137 +91,322 @@ impl<R: RuleType> BitOr for Operator<R> {
...
+///     suffix   =  _{ fac }
+///       fac    =   { "!" } # Factorial
+///     primary  =  _{ int | "(" ~ expr ~ ")" }
+///       int    =  @{ ASCII_NONZERO_DIGIT ~ ASCII_DIGIT* }
+/// ```
+///
+/// Below is a [`PrecClimber`] that is able to climb an `expr` in the above grammar. The order
+/// of precedence corresponds to the order in which [`op`] is called. Thus, `mul` will
+/// have higher precedence than `add`. Operators can also be chained with `|` to give them equal
+/// precedence. In the current version, prefix operators precede suffix, and suffix
```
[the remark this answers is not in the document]
segeljakt: The algorithm should now be able to deal with prefix/postfix operator precedence

pest/src/prec_climber.rs
```diff
@@ -79,137 +91,322 @@ impl<R: RuleType> BitOr for Operator<R> {
...
+///             Rule::int  => primary.as_str().parse().unwrap(),
+///             Rule::expr => parse_expr(primary.into_inner(), climber), // from "(" ~ expr ~ ")"
+///             _          => unreachable!(),
+///         })
+///         .map_prefix(|op, rhs| match op.as_rule() {
+///             Rule::neg  => -rhs,
+///             _          => unreachable!(),
+///         })
+///         .map_suffix(|lhs, op| match op.as_rule() {
+///             Rule::fac  => (1..=lhs).product(),
```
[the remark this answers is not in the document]
segeljakt: I changed it to `(1..lhs+1)`

pest/src/prec_climber.rs
```diff
@@ -79,137 +91,322 @@ impl<R: RuleType> BitOr for Operator<R> {
...
+type SuffixFn<'pest, R, T> = Box<Fn(T, Pair<'pest, R>) -> T + 'pest>;
+type InfixFn<'pest, R, T> = Box<Fn(T, Pair<'pest, R>, T) -> T + 'pest>;
+
+/// Product of calling [`map_primary`] on [`PrecClimber`], defines how expressions should
+/// be mapped.
+///
+/// [`map_primary`]: struct.PrecClimber.html#method.map_primary
+/// [`PrecClimber`]: struct.PrecClimber.html 
+pub struct PrecClimberMap<'climber, 'pest, R, F, T>
     where
```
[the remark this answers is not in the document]
segeljakt: it should now be properly formatted

pest/src/prec_climber.rs
segeljakt: I changed so that the code panics either when:
* The stream of tokens does not match the expected pattern.
* No map function is specified for a certain kind of operator.

pest/src/prec_climber.rs
segeljakt: done

pest/tests/calculator.rs
```diff
@@ -187,14 +190,15 @@ fn expression() {
 }
 
 #[test]
-fn prec_climb() {
```
AGausmann: I see tests for `PrecClimber` removed and replaced, but `PrecClimber` is still in the codebase. I don't think these should be removed until `PrecClimber` itself is.

AGausmann:
This would be a great feature to have! Personally I think the right way to incorporate this is to keep the public APIs involving `PrecClimber` at least until next major release, but supplement them with newer APIs that use `PrattParser` instead. There's nothing wrong with keeping two different algorithms around for a little while.

segeljakt:
I recently added this to the [pest-ast](https://github.com/segeljakt/ast) crate, you can use it like [this](https://github.com/segeljakt/ast/blob/derive-prec-climber/derive/examples/arith.rs) with a grammar like [this](https://github.com/segeljakt/ast/blob/derive-prec-climber/derive/examples/arith.pest).

lberezy:
I understand that this is a little bit old now and would probably requite a bit of fixing up, but is there any movement on this?

I think it would be quite handy to enable prefix/suffix operators like this.

dragostis:
@lberezy, there hasn't been that much work here ever since, but I'm happy to merge this if it gets push forward.

segeljakt:
Hi again, I refactored this into its own library https://crates.io/crates/pratt. Feel free to copy it, add it as a dependency, or just point to it from to pest. You may close this PR if you want.

tomtau:
@segeljakt @lberezy @dragostis I opened a new PR based on this one: https://github.com/pest-parser/pest/pull/710 -- I had to make a few small changes in order to support `no_std`. I kept the original PrecClimber in order not to break the existing public API, but marked it with the `#[deprecated]` annotation.

PR #354, opened 2018-12-09
Enable gating on rustfmt/clippy

CAD97:
r? @dragostis 

I'm planning re-adding the `box_tree` macro for tests in another PR on top of this ones just to keep things relatively clean. That will be a fun cherry-pick. (This was PR done mostly automatically by the power of tooling!)

PR #355, opened 2018-12-09
Test cleanup with box_tree! macro

CAD97:
Includes #354 (the two changesets would be very conflicting otherwise). Contains the last of the changes removed from #315 for the sake of getting it merged easily (and a cleaner git history). r? @dragostis


CAD97:
Whoops, #354 apparently missed some things; fixing that then will rebase this.

PR #356, opened 2018-12-09
Fix coverage job

CAD97:
It wasn't set up completely correctly to begin with:

- `cargo install` failed if it had been cashed
- `tarpaulin`'s clean step kills the bootstrap executable if we let it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,9 @@ matrix:
     - name: coverage
       rust: nightly-2018-12-08 # Update this when updating tarpaulin
       env: RUSTFLAGS="--cfg procmacro2_semver_exempt"
-      install: cargo install cargo-tarpaulin --version 0.6.10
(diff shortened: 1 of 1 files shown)


CAD97:
Update: the problem seems to be that [tarpaulin's build is trying and failing to run the bootstrap executable](https://travis-ci.org/pest-parser/pest/jobs/465705660#L601). I'm now trying to figure out why it's not working.

CAD97:
I think I've figured it out finally: `tarpaulin` by default does a `cargo clean` then rebuilds everything; we need to prevent it from doing so or else the bootstrap build script complains (though in actuality, it _would_ work since we've done the bootstrap previously).

CAD97:
I'd not be surprised if coverage has been broken for a while because of this. Let's get it fixed!

bors: r+

dragostis:
Thanks for figuring this out! It wasn't very obvious what the issue was.

PR #360, opened 2018-12-20
add support for inline grammars

cavedweller:
I need inline definitions of grammar files. If there is a better way of doing this, please let me know!

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -52,10 +52,8 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
                 Err(error) => panic!("error opening {:?}: {}", file_name, error),
             };
             (data, Some(path.clone()))
-        },
-        GrammarSource::Inline(content) => {
-            (content, None)
-        },
+        }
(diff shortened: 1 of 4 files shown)


generator/src/lib.rs
```diff
@@ -36,25 +36,30 @@ use pest_meta::{optimizer, unwrap_or_report, validator};
...
-    };
-
-    let data = match read_file(&path) {
-        Ok(data) => data,
-        Err(error) => panic!("error opening {:?}: {}", file_name, error),
+    let (name, generics, content) = parse_derive(ast);
+
+    let (data, path) = match content {
+        GrammarSource::File(ref path) => {
+            let root = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
```
dragostis: I think this should work fine.

```suggestion
            let root = env::var("CARGO_MANIFEST_DIR").unwrap_or(".".into());
```

lwandrebeck:
@dragostis I bet that `cargo clippy` suggested the use of `or_else(|_|` instead of `or(`

PR #368, opened 2019-01-29
add elastic-rs which use pest as parser

cch123:
elastic-rs is a library which converts user input to elasticsearch DSL.
pest is used to parse the boolean expression

Thanks a lot for this elegant parser!


dragostis:
Thanks for the PR! Could you please put the project in the alphabetical order?

cch123:
@dragostis done

PR #375, opened 2019-02-28
Fix #374

### LINKED ISSUE #374: Is the `WHITESPACE` rule name supposed to be case-sensitive ?
From the docs, it seems that `WHITESPACE` is a special rule, but `whitespace` is a normal one. However, when I define my own `whitespace` rule, it triggers a special validation routine: https://github.com/pest-parser/pest/blob/28616d951fcb937021c701696c5afaa6727b829c/meta/src/validator.rs#L407-L442 .
Looking around the code, it looks like this is the only place where the lowercase `whitespace` name is used, as opposed to `WHITESPACE`, so I think it's a bug. The same thing happens with `COMMENT`/`comment`.

Nadrieril:
As explained in #374, this looks like a bug. It seems this was missed by https://github.com/pest-parser/pest/pull/235/commits/2e331804d94deb0b78b6fce46330bbb4ab7409ed

### DIFF
--- a/meta/src/validator.rs
+++ b/meta/src/validator.rs
@@ -410,7 +410,7 @@ fn validate_whitespace_comment<'a, 'i: 'a>(rules: &'a [ParserRule<'i>]) -> Vec<E
     rules
         .iter()
         .filter_map(|rule| {
-            if rule.name == "whitespace" || rule.name == "comment" {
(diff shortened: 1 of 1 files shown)


CAD97:
bors: r+

Yep, this was missed.

PR #381, opened 2019-03-05
Generate lexical information of `Pair` and `Pairs` as JSON

### LINKED ISSUE #377: Feature request: export lexical information as json/xml/whatever
I hope there can be a feature that allows us to pretty-print a json/xml/csv/whatever which includes the lexicial information (like token type, pos/line/column), like the AST printer in https://pest.rs but I want an easy way to invoke that locally and get a more easy-to-parse format (I hope it's csv (you can represent nested tree-structure via adding a `depth` attribute in each column), so I don't even need a json parser).

After this, we can make a lot of goodies like some other tools that renders a webpage of code that is syntax-highlighted according to the lexical information.

Goncalerta:
Closes #377

Implements the `to_json` function for both `Pair` and `Pairs` which generates a pretty-printed JSON that contains their information.

I followed the sketch given in #377, but I wonder if instead of `"pos": [start, end]` it would be better to have `"start"` and `"end"` as separate fields, as in pretty-printed format I personally find it easier to read with less nesting.

This is the generated string right now:
```
{
  "pos": [
    0,
    5
  ],
  "pairs": [
    {
      "pos": [
        0,
        3
      ],
      "rule": "a",
      "inner": {
        "pos": [
          1,
          2
        ],
        "pairs": [
          {
            "pos": [
              1,
              2
            ],
            "rule": "b",
            "inner": "b"
          }
        ]
      }
    },
    {
      "pos": [
        4,
        5
      ],
      "rule": "c",
      "inner": "e"
    }
  ]
}
```

This would be with `"start"` and `"end"`:
```
{
  "start": 0,
  "end": 5,
  "pairs": [
    {
      "start": 0,
      "end": 3,
      "rule": "a",
      "inner": {
        "start": 1,
        "end": 2,
        "pairs": [
          {
            "start": 1,
            "end": 2,
            "rule": "b",
            "inner": "b"
          }
        ]
      }
    },
    {
      "start": 4,
      "end": 5,
      "rule": "c",
      "inner": "e"
    }
  ]
}
```

### DIFF
--- a/pest/Cargo.toml
+++ b/pest/Cargo.toml
@@ -11,8 +11,14 @@ categories = ["parsing"]
 license = "MIT/Apache-2.0"
 readme = "_README.md"
 
+[features]
+# Enables the `to_json` function for `Pair` and `Pairs`
+pretty-print = ["serde", "serde_json"]
+
 [dependencies]
 ucd-trie = "0.1.1"
+serde = { version = "1.0.89", optional = true }
+serde_json = { version = "1.0.39", optional = true}
 
 [badges]
 codecov = { repository = "pest-parser/pest" }
--- a/pest/src/iterators/pair.rs
+++ b/pest/src/iterators/pair.rs
@@ -13,6 +13,9 @@ use std::ptr;
 use std::rc::Rc;
 use std::str;
 
+#[cfg(feature = "pretty-print")]
+use serde::ser::SerializeStruct;
+
 use super::pairs::{self, Pairs};
 use super::queueable_token::QueueableToken;
 use super::tokens::{self, Tokens};
@@ -234,6 +237,13 @@ impl<'i, R: RuleType> Pair<'i, R> {
         tokens::new(self.queue, self.input, self.start, end + 1)
     }
 
+    /// Generates a string that stores the lexical information of `self` in
+    /// a pretty-printed JSON format.
+    #[cfg(feature = "pretty-print")]
+    pub fn to_json(&self) -> String {
+        ::serde_json::to_string_pretty(self).expect("Failed to pretty-print Pair to json.")
+    }
+
     fn pair(&self) -> usize {
         match self.queue[self.start] {
             QueueableToken::Start {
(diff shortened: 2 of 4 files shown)


CAD97:
(sigh, why must unit tests still not count for coverage...)

A few drive-by notes:

- My original sketch had `pos: (start, end)` because I was thinking serde data model which does have tuples whereas JSON doesn't. If it were on the same line (full control over pretty printing) then the array/tuple solution might've been nice, but inline works nicely as well.
- Personally I expected a `R: Serialize` bound rather than using enum debug formatting. (This is equivalent in "human"/text formats but different in the serde model and in compact formats.) Though I suppose that'd require telling `serde_derive` to emit a `#[derive(Serialize)]` for the token enum as well.
- I'm not certain exactly how `serde_derive` handles "untagged enum" representations, but serializing `inner` as one of two types rather than as a serde enum scares me for some reason. I know we don't need to deserialize, but it still squicks me out a bit, even if we're only targeting self-describing, human-oriented formats.

Goncalerta:
> * My original sketch had `pos: (start, end)` because I was thinking serde data model which does have tuples whereas JSON doesn't. If it were on the same line (full control over pretty printing) then the array/tuple solution might've been nice, but inline works nicely as well.

Oh, I see. Well, in order to have full control over pretty printing, there is also the possibility of using `format!`. That would even lift the dependency on `serde` and it shouldn't be hard to implement, as the template is pretty simple. However, this would make it harder to serialize to other data formats if that is ever desired.

> * Personally I expected a `R: Serialize` bound rather than using enum debug formatting. (This is equivalent in "human"/text formats but different in the serde model and in compact formats.) Though I suppose that'd require telling `serde_derive` to emit a `#[derive(Serialize)]` for the token enum as well.

Well, I personally think that would add too much complexity, because in order to be able to call `#[derive(Serialize)]` on the macro, we would need to expose `serde` on `pest`'s API. 

> * I'm not certain exactly how `serde_derive` handles "untagged enum" representations, but serializing `inner` as one of two types rather than as a serde enum scares me for some reason. I know we don't need to deserialize, but it still squicks me out a bit, even if we're only targeting self-describing, human-oriented formats.

Well, if you prefer I could make a local enum for it, I just thought it wasn't worth the effort in this case.

CAD97:
> > My original sketch had pos: (start, end) because I was thinking serde data model which does have tuples whereas JSON doesn't. If it were on the same line (full control over pretty printing) then the array/tuple solution might've been nice, but inline works nicely as well.

> Oh, I see. Well, in order to have full control over pretty printing, there is also the possibility of using `format!`. That would even lift the dependency on serde and it shouldn't be hard to implement, as the template is pretty simple. However, this would make it harder to serialize to other data formats if that is ever desired.

Yeah; I'd potentially go so far as to _just_ do the `impl Serialize` and allow the user to specify a data format themselves. (Personally, I like RON or YAML for "read-only" data formats over JSON.)

> Well, I personally think that would add too much complexity, because in order to be able to call `#[derive(Serialize)]` on the macro, we would need to expose serde on pest's API.

Yep, the solution here is great for this.

> Well, if you prefer I could make a local enum for it, I just thought it wasn't worth the effort in this case.

I'd feel better with a `#[derive(Serialize)] #[serde(untagged)]` enum here, as there's _some_ guarantee that the serialization API isn't being misused if we go through `serde_derive` here. But I guess this is more up to @dragostis here.

ice1000:
What's the current status of this pr?

ice1000:
Can there be a `bin` that takes a text file and the grammar file and outputs the json (like to stdout or to a file)?

ice1000:
One usage can be in my IDE plugin: on-the-fly syntax highlighting your code.

PR #387, opened 2019-04-14
Update cargo-tarpaulin

CAD97:
Note: this only effects an allowed-failure CI job. Make sure it passes manually before merging!


CAD97:
Coverage job manually confirmed to be working.

bors: r+

PR #389, opened 2019-04-14
Enforce -D warnings on tests

CAD97:
We had a lot of `Pair::into_span` hanging around, including some `.clone().into_span()`.

Additionally, the code compiled for `cargo bootstrap` isn't clippy-checked, so that will need addressing at a later point.

r? @dragostis

### DIFF
--- a/pest/src/iterators/pair.rs
+++ b/pest/src/iterators/pair.rs
@@ -135,7 +135,7 @@ impl<'i, R: RuleType> Pair<'i, R> {
     /// #     state.rule(Rule::ab, |s| s.match_string("ab"))
     /// }).unwrap().next().unwrap();
     ///
-    /// assert_eq!(pair.into_span().as_str(), "ab");
+    /// assert_eq!(pair.as_span().as_str(), "ab");
     /// ```
     #[inline]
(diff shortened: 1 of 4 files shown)


pest/src/iterators/pair.rs
```diff
@@ -135,7 +135,7 @@ impl<'i, R: RuleType> Pair<'i, R> {
     /// #     state.rule(Rule::ab, |s| s.match_string("ab"))
     /// }).unwrap().next().unwrap();
     ///
-    /// assert_eq!(pair.into_span().as_str(), "ab");
+    /// assert_eq!(pair.as_span().as_str(), "ab");
```
CAD97: This got caught up in my refactoring (the doc-test for `Pair::into_span`). I'm unsure which we should use here.

dragostis: I'd leave the old one here.

dragostis:
@CAD97, would you still be interested to fix the remaining issues with the PR?

MarinPostma:
closing since this is inactive and should be fixed with #507

CAD97:
Oh wow this was still hanging around 😅

Issue #394, opened 2019-06-08
Strange results with stack and non-matching alternative


golddranks:
I'm getting strange results with the following grammar:
```
raw_string_inner = { (!PEEK_ALL ~ ANY)* }
raw_string = ${ PUSH("'")+ ~ raw_string_inner ~ POP_ALL }

named_exp = { raw_string ~ "->" ~ ASCII_ALPHA_LOWER }

exp = { named_exp | raw_string }
```

With test code `'' it's a test ''` and matching against `exp`, I'm expecting it to find a match, but it doesn't. Instead, with `'' it's a test '''` (unbalanced single quotes), it does find a match.

When I test directly against `raw_string`, it finds a match. It seems, from the behaviour, that `exp`, failing to match the first alternative, messes up the state of the stack (one-off error?) when it starts to match the next one.

It might, of course, also be that I'm misunderstanding something about how the stack works, but my general understanding is that once one alternative fails to match, it shouldn't have an effect to matching the next one...?

I haven't looked in the generated code or runtime yet; just gonna start investigating now, but figured that I'd make an issue first in case there's a quick gotcha I haven't taken into account.

golddranks:
Btw. I got it working with a better grammar that uses the stack in a saner way:

```
raw_string_inner = { (!PEEK ~ ANY)* }
raw_string = ${ PUSH("'"+) ~ raw_string_inner ~ POP }

named_exp = { raw_string ~ "->" ~ ASCII_ALPHA_LOWER }

exp = { named_exp | raw_string }
```

However, it still bothers me whether it's a bug; seems like the state of the stack is not the same when trying the next alternative after a failed one.

golddranks:
Okay, tried to instrument the generated parsing code a bit. (Btw. if there is a better way to debug than doing `cargo expand` and instrumenting by hand, it would be very helpful to mention it in the documentation.) Here is the `exp` rule parsing function:
```
#[inline]
#[allow(non_snake_case, unused_variables)]
pub fn exp(
    state: Box<::pest::ParserState<Rule>>,
) -> ::pest::ParseResult<Box<::pest::ParserState<Rule>>> {
    dbg!(&state);
	dbg!(
    state.rule(Rule::exp, |state| {
        dbg!(&state);
        dbg!(state
            .restore_on_err(|state| {
                dbg!(&state);
                dbg!(self::named_exp(state))
            })
            .or_else(|state| {
                dbg!(&state);
                dbg!(state.restore_on_err(|state|
[...]

PR #395, opened 2019-06-09
Clearing checkpoints in error handler on successful parse

golddranks:
* When `restore_on_err` is called, a checkpoint is added.
* When parsing fails inside the call, the checkpoint is resumed. Specifically, stack is resumed to the state it was before entering `restore_on_err`.
* However, when parsing inside `restore_on_err` succeeds, at the moment, the checkpoint is not cleared on the exit.
* This leads to bugs where after returning from `restore_on_err`, another error is encountered, but the checkpoint that was set in `restore_on_err` is incorrectly resumed; this causes the stack to become in inconsistent state.
* Fixed the bug by adding a function for clearing checkpoints and calling it on the successful path of `restore_on_err`.
* Resolves https://github.com/pest-parser/pest/issues/394
* Added test for this case. However, the testing infrastructure isn't quite clear for me, so it might be that the test would be better expressed somewhere else, in some other way. Please advise on this.


pest/src/parser_state.rs
```diff
@@ -1149,4 +1157,81 @@ mod test {
         assert_eq!(normalize_index(-5, 5), Some(0));
         assert_eq!(normalize_index(-6, 3), None);
     }
+
+    #[test]
+    fn test_restoring_on_err() {
+        // This corresponds roughly to Pest syntax:
+        //
+        // stack_use = @{ PUSH("A")+ ~ "B" ~ POP_ALL }
```
dragostis: I would make this part of https://github.com/pest-parser/pest/blob/master/derive/tests/grammar.pest#L38 in order not to have to define this manually.

golddranks:
All right, cleaned the test up. The original problem didn't manifest itself with the rule `repeat_mutate_stack = { (PUSH('a'..'c') ~ ",")* ~ POP ~ POP ~ POP }` so I could'n reuse it in the test.

Issue #396, opened 2019-06-15
`one or more` rules consume trailing whitespace when exactly matched once


ejoebstl:
Given the following grammar: 

```
a = { "a" ~ b+ | a }
b = { "b" }
root = { SOI ~ a* ~ EOI }
```

Then, parsing the following text: 
```
a b

```
Will associate the trailing newline after `b` with the rule for `a`, instead of the the rule for `root`.

That seems to work properly with `zero-or-more` expressions, where the whitespace will be associated with the `root` rule. 

Also, it works properly for more than one `b`. The following example:
```
a b b

```
Puts the whitespace correctly inside root.

ejoebstl:
The issue might be related to the following line, which inserts a concatenation between the first and the other repetitions. 

https://github.com/pest-parser/pest/blob/a0264c0633226708a086d298b50e59f9cbf8efef/meta/src/optimizer/unroller.rs#L18

ejoebstl:
After looking through the code, I'd say this is because `Expr::RepOnce` is translated to a sequence. 

E.g. for pest, `b+` is equal to `b ~ b*`.

ejoebstl:
Proposed fix in #397.

dragostis:
Ah, finally. I knew this day was going to come. Unfortunately, this implementation bug is probably something that people make use of in their logic. Fixing this might actually break them, but I'm happy to give it a try regardless maybe there's no actual dependee that uses it.

I have a fix for 3.0, but I haven't had that much time to work on it.

keks:
This just bit us. Are there still plans to resolve this? Is there evidence that people rely on this bug?

And lastly, is there a good workaround? The obvious idea would be to strip the white space off of everything pest returns, but I suspect there is a more elegant way, since otherwise there probably would be more people commenting here.

In our case we are having trouble to parse data in the format "x <- foo" (see [this page]).

Also, otherwise we are having a great time using pest, thank you!

cc @siccegge @chrisbrzuska

[this page]: https://codimd.faui2k.cloud/s/kjeothgO_#

PR #397, opened 2019-06-15
Implementation of `one-or-more` rule

ejoebstl:
Properly implements a rule for RepOnce, instead of translating it to `x ~ x*`. This solves the whitespace issue described in #396.

### DIFF
--- a/meta/src/optimizer/mod.rs
+++ b/meta/src/optimizer/mod.rs
@@ -63,7 +63,7 @@ fn rule_to_optimized_rule(rule: Rule) -> OptimizedRule {
             Expr::Rep(expr) => OptimizedExpr::Rep(Box::new(to_optimized(*expr))),
             Expr::Skip(strings) => OptimizedExpr::Skip(strings),
             Expr::Push(expr) => OptimizedExpr::Push(Box::new(to_optimized(*expr))),
-            Expr::RepOnce(_)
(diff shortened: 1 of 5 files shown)


dragostis:
Thanks for putting this together! As I said, I'm afraid that this might break people, but I'm willing to try to see if dependent crates pass their tests or not with this fix.

NoahTheDuke:
Any chance this could get updated?

meta/src/optimizer/mod.rs
```diff
@@ -105,6 +104,7 @@ pub enum OptimizedExpr {
     Choice(Box<OptimizedExpr>, Box<OptimizedExpr>),
     Opt(Box<OptimizedExpr>),
     Rep(Box<OptimizedExpr>),
+    RepOnce(Box<OptimizedExpr>),
```
tomtau, line 107: `enum OptimizedExpr` wasn't annotated with `#[non_exhaustive]`, so adding a new variant to it is unfortunately a semver breaking change

tomtau (commented):
the fix seems all right, but it's a breaking change: both logic-wise (if any external crates depended on the old behaviour) and API-wise due to a new variant on a public enum, so potentially something for 3.x

tomtau:
revived this PR in https://github.com/pest-parser/pest/pull/878

PR #410, opened 2019-08-20
Don't enable std feature on sha-1

roblabla:
I'm currently working on a no_std project that uses Pest in a proc macro, and also uses sha1 somewhere else in the crate graph. It uses sha1 in nostd mode. But because cargo ~~sucks~~ has [some really broken behavior](https://github.com/rust-lang/cargo/issues/4866) around build and dev dependencies, pest ends up activating the std feature for my build.

Pest doesn't really need the std feature of SHA-1 anyways. So this PR disables it.


CAD97:
bors: r+

@dragostis, can we get a patch version published with this?

PR #422, opened 2019-10-23
Add AshPaper to the list of projects using pest

shnewto:
👋 I recently wrote an interpreter for an esolang called AshPaper that interprets formally constrained poetry as compilable code. I used pest to parse and categorize the input (poetry) into corresponding instructions and it was _such_ a pleasure to work with. 

I'd be honored to make it onto the list of projects that use pest. Thanks for all you do, I love this project!


dragostis:
bors retry

PR #442, opened 2020-02-22
Remove HashMap from derive/generate

nagisa:
There is a lot of value in outputs produced by compilers being
reproducible. Rust makes it somewhat hard to get it right, but most of
the time rustc’s outputs are also reproducible. This allows us to enjoy
all the benefits as other more primitive toolchains. Most of the time.

Pest was, sadly, one of those examples where reproducibility was not a
concern and HashMap was used to implement some of the  code generation
portions. Replacing this `HashMap` with a `Vec` and a `O(nm)` loop is
the easy solution. With the values that `n` and `m` get substituted with
in practice, however, this loop should still be approximately as good as
the previous `HashMap` implementation.

### DIFF
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -7,7 +7,6 @@
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
 
-use std::collections::HashMap;
 use std::path::PathBuf;
 
 use proc_macro2::{Span, TokenStream};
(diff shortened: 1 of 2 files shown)


CAD97:
Another option would be [indexmap](https://docs.rs/indexmap/1.3.2/indexmap/), which both uses hashes and guarantees a stable (insertion) ordering. It's basically this approach plus a hash-based lookup cache to make lookups amortized constant time. But that said, this is a fine fix and a fix you have is always better than a theoretical one, which can be applied later.

bors: r+

cc @dragostis a release for this fix ASAP would be nice.

CAD97:
Gah, looks like #443 is going to block the build. I need @dragostis to fix this one 😓

dragostis:
The `Vec` should be a fine solution here.

nagisa:
Please release a version of `pest_generator` with this fix sooner rather than later.

PR #445, opened 2020-02-27
initial static precedence climber

ametisf:
Quick modification of PrecClimber to make it possible to define as a static plus a convenience macro.

I'll clean it up and/or modify if you think it's a good idea to have something like this.


CAD97:
This is cool!

Is it possible to unify the static and runtime prec climbers by e.g. using `Cow` or something (in an API compatible way)?

ametisf:
It's almost possible, except `std::borrow::Cow` requires `std::borrow::ToOwned` which requires the generic parameter `R` to be `Clone`, but for now trait bounds other than lifetimes and `Sized` are not stable for `const fn`.

It could be made a nightly-only feature gated API, so I did that.

The only change to public API that could be considered breaking is the `'static` lifetime bound required on the generic parameter `R` to `PrecClimber`, however I think it can be ignored since it's very unlikely someone is using anything else than a plain empty enum there. Loosening the other bounds isn't a breaking change.

pest/src/prec_climber.rs
```diff
@@ -9,13 +9,96 @@
...
+/// static CLIMBER: PrecClimber<Rule> = prec_climber![
+///     L   plus | minus,
+///     L   times | divide,
+///     R   power,
+/// ];
+/// ```
+#[cfg(const_prec_climber)]
+#[macro_export]
+macro_rules! prec_climber {
+    ( __assoc L ) => { $crate::prec_climber::Assoc::Left };
```
dragostis: I've been usually using `@foo` for these, but I'm not completely sure there's a precedence in pest for this yet.

ametisf: I've changed it to `@` too, it also allowed me to move the user callable pattern to the top which makes the documentation for the macro a bit more readable.

PR #451, opened 2020-03-13
Makes include_str! use a relative path

benbrittain:
In situations where `CARGO_MANIFEST_DIR` is set atypically, pest can break it's grammar file finding


benbrittain:
cc/ @dragostis

CAD97:
Test failure is just the test that shows the output of expansion. Update that and it _should_ be fine.

CAD97:
bors: r=dragostis

benbrittain:
Thanks for the speedy reviews!

Is there any change we could do a patch bump for the version? The environment I'm working in at the moment doesn't allow relying on the git URI.

Issue #455, opened 2020-04-25
Expose more ways for users to manipulate Span


tangmi:
I often find myself reaching for `pest::span::Span` to keep a mapping back to the parsed source file. I sometimes run into issues because `Span::input` is not `pub`.

I'd like to propose adding two methods to `Span` to give it more flexibility for users:
- `pub fn sub_span(&self, range: impl std::ops::RangeBounds<usize>) -> Option<Span>`
  - Allows one to sub-slice a `Span`, returning a smaller span (or "substring") which preserves the original `Span`'s input.
- `pub fn lines_span(&self) -> LinesSpan`
  - where `LinesSpan` is an iterator, much like `pest::span::Lines`, but returns `Span` instead of `&str`.
  - `Lines` can be implemented as `span.lines_span().map(|span| span.start_pos().line_of())`

I've created a draft PR with a more concrete example of this proposal: #456.

Thanks for your time!

PR #456, opened 2020-04-25
Implement `Span::sub_span` and `Span::lines_span`.

tangmi:
Prototype implementation for #455

### DIFF
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -62,6 +62,25 @@ impl<'i> Span<'i> {
         }
     }
 
+    /// Attempts to create a new span based on a sub-range.
+    ///
+    /// TODO better docs
+    pub fn sub_span(&self, range: impl std::ops::RangeBounds<usize>) -> Option<Span> {
+        Span::new(
+            self.input,
+            match range.start_bound() {
+                std::ops::Bound::Included(offset) => self.start + offset,
+                std::ops::Bound::Excluded(offset) => self.start + offset + 1,
+                std::ops::Bound::Unbounded => self.start,
+            },
+            match range.end_bound() {
+                std::ops::Bound::Included(offset) => self.start + offset + 1,
+                std::ops::Bound::Excluded(offset) => self.start + offset,
+                std::ops::Bound::Unbounded => self.end,
+            },
+        )
+    }
+
     /// Returns the `Span`'s start byte position as a `usize`.
     ///
     /// # Examples
@@ -201,6 +220,14 @@ impl<'i> Span<'i> {
     #[inline]
     pub fn lines(&self) -> Lines {
         Lines {
+            inner: self.lines_span(),
+        }
+    }
+
+    /// TODO better docs
+    #[inline]
+    pub fn lines_span(&self) -> LinesSpan {
+        LinesSpan {
             span: self,
             pos: self.start,
         }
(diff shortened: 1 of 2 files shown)


CAD97:
This looks reasonable to me at a glance. @dragostis ?

tangmi:
Some concerns I have (that may not actually be real issues):
- Is `Span` intended to be consumable by users in this way? My assumption is that pest expects/allows users to create `Span`, given that `Span::new` is exposed (although I'm proposing this PR because `Span::new` is difficult to use with a parsed document since `Span::input` is not exposed).
- I didn't put much thought into the naming of new members
  - `Span::sub_span` would be called `Span::get` to parallel `str::get` (there is no "substring" method in `std`, just range indexing)
  - `LinesSpan` was just named as such to avoid changing the name of `Lines`.

Alextopher:
I am also interested in `sub_span`. I am also happy to provide some kind of "help wanted" to see it through

Alextopher:
> Is Span intended to be consumable by users in this way?
There are similar requests for additional flexibility ( such as #469 ). There are other types that provide flexibility such as [ErrorVariant::CustomError](https://docs.rs/pest/2.2.1/pest/error/enum.ErrorVariant.html#variant.CustomError). Subspans could be similarly useful.

> `Span::sub_span` would be called `Span::get` to parallel `str::get`
I prefer `Span::get` to match other subslicing behaviors

`LinesSpan` seems fine to me, it should only ever be used as an iterator.

@tomtau is this PR blocked waiting for better documentation?

tomtau:
@Alextopher unless @tangmi plans to revive this PR soon, yes, I think two main items to fix will be to adjust the docs a little bit (to provide a full example usage) and to resolve a minor conflict with the latest target branch.
@Alextopher feel free to open a new PR based on this one.

pest/src/span.rs
```diff
@@ -62,6 +62,32 @@ impl<'i> Span<'i> {
...
+            std::ops::Bound::Excluded(&offset) => offset + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end = match range.end_bound() {
+            std::ops::Bound::Included(&offset) => Some(offset + 1),
+            std::ops::Bound::Excluded(&offset) => Some(offset),
+            std::ops::Bound::Unbounded => None,
+        };
+        let s = self.as_str();
+        if s.get(start..end.unwrap_or(s.len())).is_some() {
```
tomtau, line 80: @Alextopher one more item to fix: the merge conflict seems to be about `core` vs `std`. pest now supports `no_std`, so these `ops` may either need to be adjusted for `no_std` or sub_span should be feature-guarded for `std`

pest/src/span.rs
```diff
@@ -239,30 +275,82 @@ impl<'i> Hash for Span<'i> {
...
         if pos.at_end() {
             return None;
         }
-        let line = pos.line_of();
+        let line_start = self.pos;
         self.pos = pos.find_line_end();
-        Some(line)
+        Some(unsafe {
+            Span::new_unchecked(self.span.input, line_start, self.pos.min(self.span.end))
+        })
```
tomtau, line 310: @Alextopher could this be done without `unsafe`? if not, it should have a `Safety` comment explaining why it is safe or when things could go wrong

Alextopher:
Sounds good. I started by creating a new PR with just `Span::get` and will start looking into `LinesSpan` now.

tomtau:
closing in favour of https://github.com/pest-parser/pest/pull/681

PR #465, opened 2020-07-07
Bump `sha-1` to 0.9.

jakubadamw:


Dushistov:
It would be nice if this would be merged, because of without this PR,
I need two sha1 for my project, one sha-1 v0.9.1 and sha-1 v0.8.2 .

Issue #463, opened 2020-06-22
Very slow recursive pattern


Spu7Nix:
Parsing this expr(ession)
`a(b,{a(b,{a(b,{a(b,{a(b,{a(b,{a(b,{a(b,{})})})})})})})})`
with this grammar
```
WHITESPACE = _{ " " | "\t"}
cmp_expr = {"{" ~  expr* ~ "}"}
value = _{cmp_expr | symbol}
call = {value ~ (arguments)*}
arguments = { "(" ~ (expr ~ ",")* ~ expr? ~ ")" }
expr = {(call ~ operator)* ~ call}
operator = {"+"}
symbol = ${ASCII_ALPHA+}
```
is extremely slow (takes multiple minutes, and gets exponentially slower as you add more recursion).

SkiFire13:
When parsing `expr` your rule must parse `call` then `operator` and if it fails it has to parse `call` again. This means that for each level the number of parsing of `call` rules doubles, hence the exponential time. I'm not sure if this can be optimized in the parser, I guess it could, but for now you may want to rewrite your rules to move the optional rules at the end. The following rules should be equivalent to yours and run in linear time.

```
WHITESPACE = _{ " " | "\t"}
cmp_expr = {"{" ~  expr* ~ "}"}
value = _{cmp_expr | symbol}
call = {value ~ (arguments)*}
arguments = { "(" ~ (expr ~ ("," ~ expr)* ~ ","?)? ~ ")" }
expr = {call ~ (operator ~ call)*}
operator = {"+"}
symbol = ${ASCII_ALPHA+}
```

sbeckeriv:
Thank you for this issue. it was a very helpful example of a recursive pattern. You have helped me get unstuck!

Issue #453, opened 2020-03-29
Improving slow Rockstar parser


palfrey:
I've added some new syntax to my parser recently, and it's really slow on certain cases (10s of seconds for parsing a 10 line program for example).

Project is at https://github.com/palfrey/maiden, spec is https://github.com/palfrey/maiden/blob/master/src/rockstar.peg and one of the really bad cases is https://github.com/RockstarLang/rockstar/blob/master/tests/fixtures/arrays/arrayalike.rock

It does still eventually parse, but I'd love to try and speed it up. My guess it that it's the optional syntax in `assignable` that does it, but I've got no idea how to find out what choices I'm making slow things down so much. Are there any performance logging options in pest as I haven't seen any so far?

dragostis:
There are no performance logging options, but what I would recommend is to simply debug some of these examples manually and see what they do. PEG is deterministic and it's very easy to follow its execution.

What is probably happening in your case is that fact that the optimizer fails to find a redundant repetitions that probably lead to some locally exponential times.

SkiFire13:
Take for example the following rule:

```nor = { (or ~ "nor" ~ nor) | or }```

In the case the expression exactly matches the `or` rule this rule will have to match that rule twice. It will match it first in the `(or ~ "nor" ~ nor)`, then fail with `"nor"` and then match it again with in the `| or`. You may fix this by rewriting this rule as:

```nor = { or ~ ("nor" ~ nor)? }```

or also

```nor = { or ~ ("nor" ~ or)* }```

The same advice holds for rules like `or`, `and`, `equality_check`, `comparison`, `arithmetic`, `expression_list`, `variable_list` and maybe others too.

PR #473, opened 2020-08-20
Add more optimizations

SkiFire13:
Adds a couple more optimizations:

1. Converts `(rule ~ rest) | rule` to `rule ~ rest?`, avoiding trying to match `rule` twice. Same as the already existing `factor` but in the case the second expression is not `Expr::Seq`.

2. Converts `rule | (rule ~ rest)` to `rule` since `(rule ~ rest)` will never match if `rule` didn't. Not sure if this should go in `factorizer.rs` but the pattern looked really similar.

3. Converts `(rule ~ rest)* ~ rule` to `rule ~ (rest ~ rule)*`, avoiding matching the last `rule` twice, should have a big impact on issues like #453 and #463. Goes to a new file names `lister.rs`, not sure if the name is descriptive enough.

Actually there's a 4th optimization I thought of, converting `(rule ~ rest)* ~ rule?` to `rule ~ (rest ~ rule)* ~ rest?`, assuming `rule` is more expensive than `rest` but I don't think this can be determined if one of them contain an `Expr::Ident`. This would have the same effect of 3. on the relevant cases.


dragostis (changes requested):
Thanks a lot for taking time to write this! The change looks good, but I'd like to ask you to write some tests for this before merging.

dragostis (approved):
Neat! Again, thanks a lot for this. 👍 

bors r+

CAD97:
I'll re-r this once CI is fixed (#483), but I unfortunately really don't have time to sort through the new style warnings and apply/ignore them right now. <sub>I already spent too much time on this, grad school is really eating all of my time</sub>

CAD97 (changes requested):
Fix the merge issue and I'll merge this.

CAD97 (approved):
bors r=dragostis

Will manually merge when CI passes

MarinPostma:
bors merge

PR #480, opened 2020-10-31
Fix build after PR #454

SkiFire13:
PR #454 broke the build by adding a doctest that doesn't compile, in particular it gave the following error:

```
failures:

---- src/error.rs - error::ErrorVariant<R>::message (line 445) stdout ----
error[E0282]: type annotations needed for `pest::error::ErrorVariant<R>`
 --> src/error.rs:447:15
  |
5 | let variant = ErrorVariant::CustomError {
  |     -------   ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `R` declared on the enum `ErrorVariant`
  |     |
  |     consider giving `variant` the explicit type `pest::error::ErrorVariant<R>`, where the type parameter `R` is specified

error: aborting due to previous error

For more information about this error, try `rustc --explain E0282`.
Couldn't compile the test.

failures:
    src/error.rs - error::ErrorVariant<R>::message (line 445)

test result: FAILED. 59 passed; 1 failed; 2 ignored; 0 measured; 0 filtered out
```

This error made PRs #470's and #473's bors build fail. After this PR is merged bors should be re-run for them.

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -9,11 +9,11 @@
 
 //! Types for different kinds of parsing failures.
 
+use std::borrow::Cow;
 use std::cmp;
 use std::error;
 use std::fmt;
 use std::mem;
-use std::borrow::Cow;
 
 use position::Position;
 use span::Span;
@@ -430,9 +430,9 @@ impl<R: RuleType> ErrorVariant<R> {
     ///
     /// Returns the error message for [`ErrorVariant`]
     ///
-    /// If [`ErrorVariant`] is [`CustomError`], it returns a 
+    /// If [`ErrorVariant`] is [`CustomError`], it returns a
     /// [`Cow::Borrowed`] reference to [`message`]. If [`ErrorVariant`] is [`ParsingError`], a
-    /// [`Cow::Owned`] containing "expected [positives] [negatives]" is returned. 
+    /// [`Cow::Owned`] containing "expected [positives] [negatives]" is returned.
     ///
     /// [`ErrorVariant`]: enum.ErrorVariant.html
     /// [`CustomError`]: enum.ErrorVariant.html#variant.CustomError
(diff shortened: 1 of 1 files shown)


CAD97:
I always forget whether I have permission to

bors: r+

SkiFire13:
At least now it builds on stable, but it still fails on older versions. 

The main problem is the bump of the `sha-1` dependency's version from `0.8` to `0.9` in `pest-meta` made with PR #465, which even through [it claims to support rust 1.21](https://github.com/RustCrypto/hashes#minimum-supported-rust-version-msrv) has a transitive dependency on `generic-array v.0.14` which [requires a minimum rust version of 1.36](https://github.com/fizyk20/generic-array#generic-array) (which failed the "minimal versions" build) and a dependency on `cpuid-bool v0.1.2` which uses the macro Kleene operator, a feature stabilized in rust 1.32, and fixed-size atomics, stabilized in 1.34 (which instead triggered on the "style" build). The options here are to either revert #465 or to increase the versions used by CI to at least 1.36 but I feel like this is a decision the maintainers should make. 

Looks like `pest/src/error.rs` also needed a `cargo fmt`, that will be easy to fix.

As for why "coverage" paniced, that's a complete mistery to me.

CAD97:
This clearly improves the state of master, so I'm going to go ahead and merge it as-is currently. I think coverage has been broken for a while, but it's allowed to break because of weird issues like this (that were more prevalent when we initially set it up). I think I'm also going to set MSRV/"minimal versions" to allowed fail for the time being, so we can get master green again.

Unfortunately, I don't have time to spend actually working on projects right now, just time to review + merge what other people do.

SkiFire13:
> Unfortunately, I don't have time to spend actually working on projects right now, just time to review + merge what other people do.

If that's about PR #470 and #473, they were already reviewed by @dragostis, it's just the build that failed for this unrelated problem.

CAD97:
Yeah, I've got time to review things and will re-r those once the build is fixed, it's just new development (such as actually making CI more reliable) which I don't have time for at the moment.

PR #481, opened 2020-10-31
Bump minimal-versions test version

CAD97:
cc @SkiFire13 

We've never promised MSRV support, and the "minimal versions" build is just to make sure that our toml is actually requesting a version that we build on. So because of this, I'm just bumping the arbitrary nightly to a later arbitrary nightly (that hopefully is somewhat recent enough?); specifically, ~~I bumped the version by just under a year, so it's again about a year back from current~~.

Hopefully this fixes the build.

bors: r+

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
       - cargo clippy --all -- -D warnings
       - cargo fmt --all -- --check
     - name: minimal versions
-      rust: nightly-2018-12-08 # Arbitrary nightly for unstable cargo feature -- make this MSRV build later?
+      rust: nightly-2019-11-01 # Arbitrary nightly for unstable cargo feature -- make this MSRV build later?
       before_script:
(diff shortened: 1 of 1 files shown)


CAD97:
Bumped the version a little further; it appears we need the coherence refinement from 1.41. Nightly is still arbitrary but should hopefully have that refinement now?

bors: r+

CAD97:
That fixed minimal versions, but I missed that style needed to be bumped as well. Hopefully that doesn't mean there are style errors that need to be fixed now...

bors: r+

CAD97:
OK, remaining issues should just be applying suggested fixes from clippy:
<https://travis-ci.org/github/pest-parser/pest/jobs/740477199>

But I legitimately don't have time to get pulled into working on pest at the moment, so I'll manually merge this as-is and let someone else fix the new style warnings.

Issue #424, opened 2019-10-27
pest_bootstrap doesn't build


ksqsf:
```
➜  pest git:(master) cargo bootstrap
   Compiling pest_bootstrap v0.0.0
error[E0308]: mismatched types
  --> bootstrap/src/main.rs:24:23
   |
24 |         derive_parser(pest, false)
   |                       ^^^^ expected struct `proc_macro2::TokenStream`, found struct `quote::__rt::TokenStream`
   |
   = note: expected type `proc_macro2::TokenStream`
              found type `quote::__rt::TokenStream`
note: Perhaps two different versions of crate `proc_macro2` are being used?
  --> bootstrap/src/main.rs:24:23
   |
24 |         derive_parser(pest, false)
   |                       ^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: Could not compile `pest_bootstrap`.

To learn more, run the command again with --verbose.
```

Versions of tools that I use:
```
➜  pest git:(master) rustup toolchain list
stable-x86_64-apple-darwin (default)
nightly-x86_64-apple-darwin
➜  pest git:(master) rustc --version
rustc 1.38.0 (625451e37 2019-09-23)
➜  pest git:(master) rustc +nightly --version
rustc 1.40.0-nightly (ddf43867a 2019-09-26)
```

I'm not entirely sure why this happened, despite the fact that it also fails to build on CI. Are there any specific version requirements?

dragostis:
This is my fault. I broke the build to release an updated version. I don't really understand where the version error is coming from since I don't have a lot of experience with the new proc crates. Any help would be appreciated.

Issue #376, opened 2019-03-01
The bootstrapping breaks when used as a git dependency


Nadrieril:
I'd like to use the latest pest from the git repo, but I can't because of the bootstrapping phase.
My Cargo.toml look like
```
pest = { git = "https://github.com/pest-parser/pest" }
pest_derive = { git = "https://github.com/pest-parser/pest" }
```
which should work, but the build panics with 'Bootstrap failed because no bootstrap executable was found. Please run `cargo build --package pest_bootstrap` or `cargo bootstrap` and then try again.'
The error is clearly from the brittle bootstrapping code, but I'm not familiar enough with cargo to know how to fix it. My guess would be to use pest_bootstrap as a library instead of a binary in build.rs but there's probably a good reason it's a binary...

CAD97:
TL;DR bootstrap is hard.

Unfortunately, `pest_meta` can't depend on bootstrap directly for two main reasons:
- This introduces a cyclic dependency (`pest_meta` -> `pest_bootstrap` -> `pest` -> `pest_meta`), and
- We don't want to run the bootstrap step when building from distribution (i.e. crates-io).

A third constraint is that we don't want to put generated files into the repository if we can avoid it. It's bad practice and just asking for both merge conflicts and getting out of date.

-----

I'm thinking about ways to improve the bootstrap step. Cargo doesn't exactly make this easy, either: I wish we could `cargo bootstrap` straight from the buildscript but that causes a recursive lock attempt on the build directory, which causes deadlock.

The current workarounds:
- If you build from crates-io first and then from git, it worked once for me at least, though I have _no idea why_. (If I understand the process right, it shouldn't've.)
- Download the git repository, do `cargo bootstrap` (which is really just an alias for `cargo run --package pest_bootstrap`) and use a directory dependency.
- Do `cargo bootstrap` in a pest directory somewhere and then copy the executable to where the build script is looking for it. If I recall the process correctly, it'd be your `target/debug/pest_bootstrap/`.

If there exists a different, better solution to this problem, then I'd love to hear it. I've just not found one yet. (I'm glad my error message was clear enough, however!)

Nadrieril:
Fair enough, I've been using a local clone instead and it worked well. Thanks for the explanation !

tomtau:
with https://github.com/pest-parser/pest/commit/867da2e381b7f28db9496a5879dac67f559936f1

it should be possible to use pet as a git dependency:
```
[...]

PR #495, opened 2021-03-02
Conditional source-tree bootstrapping

iptq:
I ran into a problem when depending on my own patched version of pest, where I noticed that the bootstrapping scheme breaks when depending on source versions of pest that aren't fetched from crates.io. This is because bootstrapping creates a `grammar.rs` file in the source tree for pushing to `crates.io`, which is checked for by `meta/build.rs`. I looked at #376 and #424 (which is less relevant) but I think this patch solves some of the problems involved:

- When depending on a git version, the `meta` crate bootstraps by running `bootstrap` crate _using Cargo_ as a library. Since it's running with Cargo, the crate will be built if it doesn't exist, bypassing the hardcoded `target/debug/bootstrap` path that currently exists in the `build.rs` file. **This is the default,** since the majority of users aren't pushing to crates.io.
- When publishing to crates.io, it's preferable to have the `grammar.rs` file in the source tree, so that crates.io users won't have to bootstrap from source themselves (and also seeding for future versions of pest). To do this, pass the `--features bootstrap-in-src` flag when running the `bootstrap` crate. This will use the original logic and dump the `grammar.rs` file in the `meta/src` directory. I think there's a Cargo.toml field that you can use to always enable this feature for crates.io builds.
- As per source control conventions, no generated files are being pushed to the repository.

If there's anything I left out or need to consider additionally, please let me know. It's not too thoroughly tested except for the 2 cases that I've listed, but I hope this helps.

### DIFF
--- a/bootstrap/Cargo.toml
+++ b/bootstrap/Cargo.toml
@@ -17,3 +17,8 @@ quote = "1.0"
 codecov = { repository = "pest-parser/pest" }
 maintenance = { status = "actively-developed" }
 travis-ci = { repository = "pest-parser/pest" }
+
+[features]
+# Whether or not the bootstrapper should put the generated .rs file in the
+# source tree or in the output tree
+bootstrap-in-src = []
--- a/bootstrap/src/main.rs
+++ b/bootstrap/src/main.rs
@@ -3,17 +3,22 @@ extern crate quote;
 extern crate pest_generator;
 
 use pest_generator::derive_parser;
-use std::{fs::File, io::prelude::*, path::Path};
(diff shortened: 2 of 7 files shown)


iptq:
Whoops, looks like lints aren't very happy. I'd probably have to add `--features bootstrap-in-src` to the workflow file in order to get this to build. I'll take a look at it sometime later this week

CAD97:
Sorry for the lint issues; we accidentally upgraded the version of clippy we were enforcing warning-freedom on (see #535), so new lints got through and broke CI unexpectedly.

tomtau:
revived this PR in https://github.com/pest-parser/pest/pull/866 Thanks for the work!

PR #468, opened 2020-07-20
Update to Unicode 13

rafael-patronilo:
Updated from Unicode 11 to Unicode 13 ( #466 ). Ran all tests, all tests ok.
Unicode change logs:

> The Unicode Consortium. The Unicode Standard, Version 12.0.0, (Mountain View, CA: The Unicode Consortium, 2019. ISBN 978-1-936213-22-1)
> http://www.unicode.org/versions/Unicode12.0.0/

> The Unicode Consortium. The Unicode Standard, Version 13.0.0, (Mountain View, CA: The Unicode Consortium, 2020. ISBN 978-1-936213-26-9)
> http://www.unicode.org/versions/Unicode13.0.0/

Significant changes for this repository in both unicode updates (requires update of the Pest Book [Built-In rules](https://pest.rs/book/grammars/built-ins.html), working on an commit):
Binary Properties additions:
- BIDI_MIRRORED
- EMOJI
- EMOJI_COMPONENT
- EMOJI_MODIFIER
- EMOJI_MODIFIER_BASE
- EMOJI_PRESENTATION
- EXTENDED_PICTOGRAPHIC

PR #498, opened 2021-03-13
Make pest no_std compatible.

### LINKED ISSUE #240: No std?

01mf02:
This is another attempt to fix #240 and make pest `no_std` compatible, allowing it to be used for example in web applications via WASM. I started slowly by converting the core `pest` crate only.
All tests pass.

There is however one breaking change: In my new version, pest's `Error` type does not implement `std::error::Error`, because this trait is part of `std`. If implementing this trait is important, it could be implemented behind a feature flag.


CAD97:
In order to avoid breaking changes, please do put the std error impl behind a `std` feature flag.

Technically doing so is still breaking for anyone using `no-default-features`, but it's a breaking change that I'm much more willing to treat as "soft;" the outright removal of the impl is unquestionably objectively "hard" breaking.

01mf02:
Thanks for your suggestion, @CAD97.
I thus added a `std` feature that is enabled by default.
pest's `Error` type implements `std::error::Error` when the `std` feature is enabled.
Therefore, from a user's point of view, this version of pest should now be indistinguishable from the current master.
Furthermore, I re-enabled the `hash` test that I previously disabled.

To test `no_std`, it suffices to run `cargo test --no-default-features`.

dragostis:
Soft breaking with a `2.x` release sounds good. Thank you so much for enabling this, @01mf02.

pest/src/lib.rs
```diff
@@ -6,8 +6,8 @@
 // license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
-
 #![cfg_attr(feature = "const_prec_climber", feature(const_fn))]
+#![cfg_attr(not(feature = "std"), no_std)]
```
CAD97, line 10: Personally I prefer `#[cfg(feature = "std")] extern crate std;` to `#![cfg_attr(not(feature = "std"), no_std)]`.

Using `extern crate std;` is more clearly additive, and while it means you don't get std prelude in std areas, it also means you don't get std prelude in no_std areas, so it's a lot harder to accidentally use std types, and easier to remember to import the alloc types.

01mf02: I believe you are trying to suggest to replace `#![cfg_attr(not(feature = "std"), no_std)]` by `#[cfg(feature = "std")] extern crate std;`. But this produces a compile error:

~~~
error[E0259]: the name `std` is defined multiple times
  --> pest/src/lib.rs:67:25
   |
67 | #[cfg(feature = "std")] extern crate std;
   |                         ^^^^^^^^^^^^^^^^^ `std` reimported here
   |
   = note: `std` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
67 | #[cfg(feature = "std")] extern crate std as other_std;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
~~~

That is because the `std` crate is implicitly loaded by default, so we cannot conditionally include it. Instead, we need an explicit instruction to *disable* it. That's what `no_std` does. [Serde does it this way, for example.](https://github.com/serde-rs/serde/blob/5a8dcac2ed1407fab3f7fd23f2d56af42dcd448f/serde/src/lib.rs#L89)
I therefore think that there is no way around `#![cfg_attr(not(feature = "std"), no_std)]`.

CAD97: No, you can do both unconditional `#![no_std]` and `#[cfg] extern crate std;`.

01mf02: Ah, I understand! Thanks for pointing this out, I did not know about that before. I implemented your suggestion, see below.

01mf02:
Just to let you know: I have now also ported `derive` and `generator` to `no_std`. Please let me know how you would like me to proceed. Should I push my new changes to this branch? Or should I wait until this pull request is merged, then open another one?

01mf02:
@CAD97 & @dragostis, what is the status on this? Is there anything I can do to help my pull request getting merged?

CAD97:
Ah sorry, I meant to r+ this earlier. I don't think derive / generator need to be no_std, since they only run on the host machine? (With the new cargo feature resolver that properly isolates features, anyway.)

If you have a use case for it, feel free to open a new PR with an explanation of what it unlocks.

bors: r=@CAD97, @dragostis

CAD97:
Oh right, new clippy warnings. The quickest way to get this merged is a new PR to fix those @01mf02 , otherwise I _might_ get around to doing those eventually.

(The clippy warning should just be an `#[allow]`, the non-fmt-panic should be fixed to be a fmt panic.)

01mf02:
I just checked these clippy warnings.
They can be in principle be resolved by replacing every instance of

    #[allow(dead_code, non_camel_case_types)]

in `generator/src/generator.rs` by

    #[allow(dead_code, non_camel_case_types, upper_case_acronyms)]

However, I think that this alone will not solve the clippy warning, because the `bootstrap` package uses an version of `pest_generator` from crates.io. Therefore, I think the workflow should be something like:

1. Change `generator.rs` as outlined above
2. Bump `pest_generator` to a new version, i.e., 2.1.4, and publish on crates.io
3. Update `bootstrap/Cargo.toml` to use the new `pest_generator` version (this step might not be necessary)

I think that someone with more privileges and experience than me should check these steps. :)

01mf02:
Regarding the other clippy warning: The change required would be:

~~~ diff
diff --git a/meta/src/lib.rs b/meta/src/lib.rs
index 1f9c5bc..f049e0b 100644
--- a/meta/src/lib.rs
+++ b/meta/src/lib.rs
@@ -30,11 +30,11 @@ where
 {
     result.unwrap_or_else(|e| {
         panic!(
-            "grammar error\n\n".to_owned()
-                + &e.into_iter()
-                    .map(|error| format!("{}", error))
-                    .collect::<Vec<_>>()
-                    .join("\n\n")
+            "grammar error\n\n{}",
+            e.into_iter()
+                .map(|error| error.to_string())
+                .collect::<Vec<_>>()
+                .join("\n\n")
         )
     })
 }
~~~

Sorry, I'm reluctant to open a new PR because of the current state of my working tree.

CAD97:
Let's try this again now

bors: r+

CAD97:
Looks like bors crashed.

bors: retry

CAD97:
bors crashed twice with the same error, [I've reported it upstream](https://forum.bors.tech/t/bors-crashing-resource-not-accessible-by-integration/571).

notriddle:
I'm going to run this again. It won't work, but it will get me a request ID that I can send to github support.

bors r+

notriddle:
Opened support ticket `#1104028`

MarinPostma:
@CAD97 @notriddle, do we have any update on this? Is this an issue with Bors or Github? The link to the issue provided by @CAD97  is broken for me.

notriddle:
Nothing. 12 days ago, they said "it should be working", and haven't added anything new to the issue since.

MarinPostma:
I wonder what's the issue. We use bors on a daily basis at Meilisearch, with no issue, there must be a workaround.

testing if `bors test` works:

bors test

notriddle:
You want `bors try`, right?

CAD97:
bors: ping

bors: try

CAD97:
Immediate crash again. Immediate 403 after asking bors to try. The only information I have on what failed is `{\"message\":\"Resource not accessible by integration\",\"documentation_url\":\"https://docs.github.com/rest/reference/repos#merge-a-branch\"}`; I wish I even knew what resource fetch was failing, to allow me to even begin to try to figure out _why_ it's failing.

notriddle:
It's the actual merge operation that's failing. I manually reproduced it with `curl -v -L -H "Authorization: token ghs_[censored]" -X POST --data-raw '{"head":"ebc8e85699f55efea0fd57fddf5758640b142088","commit_message":"[ci skip][skip ci][skip netlify] -bors-staging-tmp-498","base":"staging.tmp"}' https://api.github.com/repositories/56964647/merges` while I was chatting with GitHub Support.

CAD97:
https://github.com/pest-parser/pest/pull/468 merged fine so it must be something about this PR...

I know there's some weird permission things about inline notes from actions, and as such I was surprised to see them pop up in this PR (but maybe my remembrance is outdated). I don't see any other possible reason merging this PR specifically could fail with a permissions issue.
Because I'm paranoid and #468 just merged, one more try at bors then I'll probably just give up and manually push the green button.

bors: r+

CAD97:
I've merged master into this PR to test a theory on why bors is failing: permissions requiring the branch to be up to date?

bors: r+

CAD97:
[Confirmed that this was the issue](https://app.bors.tech/repositories/4623/log#batch-139624).

@notriddle: it'd be nice if bors-ng could catch this error and report a nice "failed to merge; check your permissions" message. Unfortunately I don't have access to this repo's protection settings currently, so I don't know which one broke bors, but I have a suspicion that it's "Require branches to be up to date before merging" [as previously suggested](https://forum.bors.tech/t/badmatch-error-push-422-on-batch/341/4?u=cad97).

@dragostis: we definitely need to fix whatever permissions setting is causing bors to not work. Either you need to fix it or give me the permissions to do so.

Issue #505, opened 2021-04-01
Getting "panic message is not a string literal" from Rust 2021


applegrew:
When using Rust 2021 I am getting the following warning wherever `parses_to!` is used.

    warning: panic message is not a string literal
      --> src/parser/unit_tests.rs:11:5
       |
    11 | /     parses_to! {
    12 | |         parser: JsParser,
    13 | |         input: "10",
    14 | |         rule: Rule::numeric_literal,
    ...  |
    21 | |         ]
    22 | |     };
       | |______^
       |
       = note: this is no longer accepted in Rust 2021
       = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

applegrew:
I had to use `#[allow(non_fmt_panic)]` to turn off this warning.

nirvdrum:
I'm using Rust 2018 and I'm seeing the warning as well. It started with Rust 1.51.0, I think.

CAD97:
Yep, this is a new rustc warning that we didn't realize was a problem in the emitted code before.

CAD97:
(I won't have the bandwidth to make a PR to fix this anytime soon, likely, but I can approve a PR if someone else sets it up. I believe @01mf02 [found where the issue needs fixing](https://github.com/pest-parser/pest/pull/498#issuecomment-812881078).

CAD97:
This should be fixed by #507.

MarinPostma:
this fixed by #507

nirvdrum:
If not too much of a problem, can we please get this in a release? I can continue to use `#![allow(non_fmt_panic)]`, but this seems like the sort of thing that will be unexpected by new users.

Issue #490, opened 2021-02-12
WHITESPACE rule triggers Clippy


HollayHorvath:
The `WHITESPACE` rule has a fixed name, but it triggers the `clippy::upper_case_acronyms` lint. Since this name gets exported to Rust in the `#[grammar]` macro, this rule cannot be silenced locally.
I think the `WHITESPACE` rule should be always silent (at least I cannot imagine a situation where one would like to use this rule as-is), so I think it should not be exported to Rust at all, but I'm not familiar with the internals.

ian-h-chamberlain:
Looking at `cargo expand` output, it seems like there are already some existing lints disabled for the generated rules...
```rust
    #[allow(dead_code, non_camel_case_types)]
    pub enum Rule {
        WHITESPACE,
        Line,
        Statement,
        Expression,
        // ...
    }
```

So maybe it could just be added there in the `allow()` attribute as a quicker fix as opposed to keeping the `WHITESPACE` variant unexposed.

Meanwhile, wrapping the derive in a module is a workaround that allows me to disable the lint but avoid changing other code referencing the grammar or rules:
```rust
#[allow(clippy::upper_case_acronyms)]
mod derived {
    use pest_derive::Parser;

    #[derive(Parser)]
    #[grammar = "grammar/redcode.pest"]
    pub struct Grammar;
}

pub use derived::{Grammar, Rule};
```

CAD97:
This should be fixed by #507.

MarinPostma:
this is indeed fixed by #507

PR #507, opened 2021-04-08
fix CI workflows

MarinPostma:
This pr attempts to fix the issues encoutnered with the migration of the CI from travis to github actions. The issues that have been solved are:
- Install gcc, libssl-edv and pkg-config to the recipes that needed it.
- Install minimal rust toolchains for nightly, since they don't come with rust-fmt or clippy
- fix all clippy warn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ jobs:
         with:
           toolchain: 1.47.0 # Pinned warnings
           components: rustfmt, clippy
+      - name: Install gcc
+        run: sudo apt-get update && sudo apt-get install -y gcc
       - name: Bootstraping Grammars - Building
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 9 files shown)


CAD97:
Thanks for this!

This looks good to r=me now (fuzz tests are known failure, honestly should be disabled for now, and aren't checked by bors), but I'll hold off until you think it's good (mark the PR as not a draft).

CAD97:
I don't know if public bors supports this but may as well try

bors: d+

CAD97:
(And the full version)

bors: delegate+

MarinPostma:
@CAD97 could you re-enable fuzzing? I could run them locally using https://github.com/nektos/act, so I wonder if they run now

meiamsome:
Looks like this fixes #505 and #490

CAD97:
(sorry for the delay)

This does seem to fix fuzzing as well! The task is still being run, Bors just isn't gated on it. I'll regate bors on it later this week probably.

Thanks for your work!

bors: r+

PR #512, opened 2021-05-02
Remove fuzzing from CI

CAD97:
With fuzzit offline, it's not doing us much good, and we weren't running/checking it for PRs anyway.

bors; r+


CAD97:
I said

bors: r+

PR #513, opened 2021-05-03
Make pest_generator / pest_derive no_std compatible.

01mf02:
This is the follow-up pull request to #498.
It adds "std" feature flags for both pest_generator and pest_derive.
Currently, the correct function of this can only be tested by running `cargo test --no-default-features` *in the `derive` directory*.
Running `cargo test --no-default-features` in the root directory unfortunately does not run the tests in no_std mode.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -15,10 +15,14 @@ readme = "_README.md"
 name = "pest_derive"
 proc-macro = true
 
+[features]
+default = ["std"]
+std = ["pest/std", "pest_generator/std"]
+
 [dependencies]
 # for tests, included transitively anyway
-pest = { path = "../pest", version = "2.1.0" }
-pest_generator = { path = "../generator", version = "2.1.0" }
+pest = { path = "../pest", version = "2.1.0", default-features = false }
+pest_generator = { path = "../generator", version = "2.1.0", default-features = false }
 
 [badges]
 codecov = { repository = "pest-parser/pest" }
--- a/derive/tests/grammar.rs
+++ b/derive/tests/grammar.rs
@@ -7,6 +7,9 @@
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
 #![allow(unknown_lints, clippy)]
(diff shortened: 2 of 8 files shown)


CAD97:
Again, I prefer crates being unconditionally `#![no_std]` and then conditionally doing `extern crate std;` Other than that, everything looks good to me.

I was going to say that you can unconditionally use `::core` as a path for core items, but since we're emitting tokens with call-site spans, they'll have call-site edition hygiene, and as such won't have `core` available for edition=2015 without `extern crate core`. At least as I understand it, anyway.

01mf02:
Hi @CAD97, I made the tests unconditionally `no_std` as you suggested, thanks for spotting this.

I also tried using `::core` unconditionally, but unfortunately did not get this to work. The original patch (#243) used a similar technique.

01mf02:
Oh, I now remember why I originally did not make the tests unconditionally `#![no_std]`:
The reason for this is that `#![no_std]` makes `::core` accessible.
Therefore, indeed, somebody could change the code in `generator.rs` to use `::core::option::Option` everywhere and believe that everything is fine, because all tests (especially from `derive`) would still pass.
However, any regular user of the library that does *not* use `#![no_std]` would then see failures such as:

~~~
error[E0433]: failed to resolve: maybe a missing crate `core`?
  --> derive/tests/grammar.rs:17:10
   |
17 | #[derive(Parser)]
   |          ^^^^^^ not found in `core::option`
~~~

That is why I think it is not a good idea to unconditionally use `#![no_std]` in the `derive` tests.

01mf02:
Is there anything that blocks the merge of this PR?
Should I revert my changes that made the tests unconditionally `#![no_std]`?

01mf02:
@CAD97, @dragostis, what is the status on this?

luojia65:
Should this be merged? `no_std` pest can be used on operating system kernels that will not provide an `std` crate.

CAD97:
Sorry that this has sat for so long. If you fix the conflict, I think I can merge this. We're still waiting on @dragostis for a publish, though.

CAD97:
LGTM

bors: r+

PR #514, opened 2021-05-20
Add a function to retrieve the path from the Error struct

nkreipke:
Hi, I noticed there is currently no way to get the path set using `with_path` from `Error`. This short PR adds a function to get a reference to the currently set path.

My use case for this: I process a bunch of files and collect all parsing errors, then I'd like to filter and group them by the file that caused them. At the moment, I have the possibility to just get the location within the file, but not the path itself, because it is not public. I couldn't find a viable alternative to accomplish this.

Thank you for the great library!
Nico

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -196,6 +196,35 @@ impl<R: RuleType> Error<R> {
         self
     }
 
+    /// Returns the path set using [`Error::with_path()`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use pest::error::{Error, ErrorVariant};
+    /// # use pest::Position;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[allow(dead_code)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// # enum Rule {
+    /// #     open_paren,
+    /// #     closed_paren
+    /// # }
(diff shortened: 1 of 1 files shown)


CAD97:
This looks good to me. I'm a bit sad that this limits us to only well-formed UTF-8 paths, but a) it looks like we've already made that choice, and b) this is purely for diagnostic purposes anyway.

bors: r+

nkreipke:
Awesome!

Regarding the UTF-8 issue: I think we should be able to replace the `with_path(..., &str)` parameter with `impl AsRef<Path>` as that is also implemented for `str`. If you're interested I could provide a follow-up PR for that.

CAD97:
Unfortunately adding generic bounds to functions is a breaking change. Specifically, because the fully expanded form of the function is `<pest::error::Error::<>>::with_path::<>` — notably with an empty turbofish. Adding a new generic argument means that the empty turbofish no longer compiles, and we cannot (yet?) provide default generic types on function generic arguments.

As is, this is purely a display string, so limiting it to well-formed UTF-8 is fine. (If we could make a change, I'd be tempted to make it `impl ToString` rather than `impl AsRef<Path>`, to avoid the mandatory temporary+copy for e.g. `Path::to_string_lossy`/`Path::display`. But we can't, so moot point.)

nkreipke:
Indeed, I hadn't considered that. Thanks for the clarification.

PR #517, opened 2021-06-09
Add input() method to get the input field of a Span

sezna:
Hi! Thanks for making this great tool.

I'm writing a compiler, and I use `Span`s to keep track of error messages, warnings, source trees, etc. When it comes to rendering those messages, being able to get the reference to the original source directly from the `Span` is a really great feature -- otherwise I have to keep a mapping from `Span` to the original text and manage the lifetime differently.

If there is not a reason to exclude this method, could it be included in `Span`'s API?

Thanks,
Alex

### DIFF
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -32,6 +32,11 @@ pub struct Span<'i> {
 }
 
 impl<'i> Span<'i> {
+    /// Get the original input that this `Span` refers to without being indexed from `start` to
+    /// `end`.
+    pub fn input(&self) -> &'i str {
+        self.input
+    }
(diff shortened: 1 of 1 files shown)


CAD97:
> manage the lifetime differently

You can use the same `'i` lifetime for both the source string and the `Span`s.

> If there is not a reason to exclude this method

Semantically, `Span` is a reference only to the subslice of the span. While currently implemented as `(&str, usize, usize)`, it could also just be represented as just `(&str)`, which would not support this method.

The advantage of the latter would be smaller `Span`s, whereas the advantage of the former is having the original full input available from each `Span`.

PR #533, opened 2021-08-21
Support left recursion syntax

Abdillah:
Add support to left recursion PEG by adding marker syntax.
Inspired by this awesome paper https://www.sciencedirect.com/science/article/pii/S0167642314000288.

# Checklist
- [x] Add single rule recursive processing.
- [x] Add single rule parser tests.
- [ ] Add multiple rule recursive processing.
- [ ] Add recursive marker syntax.
- [ ] Document the syntax.

### DIFF
--- a/pest/src/iterators/queueable_token.rs
+++ b/pest/src/iterators/queueable_token.rs
@@ -13,7 +13,7 @@
 //     increased speed when pushing to the queue
 //   * it finds its pair in O(1) time instead of O(N), since pair positions are known at parse time
 //     and can easily be stored instead of recomputed
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum QueueableToken<R> {
     Start {
         end_token_index: usize,
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -53,7 +53,7 @@ pub enum MatchDir {
 /// The complete state of a [`Parser`].
 ///
 /// [`Parser`]: trait.Parser.html
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ParserState<'i, R: RuleType> {
     position: Position<'i>,
     queue: Vec<QueueableToken<R>>,
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -64,7 +64,7 @@ pub struct ParserState<'i, R: RuleType> {
     atomicity: Atomicity,
     stack: Stack<Span<'i>>,
 
-    recursive_play: bool,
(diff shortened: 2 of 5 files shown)


Abdillah:
There is still some problem about the loop, sometime it loops > 255 and gets panicked.
But, for a simple syntax, it is good enough.

Abdillah:
Hi @dragostis, this PR is in concept-complete and in minimal viable state, despite–I'm sure–still not covering edge cases especially in cases' validations. 

But, since this PR is actually violating PEG principle "non-backtracking", I want to know what's your opinion as, I believe, this project's maintainer.

oovm:
Any chance for merge this?

I want to try it, but I don't know how to configure

I have imported the dependence

```js
pest = { version = "2.1", git = "https://github.com/Abdillah/pest.git" }
```


```rs
expr = *{
    "(" ~ expr ~ ")"
  | expr ~ ("/"|"*") ~ expr
  | expr ~ ("+"|"-") ~ expr
}
```

```js
error parsing 
  --> 40:9
   |
40 | expr  = *{␊
   |         ^---
   |
   = expected `{`, `_`, `@`, `$`, or `!`

  --> 42:5
   |
42 |   | expr ~ ("*"|"/") ~ expr␊
   |     ^--^
   |
   = rule expr is left-recursive (expr -> expr); pest::prec_climber might be useful in this case
```

NoahTheDuke:
I wonder if using the algorithm described in [this](http://web.cs.ucla.edu/~todd/research/pepm08.pdf) paper would help the looping and backtracking issues.

CAD97:
Personally, I see PEG as a way to declaratively specify a _parser_, moreso than a grammar. This is because of its "first branch wins" behavior, which matches what you'd first write when writing a recursive descent parser. Any real use of pest as a parser generator is almost certainly going to want to [post process the pest parse tree to construct an abstract syntax tree](https://tomassetti.me/parse-tree-abstract-syntax-tree/). (You can provide an AST _view_ into a parse tree, but this is only realistically useful if you have a _concrete syntax tree_ (basically: lossless AST), rather than a parse tree.)

If we can support left recursion without the cost of extra looping/backtracking, as described in NoahTheDuke's linked paper or [the Pegged wiki](https://github.com/PhilippeSigaud/Pegged/wiki/Left-Recursion), I'm fine with merging support for left recursive rules, and removing the requirement that rules aren't left recursive. But if it requires backtracking, I'm against, even with colored rules. (Mainly, because how does it handle hidden/mutual/indirect left recursion?)

ethindp:
I too would appreciate this kind of feature. What blocks support of this being added?

PR #534, opened 2021-08-27
implement Copy for Position and Span

kaikalii:
I would be very nice if these types were `Copy`.
If they are not `Copy` for a reason (which I assume has to do with future-proofing), then go ahead and close this.

### DIFF
--- a/pest/src/position.rs
+++ b/pest/src/position.rs
@@ -17,7 +17,7 @@ use core::str;
 use span;
 
 /// A cursor position in a `&str` which provides useful methods to manually parse that string.
-#[derive(Clone)]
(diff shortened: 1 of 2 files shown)


CAD97:
I think at some point these contained `Rc`s, so weren't `Copy`.

bors: r+

CAD97:
Oh wait right CI is broken rn

CAD97:
Sorry for the lint issues; we accidentally upgraded the version of clippy we were enforcing warning-freedom on (see #535), so new lints got through and broke CI unexpectedly.

bors: r+

CAD97:
bors: r+

PR #537, opened 2021-09-01
Fix CI issues

Turbo87:
It looks like the GitHub Actions CI workflow is currently failing for the `master` branch. This PR fixes the two issues and makes CI pass again.

/cc @CAD97 @dragostis


CAD97:
Thanks for finishing the final bit! Maybe I can finally add the bors gate back onto the coverage build to keep it from breaking again.

bors: r+

PR #552, opened 2021-10-18
Add basic support for `node_tag` and `branch_tag`

oovm:
This idea was introduced in #550 .

This pr submitted new interfaces that does not break compatibility.

- [ParserState::tag_node]()
- [ParserState::tag_branch]()
- [Pair::as_node_tag]()
- [Pair::as_branch_tag]()

---

I think when this feature is fully prepared, the `if` here can be deleted.

https://github.com/ygg-lang/pest/blob/91133e29ab59db0e6ab7ce42b18a3b46d1ae26e2/pest/src/iterators/pair.rs#L292-L298


tomtau:
added a test in https://github.com/pest-parser/pest/pull/832

PR #553, opened 2021-10-20
Allow infix operators to prefix expressions

NoahTheDuke:
Currently, writing an expression with choice or sequence operators that is spread across multiple lines requires either appending the operator to each line except the last, or prefixing each line with the operator except the first. Both are awkward to read and use:

```pest
number = @{ '0'..'9'+ }
prefix_expr = {
    number ~ "+" ~ number
    | number ~ "-" ~ number
    | number
}
suffix_expr = {
    number ~ "+" ~ number |
    number ~ "-" ~ number |
    number
}
```

This commit seeks to solve the latter issue by allowing a single infix operator to be placed at the beginning of an expression without affecting the meaning of the expression:

```pest
number = @{ '0'..'9'+ }
expr = {
    | number ~ "+" ~ number
    | number ~ "-" ~ number
    | number
}
```

I had to remove the version requirement from the derive Cargo.toml file, as it wouldn't use the newly bootstrapped parser otherwise. (Frankly, I'm not sure exactly how to go about the bootstrap system.)


CAD97:
> I had to remove the version requirement from the derive Cargo.toml file, as it wouldn't use the newly bootstrapped parser otherwise. 

That _shouldn't_ be required; cargo should prefer the `path` dependency when working in the workspace. Either way, they're required for a publish, so we need them back 🙃

CAD97:
Should something like

```pest
rule = {
| ~ a
  ~ b
| ~ c
  ~ d
}
```

be allowed? This generally seems like a positive change, just modulo the bootstrap issues.

NoahTheDuke:
> > I had to remove the version requirement from the derive Cargo.toml file, as it wouldn't use the newly bootstrapped parser otherwise. 
> 
> 
> 
> That _shouldn't_ be required; cargo should prefer the `path` dependency when working in the workspace. Either way, they're required for a publish, so we need them back 🙃 

I have no doubt I messed up something locally so I'll add that back in.

NoahTheDuke:
@CAD97 There's also this one but it's a little more wild than the others lol.

CAD97:
I'm not comfortable merging this atm, or at least not with `~` supported. It's super unclear to me whether a leading `~` should eat trivia or not.

Also unlike `|`, you can fake a leading `~` by just having a leading `&(ANY|EOI)` lookahead, and then this clearly will eat trivia.

(Just in general I'm not super happy with how pest handles consuming trivia, because it's actually inconsistent; there are other issues about this e.g. `+` consuming trailing trivia after a single repetition but not `*`, since `e+` is actually just a macro for `e ~ e*`)

If you restrict this to just `|` (and a test that `~` isn't accepted) I'll merge this.

CAD97:
Pinging a random triager for consensus...

CAD97:
Just a test in meta that the grammar doesn't accept `~ a ~ b` would be sufficient. You could test `Parser;:parse(...).is_err()` directly or use `parses_to!` and `#[should_panic]`.

NoahTheDuke:
Thanks for the hint, I figured it out.

NoahTheDuke:
@CAD97 good to go

PR #554, opened 2021-10-20
Make keyword and builtin sets static

NoahTheDuke:
No need to initialize the keyword and builtin sets inline and then pass them around manually. Instead, define them as static (using the ~[lazy_static library](https://crates.io/crates/lazy_static)~ [once_cell library](https://docs.rs/once_cell/1.8.0/once_cell/sync/struct.Lazy.html)) and reference them "globally".


meta/src/validator.rs
```diff
@@ -165,17 +125,13 @@ pub fn validate_rust_keywords<'i>(
     errors
 }
 
-#[allow(clippy::implicit_hasher, clippy::ptr_arg)]
```
CAD97, line 168: Whoops, still need `ptr_arg` allowed here (or to change it to take a slice)

NoahTheDuke:
Okay, figured it out. These warnings don't happen in the latest version of Rust + Clippy, but Pest uses 1.56.1, from last October. If this is something worth updating, I can do that.

CAD97:
I bet I know what it is: the CI clippy is `avoid-breaking-exported-api = false` (predates that) but modern clippy won't suggest changing public API.

We definitely could use some restructuring of our CI pipeline to separate the MSRV checks from the clippy/fmt checks. See also #655.

PR #555, opened 2021-10-20
Fix all clippy warnings and errors

NoahTheDuke:
~Didn't touch any of the `allow` clippy directives, but maybe I should take a look at those too.~

Removed unnecessary clippy directives too.

PR #559, opened 2021-10-22
Refactor/optimize validation

NoahTheDuke:
(Built on top of #554 and overlaps some with #555, so once those are merged, I'll rebase to clean the history.)

I noticed that the functions called in `validate_pairs` each loop over the given `Pairs<Rule>` and build a vector of errors which are appended to a base vector. None of them rely on the full input at once (except `validate_undefined`), meaning that the internal loops can be rewritten to instead operate on a single Pair at a time, and the loop can be placed directly in `validate_pairs`.

`validate_undefined` is a little different, checking the referenced identifiers in rule expressions against the defined rule identifiers, but by inlining the referenced identifier gathering and then looping over it, we limit ourselves to two loops at most.

In doing all this, `validate_pairs` is now a little big, but I think that's okay given the limited scope of the function and the speed increases:
```
master (7422ae2549ff21ec714532aeee0f46e543e183f8):
test validate_pairs ... bench:      50,442 ns/iter (+/- 6,734)
test validate_pairs ... bench:      43,737 ns/iter (+/- 1,317)
test validate_pairs ... bench:      44,072 ns/iter (+/- 5,113)

refactor:
test validate_pairs ... bench:      30,243 ns/iter (+/- 1,063)
test validate_pairs ... bench:      32,939 ns/iter (+/- 3,940)
test validate_pairs ... bench:      31,779 ns/iter (+/- 3,547)
```

After finishing that work, my eyes fell to `validate_ast` where I saw that it "suffered" from the same repeated looping and vector building. It also had the issue of calling `to_hash_map` repeatedly (instead of defining the map once and passing it in), as well as relying on cloning to work around the borrow checker.

I didn't rearrange this as much, as each of the component functions is complex enough to warrant being left as distinct functions. Instead, I just removed the temporary vector building and extra looping, reducing error-gathering to one loop.

Once I verified that worked, I went through and cleaned up the unnecessary cloning, making a potentially breaking change to `filter_map_top_down` by making it take `self` by reference instead of by value. (This overlaps some with #555, which I will fix once that's merged too, lol. Really making a lot of work for myself here.)

With all that done, the results are quick striking (tho I suspect there's even further gains to be made):

```
master (7422ae2549ff21ec714532aeee0f46e543e183f8):
test validate_ast   ... bench:     796,490 ns/iter (+/- 46,665)
test validate_ast   ... bench:     800,321 ns/iter (+/- 80,054)
test validate_ast   ... bench:     800,774 ns/iter (+/- 49,783)

refactor:
test validate_ast   ... bench:     430,690 ns/iter (+/- 25,936)
test validate_ast   ... bench:     422,449 ns/iter (+/- 136,834)
test validate_ast   ... bench:     419,829 ns/iter (+/- 29,809)
```

### DIFF
--- a/meta/Cargo.toml
+++ b/meta/Cargo.toml
@@ -16,6 +16,7 @@ include = ["Cargo.toml", "src/**/*", "src/grammar.rs", "_README.md", "LICENSE-*"
 [dependencies]
 maplit = "1.0"
 pest = { path = "../pest", version = "2.1.0" }
+lazy_static = "1.4.0"
 
 [badges]
 codecov = { repository = "pest-parser/pest" }
--- a/meta/Cargo.toml
+++ b/meta/Cargo.toml
@@ -16,7 +16,7 @@ include = ["Cargo.toml", "src/**/*", "src/grammar.rs", "_README.md", "LICENSE-*"
 [dependencies]
 maplit = "1.0"
 pest = { path = "../pest", version = "2.1.0" }
-lazy_static = "1.4.0"
+once_cell = "1.8.0"
 
 [badges]
 codecov = { repository = "pest-parser/pest" }
--- a/meta/benches/validation.rs
+++ b/meta/benches/validation.rs
@@ -1,0 +1,43 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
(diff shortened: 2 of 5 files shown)


NoahTheDuke:
Hey @tomtau, any particular reason you want this PR reopened and merged?

tomtau:
At least those early benchmark results looked interesting. The current main bottlenecks are elsewhere though: e.g. https://github.com/pest-parser/pest/issues/560

NoahTheDuke:
Yeah this was a fun romp and I think works be beneficial overall, but I want convinced that it was worth the trade off to readability. I stopped using Pest regularly shortly after this which is why I closed it but I'd be willing to dust it off and see if it still provides "dramatic" improvements.

PR #561, opened 2021-11-02
removed unused maplit

Anton-4:
Hi,
I was trying to get my rust project to build on a M1 mac and maplit was breaking my build.
From what I can see maplit is not used, so I removed it.


CAD97:
Seems not! This is a good change to shake out the unused dep.

bors: r+

maplit is just a macro, though, so I'm really surprised that it could break compilation...

Anton-4:
> maplit is just a macro, though, so I'm really surprised that it could break compilation...

Yeah I'm not sure what went wrong, when compiling maplit, rustc errored with
`(signal: 6), SIGABRT: process abort signal)` :man_shrugging: 

Thanks for the quick merge!

PR #562, opened 2021-11-03
Don't enable "std" for ucd-trie when std not selected

andybest:
Previously, the std feature for ucd-trie was enabled by default, which was preventing builds in no_std environments. This PR only enables the std feature when std is selected for pest.


CAD97:
This is _technically_ breaking, but imo acceptable breakage, since `ucd-trie` isn't a public dependency.

bors: r+

PR #574, opened 2021-12-25
Update to unicode 14

Keruspe:
Generated by `update_unicode.sh`


tomtau:
perhaps ok to merge? Or would you like to re-generate it -- there were two patch versions of ucd-generate since then: https://github.com/BurntSushi/ucd-generate/compare/0.2.10...0.2.12 ?

CAD97:
Probably best to regenerate.

PR #586, opened 2022-03-02
chore: update dependencies by unpinning from patch version and rely on minor version

bryantbiggs:
## Changes
- Remove patch version constraint on external dependencies

## Motivation
- This allows for the latest patch version to be pulled by users which should be quite stable since the minor version is not changing
- https://github.com/dtolnay/quote/issues/204

### DIFF
--- a/bootstrap/Cargo.toml
+++ b/bootstrap/Cargo.toml
@@ -10,5 +10,5 @@ publish = false
 license = "MIT/Apache-2.0"
 
 [dependencies]
-pest_generator = "2.1.1" # Use the crates-io version, which (should be) known-good
+pest_generator = "2.1" # Use the crates-io version, which (should be) known-good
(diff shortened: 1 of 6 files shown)


bryantbiggs:
apologies, thought I had marked this as draft

good to know on the versions though, I thought it worked differently. However, I see that I can do something like `1.0.*` which will not update the minor version, but allow the patch version to update for users

anywho, need to find time to figure out the proc_macro issue, looks like its related to https://github.com/dtolnay/quote/issues/204

birkenfeld:
Why would you not want to update the minor version? By semver it's compatible, so such pinning should only be done if there's a concrete reason for it.

CAD97:
Versions are already allowed to update for users. There's no need to change anything about how pest specifies versions for this to be true.

Issue #589, opened 2022-03-08
Add context around error


zacps:
Currently if pest fails on a blank line it will output something like this:

```
 --> 8:1
  |
8 | ␍␊
  | ^---
  |
  = expected literal, single_name_part, new_indent, or continue_indent
```

I would like it to output a configurable number of surrounding lines a la rustc error messages.

Is it possible to configure this somewhere?

PR #593, opened 2022-04-12
Add a way to access the contents of an erroneous line

jwpjrdev:
If there's currently a way to do this, I'm unaware of it


tomtau:
I think it's possible to access error lines indirectly, plus there are issues that may possibly lead to changes in internal representations (e.g. https://github.com/pest-parser/pest/issues/589 ). While this is a trivial change, I'm not sure whether it's ok to make the field public, given it could break in the near future (if the indirect access isn't ok, perhaps just add a getter for &str instead?).
@CAD97 what do you think?

jwpjrdev:
Interesting. If they're fine with it, I'll do that.

CAD97:
A `.line() -> &str` would probably be the better approach, yeah.

jwpjrdev:
Had a git noob moment and accidentally ended up with this PR being closed

PR #615, opened 2022-07-10
fix: migrated to 2018 edition

### LINKED ISSUE #300: Migrate to 2018 edition

tomtau:
closes #300

the code was migrated via `cargo fix --edition`
and the grammar benchmarks were rewritten to use `bencher`
instead of the nightly test (so that they can be built
on a stable version).


glyn:
What is your rationale for migrating to the 2018 edition rather than 2021?

tomtau:
@glyn I was originally planning 2021, but it seems the current minimal version on CI doesn't support it: https://github.com/pest-parser/pest/runs/7268848504?check_suite_focus=true hence I switched it to 2018.

maybe good for @CAD97 to comment whether it's ok to bump up MSRV.

birkenfeld:
IMO: don't bump MSRV for such a minor thing, the 2021 edition probably doesn't have anything to offer `pest`.

CAD97:
If we bump MSRV it should be in a separate patch. MSRV should ideally only be increased for a specific benefit; I think it's reasonable to set an MSRV of 1.56 for the [`rust-version` manifest key](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field). 1.56 will be a year old on October 21; stable-6 is a common community definition of LTS.

For reference:

distro | version | release date
-- | --
[debian oldstable](https://packages.debian.org/buster/rustc) | 1.41 | Jan 30 2020
[debian stable] | 1.48 | Nov 19 2020
[debian nextstable](https://packages.debian.org/testing/rust/rustc) | 1.59 | Feb 24 2022
[debian unstable](https://packages.debian.org/unstable/rust/rustc) | 1.59 | Feb 24 2022
tokio (tested) | 1.49 | Dec 31 2020
tokio (guaranteed) | stable-6 (1.56) | Oct 21 2021

pest/src/span.rs
```diff
@@ -265,6 +265,16 @@ mod tests {
     use alloc::borrow::ToOwned;
     use alloc::vec::Vec;
 
+    #[test]
```
CAD97, line 268: These changes seem to have snuck in? Not an issue though, I'm not strict about a perfectly clean git history.

CAD97 (approved):
Personally I strongly prefer using criterion, but the choice to use bencher here for a smaller diff is reasonable.

Diff is the mechanical edition update, as described.

pest/src/span.rs
tomtau: Oh yes, one extra test got in

tomtau:
Thanks @CAD97 . I originally used criterion, but it failed on CI due to one of its dependencies requiring edition 2021, hence bencher.
I guess if MSRV is bumped to stable-6, it should be ok to switch to criterion.

CAD97:
It should be fine to only run benchmarks on current stable rather than MSRV, so it shouldn't fail as a dev-dependency.

PR #641, opened 2022-07-11
Migrate to `thiserror`

bobbbay:
In an effort to update this crate, I have migrated the `error` module of `pest` to `thiserror`. This removes some boilerplate impls, and makes its errors compatible with some other error reporting tools.

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -25,7 +25,9 @@ use crate::RuleType;
 
 /// Parse-related error type.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct Error<R> {
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+#[cfg_attr(feature = "std", error("An error was encountered while parsing:\n{}", self.format()))]
+pub struct Error<R: RuleType> {
     /// Variant of the error
     pub variant: ErrorVariant<R>,
     /// Location within the input string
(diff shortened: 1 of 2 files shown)


pest/src/error.rs
```diff
@@ -39,15 +40,18 @@ pub struct Error<R> {
 
 /// Different kinds of parsing errors.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub enum ErrorVariant<R> {
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
```
CAD97, line 43: Note: new impl

pest/src/error.rs
```diff
@@ -25,7 +24,9 @@ use crate::RuleType;
 
 /// Parse-related error type.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct Error<R> {
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+#[cfg_attr(feature = "std", error("An error was encountered while parsing:\n{}", self.format()))]
```
CAD97, line 28: Note: behavior change

CAD97: I believe the standard convention starts with a lowercase character, as to fit in a display of `Error: {your_error}`. See e.g. [regex's syntax error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ecd77467077c618713db50b686ff4011) ([multiline](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cdc148b1b58aab15ec7e14ce206e0a03)).

To match existing convention, I think this would be `"parse error:\n{format}"`.

bobbbay: In fact, I think this should just be `error("{}", self.format())`, because it won't change behaviour.

pest/src/error.rs
```diff
@@ -39,15 +40,18 @@ pub struct Error<R> {
 
 /// Different kinds of parsing errors.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub enum ErrorVariant<R> {
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+pub enum ErrorVariant<R: RuleType> {
     /// Generated parsing error with expected and unexpected `Rule`s
+    #[cfg_attr(feature = "std", error("{}", self.message()))]
```
tomtau, line 46: Perhaps to distinguish the variants, this could be:
```suggestion
    #[cfg_attr(feature = "std", error("parsing error: {}", self.message()))]
```

CAD97 (approved):
lgtm

Issue #333, opened 2018-11-06
pest language evolution


dragostis:
# Summary

This RFC hopes to address the concerns in #197, #261, #271, and #329 by laying the foundation of pest's evolution and transition.

# Motivation

While pest grammars offer an expressive language for building grammars, they lack certain features we've become accustomed with in programming languages which weakens their effectives as expressive and reusable tools. With the growing popularity of the project, more and more discussion has been focused on improving the predictability of pest as a language and a number of needs have been put forth: trivia handling, reusability, expressiveness, and general consistency.

## Trivia handling complexity

Probably the hardest concept to grasp when first learning the ropes is how trivia, i.e. whitespace and comments and handled. pest has an automatic mechanism that simply permits trivia to live between expressions which is controlled by atomicity. Since, atomic rules are cascading, it's not immediately obvious is two sequenced expressions `a ~ b` accept trivia—it wholly depends on whether or not the current rule inherits atomicity.

```pest
atomic                = @{ confusing }
not_atomic            =  { confusing }
definitely_not_atomic = !{ atomic }

confusing = { a ~ b }
```

The example above illustrates how `confusing` can accept trivia in some cases but not others.

## Reusability of expressions and rules

While rules can be composed from one another, there is currently no means to parametrize them. Parametrization can be extremely useful in cases where some idioms are often reused, e.g. repeated, separated values. Currently, you need to repeat some form of `e ~ ("," ~ e)*` which is less readable than `separated(e, ",")`.

Though less immediately useful, another addition would be to be able to use rules from different grammars.

## Expressiveness

Improving expressiveness is somewhat of a continuously open question. In 2.0 we've added additional stack calls that help recognize indentation-sensitive languages, namely `PEEK_ALL`, `POP_ALL`, and `DROP`. This conservative design was adopted in order to better understand what exactly is needed in real-world examples.

However, legitimate use of more refined localization within the stack has been illustrated in #329. Being able to accurately slice the stack for every one of the `PEEK`, `POP`, `DROP` calls seems to be required going forward.

## General consistency

With the introduction of built-in rules, capitalization has been selected as a form of differentiation from user-defined rules. Capitalized are also stack calls, start- and end-of-input calls, and unicode categories. The only way of differentiating between them is to simply know ahead of time what they do.

# Guide-level explanation

## Versioning

[...]

Issue #197, opened 2018-02-19
Sharing rules between grammars?


waywardmonkeys:
I have a few grammars for related formats / languages that share some common rules. Is there a way to readily share rules between separate grammar files?

It wouldn't make sense to just have a single big grammar file.

In `nom`, I would be able to define new combinators that I'd use for things like basic types and such.

dragostis:
Good point. There should be an import command in `.pest` files. This could be added in `1.1`. Let me know if you want to tackle it.

waywardmonkeys:
Could you provide a couple of getting started pointers and I will see if I can tackle it?

dragostis:
Of course. The necessary grammar should be [added to meta](https://github.com/pest-parser/pest/blob/master/meta/src/grammar.pest). Probably something like `use generic.pest` would do well.

Then this should be somehow integrated into the parser and validator in the `meta` crate. I'm not completely sure what the best approach would be in this case. My current hunch is to do the actual importing in [`parser::convert_rules`](https://github.com/pest-parser/pest/blob/master/meta/src/parser.rs#L101), where one could call parse and validate recursively, but this might not work and a change in architecture may be needed.

PR #660, opened 2022-07-22
Implementation of modules/use statements

bobbbay:
This is an initial draft implementation for modules, that I picked up about a week ago.  Hopefully, this will lead to use statements and eventually, a standard library.

## Syntax

A module can be defined similarly to Rust:

```
mod x {
  y = ...
}
```

Any element in a module should be referenced with the module prefix:

```
a = { x::y }
```

Unless imported into scope.

```
use x::y
a = { y }
```

Additionally, pest searches a default location for its standard library, which is added as the `std` module.

```
use std::...
```

### DIFF
--- a/meta/src/ast.rs
+++ b/meta/src/ast.rs
@@ -7,6 +7,8 @@
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
 
+use std::vec::IntoIter;
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Rule {
     pub name: String,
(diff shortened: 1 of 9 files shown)


meta/src/ast.rs
```diff
@@ -219,13 +236,15 @@ impl ExprTopDownIterator {
...
             right_branches: vec![],
+            module_context: None,
         };
         iter.iterate_expr(expr.clone());
         iter
     }
 
     fn iterate_expr(&mut self, expr: Expr) {
         self.current = Some(expr.clone());
+        dbg!("Running iter_expr on {:?} {:?}", &self.current, &self.next);
```
CAD97, line 247: Tracking: should be removed or replaced with a log (or tracing) `trace!`

CAD97 (commented):
👀

tomtau:
relevant issues:
https://github.com/pest-parser/pest/issues/197
https://github.com/pest-parser/pest/issues/333

Issue #661, opened 2022-07-23
Migration towards snapshot-based testing


bobbbay:
From the `cargo-insta` documentation,

> Snapshots tests (also sometimes called approval tests) are tests that assert values against a reference value (the snapshot). This is similar to how assert_eq! lets you compare a value against a reference value but unlike simple string assertions, snapshot tests let you test against complex values and come with comprehensive tools to review changes.
Snapshot tests are particularly useful if your reference values are very large or change often.

In pest, the reference values are indeed quite large, and change relatively often.

By far, the most popular snapshot-based testing library is [cargo-insta](https://github.com/mitsuhiko/insta), but RA's [expect-test](https://github.com/rust-analyzer/expect-test) also exists.

I do not have the permissions to add labels to this issue, but some labels I would like are `good-first-issue` and `help-wanted`.

CAD97:
Other potentially useful tools include 

- [trybuild](https://lib.rs/crates/trybuild): snapshot style testing for error messages
- [macrotest](https://lib.rs/crates/macrotest): snapshot style testing for macro expansions

PR #663, opened 2022-07-24
Conditionally implement `miette::Diagnostic` for `pest::error:Error`

### LINKED ISSUE #582: Support miette::IntoDiagnostic for pest::error::Error<R>
It would be nice to be able to more easily use [miette](https://docs.rs/miette/) to format diagnostics generated by pest errors. Currently, since `pest::error::Error::message` is private, this cannot easily be done by third-party libraries that depend on both pest and miette without re-implementing much of the message formatting logic in pest. It would be helpful if either `message` was made public, or if there was a feature-gated `impl<R> IntoDiagnostic for pest::error::Error<R>` to make it easier to include pest errors in miette diagnostics.

Thank you! ♥

bobbbay:
Hides an implementation of [miette](lib.rs/miette)'s `Diagnostic` type under the `miette` feature flag. Also adds example usage to the `parens` example.

The only issue I have with this is that the Display of `error::Error` changes based on this flag. Since miette displays the error *and then* the diagnostic, the error would be printed twice: once in pest's form, and once in miette's form. Hnce, `error.rs:32-33` changes the display value of `Error` if the miette feature is enabled. In my opinion, it's not a clean solution, so I'd love to hear feedback.

Note: public API change: `pest::error::Error::formt(&self) -> String` is publicized, so users on the miette feature flag can still read Pest's errors if needed.

Closes #582

### DIFF
--- a/pest/examples/parens.rs
+++ b/pest/examples/parens.rs
@@ -2,9 +2,6 @@ extern crate pest;
 
 use std::io::{self, Write};
 
-#[cfg(feature = "miette")]
-use miette::IntoDiagnostic;
-
 use pest::error::Error;
 use pest::iterators::Pairs;
 use pest::{state, ParseResult, Parser, ParserState};
@@ -78,6 +75,11 @@ fn main() {
 
         match parsed {
             Ok(pairs) => println!("{:?}", expr(pairs)),
+            // To print pest errors, use Display formatting.
+            #[cfg(not(feature = "miette"))]
+            Err(e) => eprintln!("\n{}", e),
+            // To print miette errors, use Debug formatting.
+            #[cfg(feature = "miette")]
             Err(e) => eprintln!("\n{:?}", e),
         };
     }
--- a/pest/examples/parens.rs
+++ b/pest/examples/parens.rs
@@ -2,6 +2,9 @@ extern crate pest;
 
 use std::io::{self, Write};
 
+#[cfg(feature = "miette")]
+use miette::IntoDiagnostic;
+
 use pest::error::Error;
 use pest::iterators::Pairs;
 use pest::{state, ParseResult, Parser, ParserState};
@@ -57,6 +60,7 @@ fn expr(pairs: Pairs<Rule>) -> Vec<Paren> {
         .collect()
 }
 
+#[cfg(not(feature = "miette"))]
 fn main() {
     loop {
         let mut line = String::new();
@@ -71,5 +75,20 @@ fn main() {
             Ok(pairs) => println!("{:?}", expr(pairs)),
             Err(e) => println!("\n{}", e),
         };
+    }
+}
+
+#[cfg(feature = "miette")]
+fn main() -> miette::Result<()> {
+    loop {
+        let mut line = String::new();
+
+        print!("> ");
+        io::stdout().flush().into_diagnostic()?;
+
+        io::stdin().read_line(&mut line).into_diagnostic()?;
+        line.pop();
+
+        ParenParser::parse(Rule::expr, &line)?;
     }
 }
--- a/pest/examples/parens.rs
+++ b/pest/examples/parens.rs
@@ -60,7 +60,6 @@ fn expr(pairs: Pairs<Rule>) -> Vec<Paren> {
         .collect()
 }
 
-#[cfg(not(feature = "miette"))]
 fn main() {
     loop {
         let mut line = String::new();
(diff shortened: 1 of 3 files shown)


CAD97:
> The only issue I have with this is that the Display of error::Error changes based on this flag.

The standard workaround for this would be a `.into_miette(self) -> impl Diagnostic` adapter. (`.as_miette(&self) -> impl '_ + Diagnostic` would be nicer, but having `'static` is necessary for normal use of miette. You *can* format non-`'static` diagnostics, but it's [more difficult](https://github.com/CAD97/tracing-filter/blob/v0.1/src/diagnostics.rs).)

pest/examples/parens.rs
```diff
@@ -73,3 +77,18 @@ fn main() {
...
+
+        print!("> ");
+        io::stdout().flush().into_diagnostic()?;
+
+        io::stdin().read_line(&mut line).into_diagnostic()?;
+        line.pop();
+
+        ParenParser::parse(Rule::expr, &line)?;
+    }
+}
```
CAD97, line 94: This would be better off keeping the behavior of `loop { try { ... } }`. That would be the same as the other `main`, e.g.

```suggestion
fn main() {
    loop {
        let mut line = String::new();

        print!("> ");
        io::stdout().flush().into_diagnostic()?;

        io::stdin().read_line(&mut line).unwrap();
        line.pop();

        let parse = ParenParser::parse(Rule::expr, &line);
        #[cfg(feature = "miette")]
        let parse = parse.map_err(<miette::Report as From>::from);
        match parse {
            Ok(pairs) => println!("{:?}", expr(pairs)),
            Err(e) => println!("\n{}", e),
        }
    }
}
```

(untested)

bobbbay:
> The standard workaround for this would be a .into_miette(self) -> impl Diagnostic adapter.

Sure, but what would the `impl Diagnostic` be? It can't be `Self`, because it will have the other, proper formatting as Display. Unless there's another type - `MietteError` that implements `Diagnostic` and `From<Error>` - but that overcomplicates the design.

pest/examples/parens.rs
bobbbay: Unfortunately, it seems `Report`'s display only displays a message:

```
> x

Failed to parse at (1, 1)
>
```

Without any colors, or fancy source code formatting. The real fix would probably be to use an `impl ReportHandler`, but in order to see that, I need to acquire a `&mut Formatter`...

CAD97: You can use `{:?}` here to get the return-error-from-main formatting.

To get `&mut Formatter`, you `impl Display::fmt`.

CAD97:
Personally, I don't think a non-exposed `MietteAdapter(Error)` (outside only sees `impl Diagnostic`) is a significant complication. The only difference is implementing `Display`/`Diagnostic` on `MietteAdapter` instead of `Error` directly.

For comparison, this is also the pattern used by `Path::display`.

pest/src/error.rs
```diff
@@ -524,6 +540,43 @@ fn visualize_whitespace(input: &str) -> String {
...
+
+    use super::{Error, InputLocation, RuleType};
+
+    use miette::{Diagnostic, LabeledSpan, SourceCode};
+
+    #[derive(thiserror::Error, Debug)]
+    #[error("Failure to parse at {}", self.0.line_col)]
+    pub(crate) struct MietteAdapter<R: RuleType>(pub(crate) Error<R>);
+
+    impl<R: RuleType> Diagnostic for MietteAdapter<R> {
```
tomtau, line 556: errors could also contain `path` and `continued_line` -- should those also be included in a `Diagnostic` (e.g. as additional labels) or would including those make the Miette default formatted print outs look strange?

bobbbay: Yes, it would make the miette-fornatted output strange. Adding it to the Display is fine, though.

pest/examples/parens.rs
```diff
@@ -67,9 +67,20 @@ fn main() {
...
+
+        match parsed {
             Ok(pairs) => println!("{:?}", expr(pairs)),
-            Err(e) => println!("\n{}", e),
+            // To print pest errors, use Display formatting.
+            #[cfg(not(feature = "miette"))]
+            Err(e) => eprintln!("\n{}", e),
+            // To print miette errors, use Debug formatting.
+            #[cfg(feature = "miette")]
+            Err(e) => eprintln!("\n{:?}", e),
```
CAD97, line 83: Is this true? (Does `impl Debug for Error` use the miette formatting? If not yet, doing so (when alt is false) seems like a good idea.)

CAD97 (changes requested):
I'd want to see a snapshot-style test that shows one or more instances of the miette-formatted error before merging this.

(This could be as simple as `assert_eq!(format!("{it:?}"), "expected")` or as complex as (waiting for) #661 and using insta or expect_test.)

That said, it's about time to pass the buck on pressing the green button, so...

CAD97:
(apparently I forgot to push the button to unassign the team when it chooses a load-balanced member)

pest/examples/parens.rs
bobbbay: `impl Debug for Error` does not, but `impl Debug for Report` does. And in this scenario, `e: Report`. I honestly don't thing that `Debug` for `Error` should print miette-styled messages, it should still be opt-in.

CAD97: I missed the conversion of the error into a `miette::Report` above.

bobbbay:
Great!

(blocked by #666)

prsabahrami:
Have you considered reopening this? I'm interested in working on this.

PR #666, opened 2022-07-25
Snapshot-based testing with `cargo-insta`

### LINKED ISSUE #661: Migration towards snapshot-based testing

bobbbay:
Closes #661.

### DIFF
--- a/meta/Cargo.toml
+++ b/meta/Cargo.toml
@@ -21,3 +21,6 @@ once_cell = "1.8.0"
 
 [build-dependencies]
 sha-1 = { version = "0.10", default-features = false }
+
+[dev-dependencies]
+insta = "1.16.0"
--- a/meta/src/ast.rs
+++ b/meta/src/ast.rs
@@ -281,10 +281,10 @@ mod tests {
             Box::new(Expr::Str(String::from("b"))),
         );
         let mut top_down = expr.iter_top_down();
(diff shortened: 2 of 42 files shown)


tomtau (commented):
Looks good and definitely an improvement for those textual snapshot assertions such as for errors.
I'm thinking what the debug snapshot diffs would be like (as the original assertions were sometimes on sub-expressions) and whether it'd be any better with other types of snapshot diffing... anyway, that could be explored later

CAD97:
minver patch submitted upstream: https://github.com/mitsuhiko/insta/pull/248

PR #667, opened 2022-07-28
ci: added a release flow and bumped crate versions to 2.2.0

tomtau:
the used GitHub action should hopefully figure out the correct
order for publishing the crates automatically


NoahTheDuke:
I don't know a lot about Github Actions or crates.io, so is there some way we could test this? Looks like the repo doesn't have anything about how to cut a new release. Otherwise, I don't see anything that stands out as incorrect.

tomtau:
it has the `dry-run` option: https://github.com/katyo/publish-crates#inputs
 i could perhaps push a test commit that removes the release event condition and adds this "dry-run" option? if it works, I'll revert that test commit.
or duplicate the repo and rename the crates there as "test-pest..." and try to publish those?
or any other ideas?

tomtau:
@NoahTheDuke I tried a dry-run test commit, it failed due to inconsistent versions -- after I bumped them, it seems to work: https://github.com/pest-parser/pest/runs/7563635745?check_suite_focus=true#step:6:16

NoahTheDuke (approved):
Let's see how this goes!

.github/workflows/ci.yml
```diff
@@ -131,6 +131,28 @@ jobs:
           fail_ci_if_error: false
           verbose: true
 
+  publish:
+    name: Publish crates on release
+    # Only do this job when publishing a release
+    if: github.event_name == 'release' && github.event.action == 'published'
```
CAD97, line 137: It seems better to do this in a separate workflow with its own [`on: release: types: [published]`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) trigger.

.github/workflows/ci.yml
```diff
@@ -131,6 +131,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install toolchain
+        uses: dtolnay/rust-toolchain@stable
```
CAD97, line 145: I forget; do we specify `rust-version`? If so, this needs to be that toolchain version.

.github/workflows/ci.yml
```diff
@@ -131,6 +131,28 @@ jobs:
...
+      - name: Bootstraping Grammars - Building
+        run: cargo build --package pest_bootstrap
+      - name: Bootstraping Grammars - Executing
+        run: cargo run --package pest_bootstrap
+      # it should figure out the right order: https://github.com/katyo/publish-crates#features
+      - name: Publish crates
+        uses: katyo/publish-crates@v1
+        with:
+          registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
CAD97, line 154: I forget whether publishing with the bootstrap info included requires `--allow-dirty`. I just know other crates doing similar things have required it in the past.

CAD97 (commented):
Some suggestions.

It might also make sense to set up the publish workflow to run [`on: workflow_dispatch:`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) and add a `--dry-run` flag if so.

I think `${{ (github.event_name == 'workflow_dispatch' && '--dry_run') || '' }}`.

tomtau:
Thanks @CAD97. Yes, I think having the publish workflow on a dedicated trigger is better.
As for `--allow-dirty`, dry-run seems to pass without it, but not sure whether the real publish flow would have issues without it. Also, it seems `--all-features` may be needed. Anyway, I added those two flags.

PR #670, opened 2022-07-29
ci: fix release

tomtau:
building with `--all-features` fails on 1.56 :(

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install toolchain
-        uses: dtolnay/rust-toolchain@1.56.1
+        uses: dtolnay/rust-toolchain@1.61.0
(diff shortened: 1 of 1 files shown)


tomtau:
@NoahTheDuke dry-run didn't catch this: https://github.com/pest-parser/pest/runs/7579364585?check_suite_focus=true so I tried manually building with `--all-features` and 1.61 seems to work fine

NoahTheDuke:
I wonder if we should update the `rust-version` in the Cargo.tomls to match. If we have to use a more recent toolchain to build releases, I expect that downstream users would need the same.

tomtau:
@NoahTheDuke the default features build with 1.56 -- it's for `const_prec_climber` which previously only built with nightly (now it can be built with 1.61 and above).
I'll open an issue for it though

NoahTheDuke:
Ah, it's a #feature. I didn't realize that. Cool, thanks for handling all of this!

Issue #672, opened 2022-07-29
Breaking changes in latest release


alice-i-cecile:
The most recent minor version release has breaking changes in it, which appears to have broken [tera](https://github.com/Keats/tera).

This in turn [broke our CI](https://github.com/bevyengine/bevy/pull/5484); see [this failed run](https://github.com/bevyengine/bevy/actions/runs/2761136096).

alice-i-cecile:
Upstream issue: https://github.com/Keats/tera/issues/739

NoahTheDuke:
Thanks for the report!

regar42:
I believe the culprit is here: https://github.com/pest-parser/pest/commit/723d00644e82ebe673a7c7903fd4d5038d46ab77

mockersf:
I tried fixing the path in Tera, a lot of other errors appeared after
```
error: could not compile `tera` due to 379 previous errors; 1 warning emitted
```

MarijnS95:
@regar42 I've also "bisected" this to be caused by https://github.com/pest-parser/pest/pull/522; that change doesn't account for modules/folders deeper than `$CARGO_MANIFEST_DIR/src/<the file>.pest` it seems.

regar42:
Could we maybe yank the release 2.2.0 until this is fixed? :blush:

tomtau:
> Could we maybe yank the release 2.2.0 until this is fixed? 😊

It's yanked now.

MarijnS95:
@tomtau Shall #522 be reverted so that `2.3.0` can be published? It's both a breaking change and something that I don't see how to fix in a trivial way.

tomtau:
@MarijnS95 yes: https://github.com/pest-parser/pest/pull/673
Could you try to test to use the version with the reverted commit if everything else works as before?
```
pest = { git = "https://github.com/tomtau/pest.git", rev = "39595d56729c6613e5792e0bf3e6c11d9b89d03d" }
```
I'm not sure if there are any other sneaky breaking changes.

MarijnS95:
@tomtau Tested, fixes the errors again!

Note that I couldn't use your branch as a patch as it started complain about `bootstrap`, had to check it out locally to make it bootstrap first :)

alice-i-cecile:
Thank you very much for the prompt fix!

SafariMonkey:
[...]

PR #522, opened 2021-06-30
Use the source-relative path in include_str!

tmandry:
Pest needs the full path with CARGO_MANIFEST_DIR to read the grammar
file, but when emitting `include_str!` the original path (relative to
the Rust source file) should be used.

PR #673, opened 2022-07-29
fix: undo https://github.com/pest-parser/pest/pull/522

tomtau:
closes https://github.com/pest-parser/pest/issues/672

+ bumped version numbers and split up the release flow as a workaround for a bug in katyo/publish-crates GH action


NoahTheDuke (approved):
Revert looks good to me, I'll trust you on the GH Actions change.

NoahTheDuke:
This makes me think we need some way to automate updating the version numbers lol. A lot of places to change every time we cut a new release.

MarijnS95 (approved):
Thanks a ton, tested and this fixes #672 :)

PR #681, opened 2022-08-03
Implement `Span::get`

### LINKED ISSUE #455: Expose more ways for users to manipulate Span

Alextopher:
Partially revives #456 and merge would partially close #455.

This implementation changes the name of the new feature in #456 from `Span::sub_span` to `Span::get` to have parity with `str::get`.

This implementation does not require `std`.

### DIFF
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -86,14 +86,11 @@ impl<'i> Span<'i> {
             std::ops::Bound::Unbounded => self.as_str().len(),
         };
 
-        match self.as_str().get(start..end) {
-            Some(_) => Some(Span {
-                input: self.input,
-                start: self.start + start,
-                end: self.start + end,
-            }),
-            None => None,
-        }
+        self.as_str().get(start..end).map(|_| Span {
+            input: self.input,
+            start: self.start + start,
+            end: self.start + end,
+        })
     }
 
     /// Returns the `Span`'s start byte position as a `usize`.
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -61,6 +61,41 @@ impl<'i> Span<'i> {
         }
     }
 
+    /// Attempts to create a new span within the bounds of `this` span. Will return `None` if
+    /// `range` does not fit within `this`.
+    ///
+    /// ```
+    /// use pest::Span;
+    /// let input = "Hello World!";
+    /// let world = Span::new(input, 6, input.len()).unwrap();
+    /// let orl = world.get(1..=3);
+    /// assert!(orl.is_some());
+    /// assert_eq!(orl.unwrap().as_str(), "orl");
+    /// ```
+    ///
+    /// # Examples
+    pub fn get(&self, range: impl std::ops::RangeBounds<usize>) -> Option<Span<'i>> {
+        let start = match range.start_bound() {
+            std::ops::Bound::Included(offset) => *offset,
+            std::ops::Bound::Excluded(offset) => *offset + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end = match range.end_bound() {
+            std::ops::Bound::Included(offset) => *offset + 1,
+            std::ops::Bound::Excluded(offset) => *offset,
+            std::ops::Bound::Unbounded => self.as_str().len(),
+        };
+
+        match self.as_str().get(start..end) {
+            Some(_) => Some(Span {
+                input: self.input,
+                start: self.start + start,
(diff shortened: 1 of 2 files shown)


pest/src/span.rs
```diff
@@ -61,6 +61,40 @@ impl<'i> Span<'i> {
...
+    /// use pest::Span;
+    /// let input = "Hello World!";
+    /// let world = Span::new(input, 6, input.len()).unwrap();
+    /// let orl = world.get(1..=3);
+    /// assert!(orl.is_some());
+    /// assert_eq!(orl.unwrap().as_str(), "orl");
+    /// ```
+    ///
+    /// # Examples
+    pub fn get(&self, range: impl std::ops::RangeBounds<usize>) -> Option<Span<'i>> {
```
tomtau, line 76: this may need to be feature-guarded in order not to break no_std support:
```suggestion
    #[cfg(feature = "std")]
    pub fn get(&self, range: impl std::ops::RangeBounds<usize>) -> Option<Span<'i>> {
```

pest/src/span.rs
```diff
@@ -61,6 +61,40 @@ impl<'i> Span<'i> {
...
+            std::ops::Bound::Excluded(offset) => *offset + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end = match range.end_bound() {
+            std::ops::Bound::Included(offset) => *offset + 1,
+            std::ops::Bound::Excluded(offset) => *offset,
+            std::ops::Bound::Unbounded => self.as_str().len(),
+        };
+
+        match self.as_str().get(start..end) {
```
tomtau, line 88: nit: this could potentially be rewritten using a `map`:

```rust
self.as_str().get(start..end).map(|_| Span {
                 input: self.input,
                 start: self.start + start,
                 end: self.start + end,
             }})
```

pest/src/span.rs
Alextopher: Oops! I had checked `core::ops::RangeBounds` existed but I seem to have forgotten to make the change. I switched them

pest/src/span.rs
```diff
@@ -61,6 +62,37 @@ impl<'i> Span<'i> {
...
+    pub fn get(&self, range: impl RangeBounds<usize>) -> Option<Span<'i>> {
+        let start = match range.start_bound() {
+            std::ops::Bound::Included(offset) => *offset,
+            std::ops::Bound::Excluded(offset) => *offset + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end = match range.end_bound() {
+            std::ops::Bound::Included(offset) => *offset + 1,
+            std::ops::Bound::Excluded(offset) => *offset,
+            std::ops::Bound::Unbounded => self.as_str().len(),
```
tomtau, line 86: `std` -> `core` if it exists in `core`

tomtau (commented):
this may still fail in no_std... if you'd like to test no_std, you can try to build it as follows: `rustup run nightly -- cargo build -j1 -Z build-std=core,alloc --target x86_64-unknown-linux-gnu`

Alextopher:
I have never been in a no std environment before. Actually in general this is my first Rust OSS participation.

I couldn't get your suggestion to work but https://lib.rs/crates/cargo-no-std-check seemed to do the trick to test the build for no_std.

pest/src/span.rs
Alextopher: [5c6a613](https://github.com/pest-parser/pest/pull/681/commits/5c6a613c504546c9417a06ec8144244a3e72a9c4)

tomtau (approved):

tomtau:
Thank you and congratulations on your first Rust OSS participation! 🎉

Issue #7, opened 2016-04-24
Add Packrat parser.


dragostis:

dragostis:
Currently considering the following approach:

* use procedural macros
* statically optimize parser (remove redundancy)
* consider automatically where result would be smart to cache

dragostis:
Performance measurements do not warrant this approach.

Issue #603, opened 2022-06-16
This grammar kills the online editor


mingodad:
The grammar shown bellow seems to kill the online editor:
```
Uncaught (in promise) Error: The message port closed before a response was received.
pest-site.wasm:0x569b6 Uncaught RuntimeError: unreachable
    at __rust_start_panic (pest-site.wasm:0x569b6)
    at rust_panic (pest-site.wasm:0x567a6)
    at std::panicking::rust_panic_with_hook::h51f1c855dfa04d94 (pest-site.wasm:0x564bd)
    at std::panicking::continue_panic_fmt::h728aed66c815f8d0 (pest-site.wasm:0x56385)
    at rust_begin_unwind (pest-site.wasm:0x56309)
    at core::panicking::panic_fmt::h59066e1e00f1980f (pest-site.wasm:0x58faa)
    at core::panicking::panic::h2629fd904a66558c (pest-site.wasm:0x58ed2)
    at pest_site::parse_input::h8ebf157918e4c235 (pest-site.wasm:0x86cd)
    at stdweb::webapi::window_or_worker::funcall_adapter::h0353d924de912f2e (pest-site.wasm:0x2e16)
    at Object.Module.STDWEB_PRIVATE.dyncall (pest-site.js:425:40)
$__rust_start_panic @ pest-site.wasm:0x569b6
$rust_panic @ pest-site.wasm:0x567a6
$std::panicking::rust_panic_with_hook::h51f1c855dfa04d94 @ pest-site.wasm:0x564bd
$std::panicking::continue_panic_fmt::h728aed66c815f8d0 @ pest-site.wasm:0x56385
$rust_begin_unwind @ pest-site.wasm:0x56309
$core::panicking::panic_fmt::h59066e1e00f1980f @ pest-site.wasm:0x58faa
$core::panicking::panic::h2629fd904a66558c @ pest-site.wasm:0x58ed2
$pest_site::parse_input::h8ebf157918e4c235 @ pest-site.wasm:0x86cd
$stdweb::webapi::window_or_worker::funcall_adapter::h0353d924de912f2e @ pest-site.wasm:0x2e16
Module.STDWEB_PRIVATE.dyncall @ pest-site.js:425
(anonymous) @ pest-site.js:506
setTimeout (async)
__extjs_2034053b7b6771271a2dad3e5416b045a74488a1 @ pest-site.js:506
[...]

Issue #571, opened 2021-12-12
[BUG] Pest.rs hangs indefinitely with this grammar


matthew-dean:
```
// Less extends https://www.w3.org/TR/css-syntax-3/

ws = { " " | "\t" | NEWLINE }
WHITESPACE = _{ ws* }
COMMENT = _{ multi_comment | line_comment }

multi_comment = @{ "/*" ~ (!"*/" ~ ANY)* ~ "*/" }
line_comment = @{ "//" ~ (!"\n" ~ ANY)* }

escape = @{ "\\" ~ escape_text }
escape_text = { !("\n" | ASCII_HEX_DIGIT) ~ ANY | ASCII_HEX_DIGIT{1,6} ~ ws? }

nmstart = @{ "-"? ~ (ASCII_ALPHA | "_" | NON_ASCII | escape) }
nmchar = @{ ASCII_DIGIT | ASCII_ALPHA | "_" | "-" | escape }
ident_token = @{
    ( "--" | nmstart ) ~ nmchar*
}

function_token = @{ ident_token ~ "(" }
at_keyword_token = @{ "@" ~ ident_token }
hash_token = @{ "#" ~ (nmchar | escape)+ }

string1 = @{
    "\"" ~ (
        !("\"" | "\\" | "\n") ~ ANY
        | escape
        | "\\" ~ NEWLINE 
    )* ~ "\""
}
string2 = @{
    "'" ~ (
        !("'" | "\\" | "\n") ~ ANY
        | escape
        | "\\" ~ NEWLINE 
    )* ~ "'"
}
string_token = { string1 | string2 }

url_token = @{
    "url("
    ~ ws*
    ~ (
        !("\"" | "'" | "(" | ")" | "\\" | ws | NON_PRINTABLE) ~ ANY
        | escape
        | string_token
    )*
    ~ ws*
    ~ ")"
}

NON_PRINTABLE = {
	'\u{0000}'..'\u{0008}'
	| "\u{000B}"
    | '\u{000E}'..'\u{001F}'
    | "\u{007F}"
}

number_token = @{
    ("+" | "-")?
    ~ (
        ASCII_DIGIT+ ~ ("." ~ ASCII_DIGIT+)?
        | "." ~ ASCII_DIGIT+
    )
    ~ (
        ^"e"
        ~ ("+" | "-")?
        ~ ASCII_DIGIT+
    )?
}

// or percentage_token
dimension_token = @{ number_token ~ (ident_token | "%") }
[...]

Issue #402, opened 2019-07-28
Timeout in recursion


Keats:
Found while fuzzing [Tera v1](https://github.com/Keats/tera/tree/v1).

The following expression is very slow: https://github.com/Keats/tera/blob/v1/src/tera.rs#L1028
and adding a level of recursion (eg `_(p=__(p=[_(p=__(p=[_(p=[_(p=[_1(p=[_(p=[_1`) will timeout

As far as I can see, the issue is that the rule `fn_call` is in the `basic_val` (https://github.com/Keats/tera/blob/v1/src/parser/tera.pest#L82) which is part of the potential values for a function argument -> recursing.

The issue seem to be in the generated lexer since it timeouts in the pest generated code.

Keats:
I've added a test for it in https://github.com/Keats/pest/commit/4749b31c33180345f0172748a6bbce8e88b17ed0

That also allowed me to find the issue:

https://github.com/Keats/pest/commit/4749b31c33180345f0172748a6bbce8e88b17ed0#diff-0a31f1c430e1b35cb3a08b8f7e2762c4R68 is probably ambiguous in `comparison_expr` since it can start with `fn_call`. Removing those `fn_call` in `string_concat` solves the timeout issue.

Should I do a PR with the test in the meantime @dragostis ?

dragostis:
The issue is probably related to some degenerate case that passes the optimization phase, so this is most definitely a bug. Unfortunately, I still haven't had the time to take a better look at it.

Keats:
> Unfortunately, I still haven't had the time to take a better look at it.

Don't worry about it. I'll have a look when I get some time to see if I can figure it out but I haven't looked at the internals of pest yet so that's going to take some time/effort. I've tried the debugger but wasn't sure how the breakpoints and everything worked.

Keats:
@dragostis any pointers on where to look for the bug?

PR #684, opened 2022-08-09
fix: added a call limit setting to prevent crashes or timeouts

### LINKED ISSUE #674: oss-fuzz: stack overflow in `pest::parser_state::ParserState<R>::rule`
**Describe the bug**

Fuzzing found a stack overflow. Detailed report: https://oss-fuzz.com/testcase-detail/4831410241798144

Upstream tracker: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49508

**To Reproduce**

https://google.github.io/oss-fuzz/advanced-topics/reproducing/

**Expected behavior**

No stack overflow.

**Additional context**

crash revision: e30f63e9a5b0967a2d5a2f6ef7b7d8edfc526d3f

🙃 I'm getting cc'd at my primary email and can access the oss-fuzz.com details but not the bugs.chromium.org tracker. (https://github.com/google/oss-fuzz/issues/8096)

### LINKED ISSUE #675: oss-fuzz: timout in parser
**Describe the bug**

Fuzzing found a timeout. Detailed report: https://oss-fuzz.com/testcase?key=6048143472459776

Upstream tracker: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=49513

**To Reproduce**

https://google.github.io/oss-fuzz/advanced-topics/reproducing/

**Expected behavior**

Completing within the timeout window (60s).

**Additional context**

crash revision: e30f63e9a5b0967a2d5a2f6ef7b7d8edfc526d3f

🙃 I'm getting cc'd at my primary email and can access the oss-fuzz.com details but not the bugs.chromium.org tracker. (https://github.com/google/oss-fuzz/issues/8096)

tomtau:
closes #674, closes #675


pest/src/parser_state.rs
```diff
@@ -50,6 +52,63 @@ pub enum MatchDir {
...
+/// to prevent stack overflows or excessive execution times
+/// in some grammars.
+///
+/// # Arguments
+///
+/// * `limit` - The maximum call depth or the number of calls.
+///             If 0, the call depth or the number of calls is unlimited.
+/// * `finite_calls` - If `true`, the parser execution is limited to
+///                    a finite number of calls (based on `limit`).
+pub fn set_call_limit(limit: usize, finite_calls: bool) {
```
glyn, line 68: Wouldn't the API be more natural if it took a single argument of type `Option<usize>`?

tomtau: Possibly, what would be the difference between calling it with `None` vs `Some(0)`?

tomtau: If you say a single argument, maybe the description isn't clear. For completeness, the second argument is the following:
* `true` -- it won't be decrementing the counter when exiting functions, i.e. it'll terminate after the set number of calls
(this is against timeouts when parsing potentially adversarial inputs on some grammars)
* `false` -- it'll continue executing as long as the call depth stays within the set limit
(this is if one wants to make sure it doesn't crash with "stack overflow" when parsing some inputs on some grammars)

tomtau: Maybe in practice, one will only/mostly want the former option (`true`)?

glyn: > Possibly, what would be the difference between calling it with `None` vs `Some(0)`?

`Some(0)` could be construed to mean the same as `None`, i.e. unlimited call depth. But in that case, it might be better to have a single argument of type `usize` with a value of `0` meaning unlimited call depth.

Or we could avoid attaching a special meaning to `0` and have the implementation hit the call limit immediately, i.e. before any calls.

glyn: > Maybe in practice, one will only/mostly want the former option (`true`)?

Perhaps. I guess anyone wanting an unlimited call depth could use a limit of `usize::MAX`, although that would be pretty ugly as an API call.

glyn: On balance, I thing a single argument `limit: usize` with the value of `0` meaning unlimited is pretty clean, and certainly cleaner than having a `usize` and a `bool`.

NoahTheDuke: Maybe this is my limited Java experience showing through but I would expect these to be two separate functions. I agree that it feels odd to have the bool included here, at least. Maybe the finite call fn could be `set_finite_calls() { FINITE_CALL_LIMIT.store(true, Ordering::Relaxed); }` or something similar?

I don't know that I agree that using an `Option` is helpful unless we're going to disallow `0` as a valid input: `None` or `Some(100)` but never `Some(0)`.

dragostis: Why not use `Option<NonZeroUsize>`?

dragostis:
#675 looks like it shouldn't be closed by this since it's a regression. Ideally, the optimizer should be able to detect all exponential cases and have replacement rules for them. Maybe someone could take a look at [egg](https://egraphs-good.github.io/) and implement a simplified version of the grammar there so peephole optimizations can be more easily written?

pest/src/parser_state.rs
tomtau: > Perhaps. I guess anyone wanting an unlimited call depth could use a limit of usize::MAX, although that would be pretty ugly as an API call.

usize::MAX could still cause a stack overflow in that situation; but yes, depending on the grammar, one could just set a reasonable limit of finite calls that would parse common cases and wouldn't crash on adversarial inputs

> Why not use `Option<NonZeroUsize>`?

That's a good suggestion, I think we can go for it. The downside is a bit wordy calling of that function with: `set_call_limit(Some(xxx.try_into().unwrap()),...)` instead of `set_call_limit(xxx, ...)`

tomtau:
@dragostis thanks. Yes, I agree that for the timeout issue, there should be a fix for the root cause. For the edge case found by the fuzzer, it seems it's due to the block comment feature introduced in this PR: https://github.com/pest-parser/pest/pull/332
There are, however, a few other open issues related to performance which may or may not have the same root cause, e.g. https://github.com/pest-parser/pest/issues/603 https://github.com/pest-parser/pest/issues/571 https://github.com/pest-parser/pest/issues/402 https://github.com/pest-parser/pest/issues/279

Given the investigation and the corresponding proper backwards-compatible fix for timeouts may take time, there are potentially two ways (with regards to this API: https://github.com/pest-parser/pest/pull/684#discussion_r941397159 ) to keep `set_call_limit` with a single argument:

1. make it track overall calls (`true`). This should fix those fuzzer issues, as demonstrated by the unit tests.
2. make it track call depths (`false`). This should fix only the stack overflow crashes, but not the timeout issues.

I'm leaning towards the first one, because, in the meantime, it's practical in the sense it'd allow people to use pest on potentially problematic inputs (without worrying it'd eat up all CPU time) if their grammars suffered from this issue.
I'd then open up an issue that would reference the fuzzer timeout issue as well the other performance issues.

--
BTW, as for the solution, one potential option that comes to mind is packrat parsing with a fix: https://dl.acm.org/doi/10.1145/3377555.3377898 There's one older issue that was closed https://github.com/pest-parser/pest/issues/7#issuecomment-323311318 -- do you have any more insight into why packrat parsing was not used?

dragostis:
Packrat parsing was not used because caching data is slow. While it removes the exponential cases, it adds substantial overhead for all other parsing.

My solution to this was to identify such patterns and replace them in the optimizer. This has worked for some cases, but I haven't had the time to identify all patterns of interest. Also, I have no proof that all such patterns are actually replaceable with equivalent but non-exponential patterns.

My vision for pest when I started it was to actually use something like [ruler](https://github.com/uwplse/ruler) to find these rules and use egg to optimize the grammars. Currently, I think this is a bit hard since pest's grammar is in need of a cleanup. I started working on a new meta grammar in [pest3](https://github.com/pest-parser/pest3) a few years ago, but didn't have a chance to make it work. It has a few interesting ideas, including much better code generation, if you have the time to take over that work.

tomtau:
@glyn The API was simplified to only track calls as per the discussion: https://github.com/pest-parser/pest/pull/684#discussion_r941383935

`Option<NonZeroUsize>` makes it a bit wordy with literals, but that may be improved in future Rust versions:
https://github.com/rust-lang/rust/issues/69329

pest/src/parser_state.rs
```diff
@@ -50,6 +53,48 @@ pub enum MatchDir {
...
+/// Sets the maximum call limit for the parser
+/// to prevent stack overflows or excessive execution times
+/// in some grammars.
+///
+/// # Arguments
+///
+/// * `limit` - The maximum call depth or the number of calls.
+///             If 0, the call depth or the number of calls is unlimited.
+/// * `finite_calls` - If `true`, the parser execution is limited to
+///                    a finite number of calls (based on `limit`).
```
glyn, line 67: These docs don't match the API.

glyn (changes requested):
Please update the docs of `set_call_limit`.

pest/src/parser_state.rs
tomtau: oh, thanks!

pest/src/parser_state.rs
```diff
@@ -61,10 +61,8 @@ static CALL_LIMIT: AtomicUsize = AtomicUsize::new(0);
 ///
 /// # Arguments
 ///
-/// * `limit` - The maximum call depth or the number of calls.
-///             If 0, the call depth or the number of calls is unlimited.
-/// * `finite_calls` - If `true`, the parser execution is limited to
-///                    a finite number of calls (based on `limit`).
+/// * `limit` - The maximum number of calls. If None,
+///             the number of calls is unlimited.
```
glyn, line 65: That's better thanks. I wonder if you could clarify in the doc comment whether "number of calls" means the same as call depth or is just a running total. Also, please define what constitutes a "call"?

glyn (approved):
Remove blocking review. Please note I have only reviewed a small part of this PR.

pest/src/parser_state.rs
tomtau: After https://github.com/pest-parser/pest/pull/684#issuecomment-1210456883 , it's a running total. I've expanded the doc comment.

glyn (approved):
LGTM, thanks.

pest/src/parser_state.rs
```diff
@@ -50,6 +53,49 @@ pub enum MatchDir {
...
+impl CallLimitTracker {
+    fn limit_reached(&self) -> bool {
+        self.current_call_limit
+            .map_or(false, |(current, limit)| current >= limit)
+    }
+
+    fn increment_depth(&mut self) {
+        if let Some((current, limit)) = self.current_call_limit {
+            self.current_call_limit = Some((current + 1, limit));
+        }
```
NoahTheDuke, line 95:
```suggestion
        if let Some((current, _)) = &mut self.current_call_limit {
            *current += 1;
        }
```

NoahTheDuke (approved):
Left a suggestion but take it or not, this is a good change.

pest/src/parser_state.rs
```diff
@@ -50,6 +53,49 @@ pub enum MatchDir {
...
+impl CallLimitTracker {
+    fn limit_reached(&self) -> bool {
+        self.current_call_limit
+            .map_or(false, |(current, limit)| current >= limit)
+    }
+
+    fn increment_depth(&mut self) {
+        if let Some((current, _)) = &mut self.current_call_limit {
+            *current += 1;
+        }
```
glyn, line 95: @tomtau @NoahTheDuke 
Just for interest, there's yet another way to code this:
```
        if let Some((ref mut current, _)) = self.current_call_limit {
            *current += 1;
        }
```
This cropped up [here](https://youtu.be/rAl-9HwD858?t=2752) in @jonhoo's video [Crust of Rust: Lifetime Annotations](https://www.youtube.com/watch?v=rAl-9HwD858).

NoahTheDuke: Oh that's cool. I didn't know you could embed those.

PR #700, opened 2022-08-20
ci: release 2.3.0

tomtau:
added a custom release bash script, because the previous GH action
failed at different stages, so a part of it needed to be published
manually. we'll see if this works this time.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+set -e
+
+CRATES="pest pest_meta pest_vm pest_generator pest_derive pest_grammars"
+
+get_local_version() {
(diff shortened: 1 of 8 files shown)


release.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+wait_until_available() {
+  echo "Waiting for crate ${1} to become available via crates.io..."
+  for retry in {1..5}; do
+    sleep 5
+    ONLINE_DATE="$(check_version_online "${1}" "${2}")"
+    if [ -n "${ONLINE_DATE}" ]; then
+      echo "Crate ${crate} is now available online"
+      break
```
NoahTheDuke, line 32: If this line is hit, it will still wait 10 seconds at the end?

tomtau: yes

NoahTheDuke (approved):
I'm cautiously optimistic!

PR #702, opened 2022-08-23
fix: Search for grammar file from CARGO_MANIFEST_DIR

### LINKED ISSUE #325: Grammar file not found on workspace
I use pest in a crate of a workspace, so this crate has no `src` folder. I was forced to create an empty `src` subfolder in my crate so that pest could find my file, it's not clean.

It's easy to reproduce this bug, create a new workspace project and try to import pest grammar file from a crate of workspace.

Nukesor:
This is a proposeal to fix #325.

The new logic looks for the `[grammar = "$path"]` path in `CARGO_MANIFEST_DIR/` before using the fallback of `CARGO_MANIFEST_DIR/src`.

This allows projects that don't have a `src` folder to use a grammar file, while staying backward compatible.

**Alternative implementation approach:**

One could theoretically use the [absolute](https://doc.rust-lang.org/std/path/fn.absolute.html) function.
Sadly, this function is still nightly, which makes this approach unfeasible.

**Current workarounds:**
- Restructure the project so it has a `src` folder
- Create a stub `src` folder, so one can use relative paths
- Inline all of the grammar via `grammar_inline`

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -41,7 +41,21 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
     let (data, path) = match content {
         GrammarSource::File(ref path) => {
             let root = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
-            let path = Path::new(&root).join("src/").join(&path);
+
+            // Check whether we can find a file at the path relative to the CARGO_MANIFEST_DIR
+            // first.
+            //
+            // If we cannot find the expected file over there, fallback to the
+            // `CARGO_MANIFEST_DIR/src`, which is the old default and kept for convenience
+            // reasons.
+            // TODO: This could be refactored once `std::path::absolute()` get's stabilized.
+            // https://doc.rust-lang.org/std/path/fn.absolute.html
+            let path = if Path::new(&root).join(&path).exists() {
+                Path::new(&root).join(&path)
+            } else {
+                Path::new(&root).join("src/").join(&path)
+            };
+
             let file_name = match path.file_name() {
                 Some(file_name) => file_name,
                 None => panic!("grammar attribute should point to a file"),
--- a/derive/tests/grammar.rs
+++ b/derive/tests/grammar.rs
@@ -16,7 +16,7 @@ extern crate pest;
 extern crate pest_derive;
 
 #[derive(Parser)]
-#[grammar = "../tests/grammar.pest"]
+#[grammar = "tests/grammar.pest"]
 struct GrammarParser;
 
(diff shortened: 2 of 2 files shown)


generator/src/lib.rs
```diff
@@ -41,7 +41,19 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
...
         GrammarSource::File(ref path) => {
             let root = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
-            let path = Path::new(&root).join("src/").join(&path);
+
+            // Check whether we can find a file at the path relative to the CARGO_MANIFEST_DIR
+            // first.
+            //
+            // If we cannot find the expected file over there, fallback to the
+            // `CARGO_MANIFEST_DIR/src`, which is the old default and kept for convenience
+            // reasons.
```
tomtau, line 50: maybe can add a TODO with a link to https://doc.rust-lang.org/std/path/fn.absolute.html ?

tomtau (approved):
thanks! I think that works -- it seems both options are covered by tests, so it should be fine. CC @MarijnS95 this fix should be backwards-compatible, but you can have a quick look.

generator/src/lib.rs
Nukesor: Done :)

Nukesor:
Hey @tomtau :) 

I kind of need this fixe for a [feature](https://github.com/Nukesor/pueue/pull/353) I currently work on.
It would be great to get a patch release that contains this fix. Do you have a rough timeline when the next release is planned?

tomtau:
@Nukesor I think it's all right to make a quick point release after this is merged (plus perhaps after one small issue fix): https://github.com/pest-parser/pest/pull/705 (to test the fixed CI release flow).

Nukesor:
Awesome, thanks for your swift response and for taking care of this project :)

MarijnS95:
> CC @MarijnS95 this fix should be backwards-compatible, but you can have a quick look.

@tomtau Fwiw I haven't looked into `pest` in detail but only got blocked on the previous release accidentally breaking transitive dependencies while updating them. Any contributor that intends to make such changes as part of a patch release should make sure to (build-)test those broken crates to make sure this change can safely be made without a breaking release.

tomtau:
@Nukesor it's out now: https://crates.io/crates/pest/2.3.1

@MarijnS95 I included this flow using `cargo-semver-checks` https://github.com/pest-parser/pest/blob/master/semvercheck.sh to check it didn't introduce (API-level) breaking changes, but it wouldn't probably catch that v2.2.0 breaking change (it'd catch the one in v2.2.1 though).

PR #704, opened 2022-08-27
Use new `sha-1` package name

davidkna:
`sha1` is the primary name now.


tomtau (approved):
ref: https://github.com/RustCrypto/hashes/pull/350

PR #710, opened 2022-09-20
feature: unary prefix/suffix operator support via PrattParser

### LINKED ISSUE #461: make PrecClimber support prefix and postfix operators
## Problem

The PrecClimber handles precedence between infix operators, like `a + b * c`, but doesn't handle precedence between prefix or postfix operators, like `-a.b + c`.

## Expectation

The PrecClimber should be able to resolve prefix and postfix operators along with infix operators.

tomtau:
Closes #461

based on #344

Co-authored-by: @segeljakt

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "pest_derive"
 description = "pest's derive macro"
-version = "2.3.1"
+version = "2.4.0"
 edition = "2018"
 authors = ["Dragoș Tiselice <dragostiselice@gmail.com>"]
 homepage = "https://pest-parser.github.io/"
@@ -23,5 +23,5 @@ std = ["pest/std", "pest_generator/std"]
 
 [dependencies]
 # for tests, included transitively anyway
-pest = { path = "../pest", version = "2.3.1", default-features = false }
-pest_generator = { path = "../generator", version = "2.3.1", default-features = false }
+pest = { path = "../pest", version = "2.4.0", default-features = false }
+pest_generator = { path = "../generator", version = "2.4.0", default-features = false }
--- a/derive/examples/calc.pest
+++ b/derive/examples/calc.pest
@@ -1,0 +1,16 @@
+WHITESPACE   =  _{ " " | "\t" | NEWLINE }
+
+ program      =   { SOI ~ expr ~ EOI }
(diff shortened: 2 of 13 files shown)


tomtau:
@pest-parser/maintainers is it ok to merge and release it? It's non-breaking, but not sure if marking `PrecClimber` deprecated right away is ok

NoahTheDuke:
Sorry, been busy. Is there any reason to deprecate the old parser instead of letting it live side by side?

tomtau:
> Sorry, been busy. Is there any reason to deprecate the old parser instead of letting it live side by side?

because it's a duplicate -- `PrattParser` can do what `PrecClimber` can do, but is more general (so it can handle more cases, such as unary ops)

NoahTheDuke:
If it's purely additive, is there any reason to rename it? Or is the api fundamentally different somehow?

tomtau:
> If it's purely additive, is there any reason to rename it? Or is the api fundamentally different somehow?

The API is different. See e.g. the calculator test: https://github.com/pest-parser/pest/pull/710/files#diff-e2dbd27540c35bdae26743d1b758d99b1186ec63a0be6abc83dfd4ca70c571edR132

tomtau:
(plus the original PR's comments: https://github.com/pest-parser/pest/pull/344#issuecomment-489008650 )

NoahTheDuke (changes requested):
Okay, I've had a chance to read through the code. I'm glad to see that the changes to Pest internals are minimal. Overall, this is a good PR and I think it should be merged. Two thoughts:

This might be my Clojure background coming out but I think we should make it clear that we're not going to remove PrecClimber in a future version, just marking it as less powerful/useful than PrattParser. It's important that people can trust the code they write today will still work in the future, even across multiple versions.

I also think that the internal method names `nud`, `led`, and `lbp` are quite terse. I have no idea what they stand for, lol. Given that they're only used internally and only in one place, I think they should get more descriptive names, ideally with at least minimal docstrings.

Thanks so much for getting this cleaned up!

NoahTheDuke (approved):

tomtau:
@NoahTheDuke thanks! I extended the deprecation note.
As for the internal methods, I kept them, but added doc strings on them to explain what they stand for. While those terse names are a bit scary, they are ubiquitous in Pratt parser implementations (given that's what the [original 1973 paper](https://web.archive.org/web/20151223215421/http://hall.org.ua/halls/wizzard/pdf/Vaughan.Pratt.TDOP.pdf) used).

PR #705, opened 2022-08-29
ci: better CI hygiene

tomtau:
closes #655

PR #654, opened 2022-07-15
chore: switch benches to criterion plus MSRV bump to 1.56

tomtau:
as per https://github.com/pest-parser/pest/pull/615#issuecomment-1179755352
besides criterion, it is now possible to compile the
`const_prec_climber` feature on stable, because
`const_fn_trait_bound` was stabilized

PR #715, opened 2022-10-04
Update deps and swith to 2021 edition

lwandrebeck:
cargo bootstrap and build run fine.
Update to 2021 edition.
Update to latest ucd-trie, serde, serde_json, thiserror.

### DIFF
--- a/bootstrap/Cargo.toml
+++ b/bootstrap/Cargo.toml
@@ -2,7 +2,7 @@
 name = "pest_bootstrap"
 description = "pest bootstrap script"
 version = "0.0.0"
-edition = "2018"
+edition = "2021"
 authors = ["Dragoș Tiselice <dragostiselice@gmail.com>"]
 homepage = "https://pest-parser.github.io/"
 repository = "https://github.com/pest-parser/pest"
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -2,7 +2,7 @@
 name = "pest_derive"
 description = "pest's derive macro"
 version = "2.4.0"
-edition = "2018"
+edition = "2021"
 authors = ["Dragoș Tiselice <dragostiselice@gmail.com>"]
 homepage = "https://pest-parser.github.io/"
 repository = "https://github.com/pest-parser/pest"
--- a/generator/Cargo.toml
+++ b/generator/Cargo.toml
@@ -2,7 +2,7 @@
 name = "pest_generator"
 description = "pest code generator"
 version = "2.4.0"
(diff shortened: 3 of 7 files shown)


lwandrebeck:
Some errors do appear here: https://github.com/pest-parser/pest/actions/runs/3180773049/jobs/5184774281 even though test pass properly. Digging into it.

tomtau (approved):
lgtm; I tried to run `cargo fix --edition` https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html and no files were changed, so this should be fine as it is.

The semver error looks like the one I mentioned in the third point here: 
https://github.com/pest-parser/pest/pull/705#issuecomment-1235712075 because those reported violations on `pest_meta` look to be false positives.

lwandrebeck:
@tomtau : I’ve ran it and got no change too. But, panic! is supposed not to be used as-is in 2021 edition. Maybe something slipped through the cracks between documentation and actual code ?

NoahTheDuke:
I don't remember all of the details about editions; will this have any effects on MSRV?

lwandrebeck:
1.56 brings 2021 edition, and we’re at 1.56.1 according to README.md
https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html

tomtau:
There are some cases where the automated fixes are not applied: https://doc.rust-lang.org/edition-guide/editions/advanced-migrations.html#migrating-macros (macros, features, generated code, plus some fixes are optional, so it needs the `--edition-idioms` flag).
@lwandrebeck are you sure about the panic fix? My understanding of https://doc.rust-lang.org/edition-guide/rust-2021/panic-macro-consistency.html is that 2021 makes `panic!` usage to be consistent with `println!`, and that original line was consistent with it (hence there wasn't any automated fix)?

tomtau:
For the context, previously MSVR was lower than 1.56, hence the edition was only up to 2018: https://github.com/pest-parser/pest/pull/615#issuecomment-1179670261 (but it was later increased to 1.56: https://github.com/pest-parser/pest/pull/654 )

lwandrebeck:
@tomtau : I’ve never put my hands _at all_ into macros to be fair, so that’s out of my league. I’ve just re-run `cargo fix --edition --edition-idioms --allow-dirty` after putting back 2018 in `Cargo.toml` files without getting any change once again.
I’ll try to dig a bit more into that subject, but that won’t be until next day at best. If anyone is willing to check that point, go ahead :)

PR #372, opened 2019-02-18
Adding HTTP grammar

adamAndMath:
I've tried to make a http request grammar file, as mentioned in [#210](https://github.com/pest-parser/pest/issues/210). It doen't restrict the allowed characters in tokens as much as they should be, as I don't know how to properly express the range of allowed characters.
It is currently untested as I don't have much experience with the format, and don't quite know what I should and shouldn't test.

PR #726, opened 2022-10-21
feature: added http request grammar

tomtau:
plus tests, benchmark and a fuzzing target

based on #372 by @adamAndMath


tomtau:
(note that the grammar is a bit more permissive than it should be, but probably ok for this sample grammar testing or it could be refined in a follow up PR)

NoahTheDuke (approved):
I don't think 100% accuracy is needed for fuzzing targets, and we can tighten it up later down the road.

Issue #551, opened 2021-10-18
README.md and website example grammar doesn't handle leading digit for all idents in ident_list


SeedyROM:
The current example grammar will allow for every `ident` after the first to start with a `digit`.

For instance `"a1 2b"` is a valid `ident_list` with the current grammar.

Wouldn't this make sense for the ident rule?

```pest
...
ident = { !digit ~ (alpha | digit)+ }

ident_list = _{ ident ~ (" " ~ ident)+ }
```
I'm new to PEG style grammars so maybe I'm wrong?

Alextopher:
Commenting since it's a "good-first-issue". 

The issue is correct the example grammar does allow every `ident` after the first to start with a digit. The text above the example in the REAME says as much (both Today and in Oct 2021).
> The following is an example of a grammar for a list of alpha-numeric identifiers where the first identifier does not start with a digit:

So I believe this issue may be closed with `wont-fix`. 

However, I don't think the confusion is unwarranted, I can not think of a time I wanted an "identifier list but the first identifier is not like the others." Looking at git blame I can see at some point in the past the example matched the suggested grammar. 

I also recall discovering https://pest.rs, copying the example into the Editor, and feeling confused as to why `a _` did not return an error while `_` does. After reading the book I understand it's the lack of any EOI requirement. 

Switching things around like this also means
```
ident_list = _{ ident ~ (" " ~ ident)+ }
```
should be
```
ident_list = _{ ident ~ (" " ~ ident)* }
                                     ^
```

If there is interest in updating the example I would be willing to create a draft PR. Otherwise, I think this issue should be closed.

tomtau:
@Alextopher feel free to open PRs to improve the examples in README here and the book / site repositories.

PR #728, opened 2022-10-31
Update README.md

carnoxen:
Fix a pest example. ```ident``` rule is more evidently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ where the first identifier does not start with a digit:
 alpha = { 'a'..'z' | 'A'..'Z' }
 digit = { '0'..'9' }
 
-ident = { (alpha | digit)+ }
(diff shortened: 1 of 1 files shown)


tomtau (commented):
Thank you for fixing the example!
I assume this tries to fix the following issue: https://github.com/pest-parser/pest/issues/551 , is that correct ? 

if the fix is done in that way, it'll prevent all identifiers to start with a digit, not only the first one... so perhaps that description:
> (...) where _the first identifier_ does not start with a digit:

should be updated to:
> (...) where identifiers do not start with a digit:

carnoxen:
> Thank you for fixing the example! I assume this tries to fix the following issue: #551 , is that correct ?

Yes it is. And I fixed the above sentence.

PR #733, opened 2022-11-09
chore: clippy fixes and ignores

tomtau:
code was fixed where possible;
elsewhere it'd need API-breaking changes,
so it was just annotated to allow that for now.

### DIFF
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -40,6 +40,7 @@ impl Vm {
     }
 
     /// Runs a parser rule on an input
+    #[allow(clippy::perf)]
     pub fn parse<'a, 'i>(
         &'a self,
         rule: &'a str,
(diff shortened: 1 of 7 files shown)


vm/src/lib.rs
```diff
@@ -48,6 +49,7 @@ impl Vm {
         pest::state(input, |state| self.parse_rule(rule, state))
     }
 
+    #[allow(clippy::almost_complete_letter_range)]
     fn parse_rule<'a, 'i>(
```
tomtau, line 53: @pest-parser/triage One other possibility for this would be to introduce a new `pub fn match_range_inclusive(mut self: Box<Self>, range: RangeInclusive<char>) -> ParseResult<Box<Self>>` and use that in this function, and mark the old `match_range` as deprecated:

https://github.com/pest-parser/pest/blob/df487a6841c5d9e2990d13c4cc1254611efa3912/pest/src/parser_state.rs#L685

glyn (approved):
LGTM, thanks.

PR #736, opened 2022-11-20
feature: added a `pest_debugger` crate (fixes #98)

### LINKED ISSUE #98: Grammar backtrace

tomtau:
based on the old PR by @dragostis: https://github.com/pest-parser/pest/pull/277

Changes that were made:
- debugger core context was refactored and extracted to a lib (so that it could be used in other frontends, e.g. editor plugins)
- CLI was extended using rustyline helpers to provide file completions, history etc.
- applied suggestions from @hansihe from the old PR (https://github.com/pest-parser/pest/pull/277#issuecomment-421838669):
1. added `ba` (add breakpoints at all rules) which is useful for stepping through the entire grammar, plus breakpoint deletions and loading input directly from readline;
2. added command line arguments.
- changed the listener function to return a boolean, so that the debugger can signal back to a parsing thread to finish before reaching its input's EOF.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "bootstrap",
+    "debugger",
     "derive",
     "generator",
     "grammars",
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,7 @@
 comment: off
 
 ignore:
+ - debugger/src/main.rs
  - derive/src/lib.rs
  - pest/examples/parens.rs
  - pest/src/unicode/*
--- a/debugger/Cargo.toml
+++ b/debugger/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "pest_debugger"
+description = "pest grammar debugger"
+version = "2.5.0"
+edition = "2021"
+authors = ["Dragoș Tiselice <dragostiselice@gmail.com>", "Tomas Tauber <me@tomtau.be>"]
+homepage = "https://pest.rs/"
+repository = "https://github.com/pest-parser/pest"
+documentation = "https://docs.rs/pest"
+keywords = ["pest", "grammar", "debugger"]
+categories = ["parsing"]
+license = "MIT/Apache-2.0"
+readme = "_README.md"
+rust-version = "1.56"
+
+[dependencies]
+pest = { path = "../pest", version = "2.5.0" }
+pest_meta = { path = "../meta", version = "2.5.0" }
+pest_vm = { path = "../vm", version = "2.5.0" }
+rustyline = "10"
+thiserror = "1"
+
+[badges]
+codecov = { repository = "pest-parser/pest" }
+maintenance = { status = "actively-developed" }
+github = { repository = "pest-parser/pest" }
--- a/debugger/_README.md
+++ b/debugger/_README.md
@@ -1,0 +1,1 @@
+../README.md
--- a/debugger/src/lib.rs
+++ b/debugger/src/lib.rs
@@ -71,7 +71,11 @@ use std::{
 };
 
 use pest::{error::Error, Position};
-use pest_meta::{optimizer::OptimizedRule, parse_and_optimize, parser::Rule};
+use pest_meta::{
+    optimizer::OptimizedRule,
+    parse_and_optimize,
+    parser::{rename_meta_rule, Rule},
+};
 use pest_vm::Vm;
 
 /// Possible errors that can occur in the debugger context.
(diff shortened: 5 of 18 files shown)


tomtau:
Hm, two unintended semver violations:
```
Description:
A publicly-visible function now takes a different number of parameters.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#fn-change-arity
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.13.0/src/queries/function_parameter_count_changed.ron

Failed in:
  pest_meta::validator::validate_undefined now takes 2 parameters instead of 3, in meta/src/validator.rs:186
  pest_meta::validator::validate_pest_keywords now takes 1 parameters instead of 2, in meta/src/validator.rs:141
  pest_meta::validator::validate_rust_keywords now takes 1 parameters instead of 2, in meta/src/validator.rs:120
```

```
--- failure auto_trait_impl_removed: auto trait no longer implemented ---

Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.13.0/src/queries/auto_trait_impl_removed.ron

Failed in:
  type Vm is no longer Sync, in vm/src/lib.rs:37
  type Vm is no longer RefUnwindSafe, in vm/src/lib.rs:37
  type Vm is no longer Send, in vm/src/lib.rs:37
```

tomtau:
The first semver violation is unrelated to the current PR, it's due to this older change: https://github.com/pest-parser/pest/pull/554/files that slipped the older semver violation checks... shall we add back the original fn arity and just pass it the statics, or leave it as it is (those functions aren't probably externally depended on / they looked that they shouldn't have been made public anyway)?

The second semver violation is legit... I guess two possible ways around it are: 1) create a new wrapper struct (e.g. `TracingVm`) that will have the listener callback and isn't `Send + Sync + RefUnwindSafe`, 2) wrap the `listener` field in `Arc` and `Mutex`?
@NoahTheDuke any thoughts?

tomtau:
or one more alternative is to change `ListenerFn` to `Box<dyn Fn(String, &Position<'_>) -> bool + Sync + Send + RefUnwindSafe>` ... but it'd require using a different channel implementation, e.g. the crossbeam one: https://docs.rs/crossbeam-channel/0.5.6/crossbeam_channel/struct.Sender.html (the standard library one is only Send, i.e. it isn't Sync and RefUnwindSafe)

tomtau:
> or one more alternative is to change `ListenerFn` to `Box<dyn Fn(String, &Position<'_>) -> bool + Sync + Send + RefUnwindSafe>` ... but it'd require using a different channel implementation, e.g. the crossbeam one: https://docs.rs/crossbeam-channel/0.5.6/crossbeam_channel/struct.Sender.html (the standard library one is only Send, i.e. it isn't Sync and RefUnwindSafe)

maybe this is the easiest / most straightforward: https://github.com/pest-parser/pest/pull/736/commits/7f8e6053c2ca2450033c652e539ecba3042de4c5

NoahTheDuke:
> fn-change-arity

Bummer that I introduced a SemVer violation. I agree that those functions _probably_ shouldn't be public, but they are and we have no plans on hiding them. I think you're right that re-adding the argument and just calling it with the static variable would be a good solution.

That being said, it's been a year and no one has complained. Is there any way for us to survey public usage of Pest and see if anyone is even using the functions? Could help us know 1) our options and 2) the best way to move forward.

> auto_trait_impl_removed

I don't know much about this stuff (Send and Sync specifically) but your proposed solution looks like it works well.

tomtau:
> That being said, it's been a year and no one has complained. Is there any way for us to survey public usage of Pest and see if anyone is even using the functions? Could help us know 1) our options and 2) the best way to move forward.

@NoahTheDuke if we take code available on GH as an indicator, then those functions weren't used by anyone except from within pest itself (code search only finds pest forks with those functions):

https://github.com/search?q=validate_pest_keywords+language%3ARust&type=code&l=Rust
https://github.com/search?q=validate_rust_keywords+language%3ARust&type=code&l=Rust
https://github.com/search?q=validate_undefined+language%3ARust&type=code&l=Rust

NoahTheDuke (approved):

NoahTheDuke:
That's great news. Then we're probably good with leaving them as is and maybe noting it in a changelog some place.

PR #737, opened 2022-11-22
perf: optimize `skip_until` using memchr and aho-corasick

tomtau:
so far, SIMD support is only available for x86_64 and wasm

### DIFF
--- a/pest/Cargo.toml
+++ b/pest/Cargo.toml
@@ -16,7 +16,7 @@ rust-version = "1.56"
 [features]
 default = ["std"]
 # Implements `std::error::Error` for the `Error` type
-std = ["ucd-trie/std", "thiserror"]
+std = ["aho-corasick/std", "memchr/std", "ucd-trie/std", "thiserror"]
 # Enables the `to_json` function for `Pair` and `Pairs`
 pretty-print = ["serde", "serde_json"]
 # Enable const fn constructor for `PrecClimber`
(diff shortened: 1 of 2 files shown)


tomtau:
oh, aho-corasick is std-only: https://github.com/BurntSushi/aho-corasick/blob/7e231db4b4ac192ebc674078f2b03cd37b9ed5b9/src/lib.rs#LL187

Probably not worth it for now.

Just for the context: according to my understanding, `skip_until` is fairly common (e.g. from negative predicates in grammars) and good to optimize for. SIMD could give some gains here.  In the json benchmark, I got a >10% speedup on x86_64:
```
json parser             time:   [21.152 µs 22.303 µs 23.872 µs]                         
                        change: [-17.602% -11.654% -5.0965%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  18 (18.00%) high severe
```

but I found it was actually slower on Arm64 (Apple M1), because memchr / aho-corasick don't support Arm64 SIMD (NEON or SVE) yet... @pest-parser/triage I'll close this PR for now, but if it seems good to have this (e.g. under a feature flag), I can re-open it.

NoahTheDuke:
This definitely feels worth watching for in the future. If they can get Arm64 SIMD support (even just to reach speed parity, not gains), I think this is worth pursuing.

tomtau:
@NoahTheDuke do you think it's worth putting it in the codebase under a feature flag? Right now, there's https://github.com/pest-parser/pest/blob/master/pest/Cargo.toml#LL26 , so perhaps that feature could be renamed as "simd" to cover other SIMD-optimizations (not just `Position::line_col`)?

dragostis:
SIMD is quite easy to add with the [stabilized Arm intrinsics](https://doc.rust-lang.org/core/arch/aarch64/index.html).

tomtau:
@dragostis Arm has this official source for optimised memchr: https://github.com/ARM-software/optimized-routines/blob/master/string/aarch64/memchr-sve.S
There's this WIP PR for the Rust crate (only memchr, not memmem or aho-corasick): https://github.com/BurntSushi/memchr/pull/114
And there's that portable SIMD effort: https://doc.rust-lang.org/std/simd/index.html

In any case, AFAIK there are two issues at the moment:
1) aho-corasick requires std;
2) a lot of the SIMD features require either MSRV higher than 1.59 (we currently support 1.56+) or nightly.

Hence, I suggested putting it under a feature flag (instead of being the default) for now.

PR #741, opened 2022-11-25
ci: add pest_debugger to semver checks

tomtau:


tomtau:
> Error: rustdoc format v23 for file /tmp/baseline-pest_debugger.json is not supported

hm

tomtau:
> > Error: rustdoc format v23 for file /tmp/baseline-pest_debugger.json is not supported
> 
> hm

it was due to semver checks incompatibility with the latest nightly...it was fixed by going to a nightly version from a week ago

PR #743, opened 2022-11-29
doc: fix tables of (non-)terminals

birkenfeld:
For some reason, rustdoc does not recognize the table when it is indented.  (This behavior seems to have changed between the 2.0 and 2.1 builds on docs.rs, while the markup stayed unchanged.)

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -152,12 +152,12 @@
 //!
 //! 1. Terminals
 //!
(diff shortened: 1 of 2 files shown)


birkenfeld:
See https://docs.rs/pest_derive/latest/pest_derive/index.html#expressions for the current state.

NoahTheDuke (approved):

tomtau (approved):

PR #744, opened 2022-11-29
doc: explain that matching is greedy without backtracking

birkenfeld:
Example borrowed and adapted from #742


tomtau (approved):
nice, thanks for improving the docs!

PR #707, opened 2022-09-10
fix: speed up `Position::line_col` for large inputs using SIMD

tomtau:
note that this may have extra overhead for small inputs and requires two extra dependencies, hence the faster `line_col` was put under the optional `fast-line-col` feature flag.

closes #560

Issue #560, opened 2021-10-25
line_col method N times slow in large file.


huacnlee:
When use pest in a large content file (about 700 KB JSON file), the `Pair<R>` to call `line_col` looks like has performance issue **O(N)**.

Source code ref:
https://github.com/huacnlee/autocorrect/blob/v1.4.4/src/lib/src/code.rs#L56

The large file case:

https://github.com/microsoft/vscode-loc/blob/release/1.62.0/i18n/vscode-language-pack-zh-hans/translations/main.i18n.json

```rust
// 200ns
let span = item.as_span();
// 48ns
let start_pos = span.start_pos();
let start = std::time::Instant::now();
let (part_line, part_col) = start_pos.line_col();
println!(
    "line_col: {}ns, {}ms",
    start.elapsed().as_nanos(),
    start.elapsed().as_millis()
);
```

The benchmark result:

```
line_col: 1545291ns, 1ms
line_col: 1533083ns, 1ms
line_col: 1581541ns, 1ms
line_col: 1537250ns, 1ms
line_col: 1543208ns, 1ms
....
line_col: 2226208ns, 2ms
line_col: 2244875ns, 2ms
line_col: 2229541ns, 2ms
line_col: 2299000ns, 2ms
line_col: 2235958ns, 2ms
....
line_col: 3397791ns, 3ms
line_col: 3398291ns, 3ms
line_col: 3402250ns, 3ms
line_col: 3433666ns, 3ms
line_col: 3409250ns, 3ms
....
line_col: 4089000ns, 4ms
line_col: 4078750ns, 4ms
line_col: 4084083ns, 4ms
line_col: 4128250ns, 4ms
line_col: 4095708ns, 4ms
....
line_col: 5021375ns, 5ms
line_col: 5118083ns, 5ms
line_col: 5028333ns, 5ms
line_col: 5065166ns, 5ms
line_col: 5030750ns, 5ms
...
line_col: 6076458ns, 6ms
line_col: 6124375ns, 6ms
line_col: 6077041ns, 6ms
line_col: 6078041ns, 6ms
line_col: 6081791ns, 6ms
line_col: 6082208ns, 6ms
line_col: 6119625ns, 6ms
...
line_col: 7365125ns, 7ms
line_col: 7368541ns, 7ms
line_col: 7374416ns, 7ms
line_col: 7373250ns, 7ms
line_col: 7366041ns, 7ms
line_col: 7372750ns, 7ms
...
[...]

PR #754, opened 2022-12-27
Improve line, col calculate performance by use move cursor on Pairs Iterator.

huacnlee:
ref: #707, #560

This implementation is based by I use in AutoCorrect: https://github.com/huacnlee/autocorrect/blob/v2.5.5/autocorrect/src/result.rs#L267

## Add `line_col` method for get `(line, col)` in pair

```rust
let mut pairs = JSONParser::parse(Rule::a, "....").unwrap();
let pair = pairs.next().unwrap();
let (line, col) = pair.line_col()
```

In this method, line, col has managed by move_cursor on Pairs iterator `next` method. For a large input string this method will be faster, because it does not need to re-calculate all of the input string. The line, col value has been cached in the pair and accumulate by `next` method.

And I keep the old `line_col` method in `Position` struct, because I found that has been used in other without `Pairs` iterator. This method still works, but it will be slower than the new method.

## Benchmark

```
pair.line_col                              time:   [2.9937 µs 2.9975 µs 3.0018 µs]
position.line_col                          time:   [212.59 µs 213.38 µs 214.29 µs]
position.line_col (with fast-line-col)     time:   [18.241 µs 18.382 µs 18.655 µs]
```

> 🎊 9x fast than position.line_col  with `fast-line-col`
> 🌈 100x fast than old position.line_col  

## Limitation

This implementation only work for forward iteration for `Pairs`. So iterate backward will fallback to `Position::line_col` method.

### DIFF
--- a/pest/src/iterators/pairs.rs
+++ b/pest/src/iterators/pairs.rs
@@ -25,6 +25,82 @@ use super::queueable_token::QueueableToken;
 use super::tokens::{self, Tokens};
 use crate::RuleType;
 
+#[derive(Clone)]
+pub struct Cursor {
+    pub line: usize,
+    pub col: usize,
+}
+
+impl Default for Cursor {
+    fn default() -> Cursor {
+        Cursor { line: 1, col: 1 }
+    }
+}
+
+impl Cursor {
+    fn get(&self) -> (usize, usize) {
+        (self.line, self.col)
+    }
+}
+
+pub trait CursorPairs {
+    fn cursor(&self) -> Cursor;
+    fn cursor_mut(&mut self) -> &mut Cursor;
+
+    /// Move the (line, col) with string part
+    fn move_cursor(&mut self, part: &str) -> (usize, usize) {
+        let (l, c, has_new_line) = self.line_col(part);
+
+        let (prev_line, prev_col) = self.cursor().get();
+
+        self.cursor_mut().line += l;
+        if has_new_line {
+            self.cursor_mut().col = c;
+        } else {
+            self.cursor_mut().col += c;
+        }
+        (prev_line, prev_col)
+    }
+
+    /// Calculate line and col number of a string part
+    /// Fork from Pest for just count the part.
+    ///
+    /// https://github.com/pest-parser/pest/blob/85b18aae23cc7b266c0b5252f9f74b7ab0000795/pest/src/position.rs#L135
+    fn line_col(&self, part: &str) -> (usize, usize, bool) {
+        let mut chars = part.chars().peekable();
+
+        let mut line_col = (0, 0);
+        let mut has_new_line = false;
+
+        loop {
+            match chars.next() {
+                Some('\r') => {
+                    if let Some(&'\n') = chars.peek() {
+                        chars.next();
+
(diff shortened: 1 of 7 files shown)


pest/src/iterators/pairs.rs
```diff
@@ -25,6 +25,82 @@ use super::queueable_token::QueueableToken;
...
+        let mut line_col = (0, 0);
+        let mut has_new_line = false;
+
+        loop {
+            match chars.next() {
+                Some('\r') => {
+                    if let Some(&'\n') = chars.peek() {
+                        chars.next();
+
+                        line_col = (line_col.0 + 1, 1);
```
NoahTheDuke, line 81: Are these supposed to be `(line_col.0 + 1, line_col.1)`?

huacnlee: This method has copied from AutoCorrect it has passed a lot of test case and use in production:

- https://github.com/huacnlee/autocorrect/blob/v2.5.5/autocorrect/src/result.rs#L287
- https://github.com/pest-parser/pest/blob/85b18aae23cc7b266c0b5252f9f74b7ab0000795/pest/src/position.rs#L135

And this line means:

```
If next char is `\n`, returns line + 1 and col reset to 1
```

NoahTheDuke: Ah yes, I see how that works now. Thanks.

pest/src/iterators/pairs.rs
```diff
@@ -25,6 +25,82 @@ use super::queueable_token::QueueableToken;
...
+            self.cursor_mut().col += c;
+        }
+        (prev_line, prev_col)
+    }
+
+    /// Calculate line and col number of a string part
+    /// Fork from Pest for just count the part.
+    ///
+    /// https://github.com/pest-parser/pest/blob/85b18aae23cc7b266c0b5252f9f74b7ab0000795/pest/src/position.rs#L135
+    fn line_col(&self, part: &str) -> (usize, usize, bool) {
```
tomtau, line 69: there's a bit of a duplicate code with position... could the common code be extracted to a shared `pub(crate)` or `pub(in ...)` function?

pest/src/position.rs
```diff
@@ -116,6 +116,10 @@ impl<'i> Position<'i> {
 
     /// Returns the line and column number of this `Position`.
     ///
+    /// This is an O(n) operation, where n is the number of lines in the input.
+    /// You better use `pair.line_col()` instead.
```
tomtau, line 120: nit: this can preferably link the corresponding rustdoc, perhaps: 
```
[`pair.line_col()`](struct.Pair.html#method.line_col)
```

or see https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

tomtau (approved):
one alternative for testing may be a different test parser definition, plus check that position and pair results match ?

btw is there a need for `fast-line-col` with this, or should that feature be removed? @huacnlee @NoahTheDuke

huacnlee:
We still need `fast-line-col`, because my implementation only work for call from Pairs.

There have many case use `Position` in directly.

huacnlee:
Hi @tomtau

I have done all, please review again.

PR #758, opened 2023-01-04
Add to support load multiple grammars for derive generator. 

### LINKED ISSUE #197: Sharing rules between grammars?

huacnlee:
Resolve #197 #333

### Allows load multiple grammar files for derive generator.

Example:

```rust
#[derive(Parser)]
#[grammar = "base.pest"]
#[grammar = "json.pest"]
pub struct JSONParser;
```

By this changes, we can split the sharing rules in to a single file, and then load it for other grammar files.

In my AutoCorrect case, this will useful, because there have so many duplicate rules (e.g. `space`, `comment`, `newline`) defined in each grammars.

https://github.com/huacnlee/autocorrect/tree/v2.5.5/autocorrect/grammar

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -41,39 +41,49 @@ use pest_meta::{optimizer, unwrap_or_report, validator};
 /// "include_str" statement (done in pest_derive, but turned off in the local bootstrap).
 pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
     let ast: DeriveInput = syn::parse2(input).unwrap();
-    let (name, generics, content) = parse_derive(ast);
+    let (name, generics, contents) = parse_derive(ast);
 
-    let (data, path) = match content {
-        GrammarSource::File(ref path) => {
-            let root = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
+    let mut data = String::new();
+    let mut path = None;
 
-            // Check whether we can find a file at the path relative to the CARGO_MANIFEST_DIR
-            // first.
-            //
-            // If we cannot find the expected file over there, fallback to the
-            // `CARGO_MANIFEST_DIR/src`, which is the old default and kept for convenience
-            // reasons.
-            // TODO: This could be refactored once `std::path::absolute()` get's stabilized.
-            // https://doc.rust-lang.org/std/path/fn.absolute.html
-            let path = if Path::new(&root).join(path).exists() {
-                Path::new(&root).join(path)
-            } else {
-                Path::new(&root).join("src/").join(path)
-            };
+    for content in contents {
+        let (_data, _path) = match content {
+            GrammarSource::File(ref path) => {
+                let root = env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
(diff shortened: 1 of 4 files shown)


tomtau (commented):
Great work, thanks! I think that multi-derive generator approach (i.e. this commit https://github.com/pest-parser/pest/pull/758/commits/157f1590379ce132259901b88e7f804e9d019db8 ) is simple and can address the basic use cases, so I'd be happy to merge that right away -- perhaps could you split it into two PRs?

For the meta grammar extension (i.e. this commit https://github.com/pest-parser/pest/pull/758/commits/30c15cf3824a97cc65228895f426fe8cbb75ee4b ), I think it's fine, but would want more feedback from other folks. If my understanding of this `use <filename>` feature is correct, it'll simply concat all definitions into one grammar (i.e. it's unlike that previous work by @bobbbay https://github.com/pest-parser/pest/pull/660 that tried to add module-level scoping). Is that correct? If so, what is the advantage of writing `use ...` inside a grammar over writing multiple derive annotations?

If this meta grammar extension is to be added, it's maybe worth adding more complex tests (e.g. use statements across multiple files A.pest -> B.pest -> C.pest, or multiple use statements in one file) and documenting this feature in the grammar reference in rustdocs/readme/book (especially its limitations or what one needs to watch for... I assume for this concat approach, there can't be a cycle / mutual dependency, one needs to make sure each grammar is included only once...?).

generator/src/lib.rs
```diff
@@ -115,13 +177,16 @@ fn parse_derive(ast: DeriveInput) -> (Ident, Generics, GrammarSource) {
         })
         .collect();
 
-    let argument = match grammar.len() {
-        0 => panic!("a grammar file needs to be provided with the #[grammar = \"PATH\"] or #[grammar_inline = \"GRAMMAR CONTENTS\"] attribute"),
-        1 => get_attribute(grammar[0]),
-        _ => panic!("only 1 grammar file can be provided"),
-    };
+    let mut grammar_sources = vec![];
```
NoahTheDuke, line 180: Maybe `Vec::with_capacity(grammar.len())`? I don't know how much this saves, especially if most folks will only be using one to two grammars, but seems worth it to me.

generator/src/lib.rs
```diff
@@ -115,13 +177,16 @@ fn parse_derive(ast: DeriveInput) -> (Ident, Generics, GrammarSource) {
...
+    for attr in grammar {
+        grammar_sources.push(get_attribute(attr))
+    }
+
+    if grammar_sources.is_empty() {
```
NoahTheDuke, line 185: Should this be moved above? "If grammar is empty, panic. Otherwise, get attrs."

huacnlee:
I have splitted https://github.com/pest-parser/pest/pull/759

tomtau (approved):
lgtm, the meta grammar extension can comments can continue on https://github.com/pest-parser/pest/pull/759

PR #759, opened 2023-01-04
Add to support `include!` for load partial for Pest grammars.

### LINKED ISSUE #197: Sharing rules between grammars?

huacnlee:
Resolve #197 #333

## Example

base.pest

```rs
WHITESPACE = _{ " " | "\t" | "\r" | "\n" }
```

json.pest

```rs
include!("base.pest")

json = { ... }
```

Will same as:

```rs
WHITESPACE = _{ " " | "\t" | "\r" | "\n" }

json = { ... }
```


huacnlee:
Or how about use:

A. Import 

```pest
import("base.pest")
```

B. Require same as Node.js, Ruby

```pest
require("base.pest")
```

C. Rust `include!` macro.

```pest
include!("base.pest")
```

huacnlee:
@CAD97 I have updated syntax to `include!("base.pest")` in last commit.

tomtau (changes requested):
I merged that original split PR, so this can be rebased on top of the latest master

huacnlee:
@tomtau I have updated

tomtau (commented):
It seems all right. I don't have any special preference for the syntax: include may be fine, just that Rust-like syntax form may perhaps be confusing to lead people into thinking they can embed Rust code.

One thing that's missing is docs: when should one use this include statement versus using multiple derive expressions?

huacnlee:
https://github.com/pest-parser/book/pull/29

tomtau:
@huacnlee from those docs, if I understand it correctly, `derive` and  this `include` meta extension achieve exactly the same outcome, i.e. there is no practical difference between them, or is there (e.g. if one of them doesn't work and the other one works in some workspace or old rustc/cargo settings?)?

huacnlee:
By my mind, `include!` just another way to let us to load partial grammars. 

Because of this is same as the Rust `include!` macro, so it easier to understand.

In most sense, we just need focus on the `.pest` files (e.g. Share the `.pest` files to other project, we just need copy the files).

generator/src/lib.rs
```diff
@@ -36,6 +36,42 @@ mod generator;
...
+    };
+
+    path
+}
+
+/// Get path relative to `path` dir, or relative to root path
+fn partial_path(path: Option<&PathBuf>, filename: &str) -> PathBuf {
+    let root = match path {
+        Some(path) => path.parent().unwrap().to_path_buf(),
+        None => join_path("./"),
```
CAD97, line 63: Since `join_path` looks for the presence of a file to choose between the manifest dir or source dir, this is going to always use the `src` dir.

We should deliberately pick one as the "location" of attribute-inline grammars, or perhaps a better choice, just disallow `include!`.

Threading up an error from here might be a bigger change, unfortunately.

tomtau (commented):
I'm putting this one on hold for now, because it's not obvious what specific problem it solves (given multiple grammar file loading is already supported via derive generators #758) and whether it's worth extending the meta grammar with a feature that duplicates what is already possible using derive generators.

If someone from @pest-parser/maintainers @pest-parser/triage finds the extension useful in its current form and champions this PR, I think it's ok to go forward with it. Otherwise, it will be better to extend the meta grammar with a feature that is not achieved using derive generators, such as module-level scoping as sketched out in #660 .

generator/src/lib.rs
```diff
@@ -36,6 +36,42 @@ mod generator;
...
+    };
+
+    // Add .pest suffix if not exist
+    let mut filename = filename.to_string();
+    if !filename.to_lowercase().ends_with(".pest") {
+        filename.push_str(".pest");
+    }
```
CAD97, line 70: It's possible to create file extension less files, so adding an extension on automatically is probably not a great choice. If we do want to add an extension, it should be done via conversion to and then modification via `PathBuf` rather than string manipulation.

generator/src/lib.rs
```diff
@@ -85,13 +103,43 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
...
+        }
+    }
+
+    if has_use {
+        // Re-parse the data again, after replacing the `use` statement
+        pairs = match parser::parse(Rule::grammar_rules, &data) {
+            Ok(pairs) => pairs,
+            Err(error) => panic!("error parsing \n{}", error.renamed_rules(rename_meta_rule)),
+        };
+    }
```
CAD97, line 138: Doing it this way rather than in a loop means that only a single layer of `include!` will be resolved; a `include!` of a file with an `include!` won't work.

At a minimum we need a test for this case. Ideally we'd make it work, but so long as it results in an error rather than silently causing corruption it's not too bad.

tomtau:
@huacnlee I think one potential use case may be in contexts where one *cannot* use Rust derive generators. For example:
- the CLI debugger loads the grammar files directly;
- the online fiddle editor doesn't even have a concept of a filesystem: https://pest.rs/#editor (but I can imagine it could perhaps be extended to have "multi-tab" editing where different grammar files are pasted into different text boxes)

So in the current form, it will also need to add support for `include!` in `pest_vm` and have some sample + test in `pest_debugger`

PR #765, opened 2023-01-09
Add to support `///` and `//!` syntax for add doc comment for rules.

### LINKED ISSUE #748: Request: Insert doc comments from `.pest` file to the `Rule` struct.
It would be incredibly helpful if the `Rule` struct and its variants have documentation to assist user.

## Example

```pest
//! Documentation of the Rule struct itself.

/// Documentation of the `my_rule` variant.
my_rule - { ... }
```

should generate:

```rust
/// Documentation of the Rule struct itself.
enum Rule {
    /// Documentation of the `my_rule` variant.
    my_rule,
}
```

huacnlee:
Resolve #748

For example:

```rust
//! A parser for JSON file.

/// Matches object, e.g.: `{ "foo": "bar" }`
object = { "{" ~ pair ~ ("," ~ pair)* ~ "}" | "{" ~ "}" }
```

should generate:

```rust
/// A parser for JSON file.
enum Rule {
    /// Matches object, e.g.: `{ "foo": "bar" }`
    object,
}
```

SCR-20230109-p3j

### DIFF
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -7,7 +7,6 @@
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
 
-use std::collections::HashMap;
 use std::path::PathBuf;
 
 use proc_macro2::TokenStream;
@@ -18,34 +17,7 @@ use pest::unicode::unicode_property_names;
 use pest_meta::ast::*;
 use pest_meta::optimizer::*;
 
-#[derive(Debug)]
-pub(crate) struct DocComment {
-    /// Multi-line grammar doc, (joined with `\n`)
-    ///
-    /// e.g.
-    ///
-    /// ```ignore
-    /// "grammar doc 1\ngrammar doc 2"
-    /// ```
-    grammar_doc: String,
-    /// HashMap rule name and doc comments (joined with `\n`)
-    ///
-    /// e.g.
-    ///
-    /// ```ignore
-    /// { "foo": "line doc 1\nline doc 2", "bar": "line doc 3" }
-    /// ```
-    line_docs: HashMap<String, String>,
-}
-
-impl DocComment {
-    pub fn new(grammar_doc: String, line_docs: HashMap<String, String>) -> Self {
-        Self {
-            grammar_doc,
-            line_docs,
-        }
-    }
-}
+use crate::docs::DocComment;
 
 pub(crate) fn generate(
     name: Ident,
@@ -705,9 +677,10 @@ fn option_type() -> TokenStream {
 
 #[cfg(test)]
 mod tests {
-    use proc_macro2::Span;
-
     use super::*;
+
+    use proc_macro2::Span;
+    use std::collections::HashMap;
 
     #[test]
     fn rule_enum_simple() {
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -23,6 +23,7 @@ pub fn generate(
     path: Option<PathBuf>,
     rules: Vec<OptimizedRule>,
     defaults: Vec<&str>,
+    grammar_docs: Vec<&str>,
     include_grammar: bool,
 ) -> TokenStream {
     let uses_eoi = defaults.iter().any(|name| *name == "EOI");
@@ -36,7 +37,7 @@ pub fn generate(
     } else {
         quote!()
     };
-    let rule_enum = generate_enum(&rules, uses_eoi);
+    let rule_enum = generate_enum(&rules, grammar_docs, uses_eoi);
     let patterns = generate_patterns(&rules, uses_eoi);
     let skip = generate_skip(&rules);
 
@@ -181,10 +182,26 @@ fn generate_include(name: &Ident, path: &str) -> TokenStream {
     }
 }
 
-fn generate_enum(rules: &[OptimizedRule], uses_eoi: bool) -> TokenStream {
-    let rules = rules.iter().map(|rule| format_ident!("r#{}", rule.name));
+fn generate_enum(rules: &[OptimizedRule], grammar_docs: Vec<&str>, uses_eoi: bool) -> TokenStream {
+    let rules = rules.iter().map(|rule| {
+        let rule_name = format_ident!("r#{}", rule.name);
+        if rule.comments.is_empty() {
+            quote! {
+                #rule_name
+            }
+        } else {
+            let comments = rule.comments.join("\n");
+            quote! {
+                #[doc = #comments]
+                #rule_name
+            }
+        }
+    });
+
+    let grammar_docs = grammar_docs.join("\n");
     if uses_eoi {
         quote! {
+            #[doc = #grammar_docs]
             #[allow(dead_code, non_camel_case_types, clippy::upper_case_acronyms)]
             #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
             pub enum Rule {
@@ -194,6 +211,7 @@ fn generate_enum(rules: &[OptimizedRule], uses_eoi: bool) -> TokenStream {
         }
     } else {
         quote! {
+            #[doc = #grammar_docs]
             #[allow(dead_code, non_camel_case_types, clippy::upper_case_acronyms)]
             #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
             pub enum Rule {
@@ -208,6 +226,7 @@ fn generate_patterns(rules: &[OptimizedRule], uses_eoi: bool) -> TokenStream {
         .iter()
         .map(|rule| {
             let rule = format_ident!("r#{}", rule.name);
+
             quote! {
                 Rule::#rule => rules::#rule(state)
             }
@@ -667,14 +686,17 @@ mod tests {
             name: "f".to_owned(),
             ty: RuleType::Normal,
             expr: OptimizedExpr::Ident("g".to_owned()),
+            comments: vec!["This is rule comment".to_owned()],
         }];
 
         assert_eq!(
-            generate_enum(&rules, false).to_string(),
+            generate_enum(&rules, vec!["Rule doc", "hello"], false).to_string(),
             quote! {
+                #[doc = "Rule doc\nhello"]
                 #[allow(dead_code, non_camel_case_types, clippy::upper_case_acronyms)]
                 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
                 pub enum Rule {
+                    #[doc = "This is rule comment"]
                     r#f
                 }
             }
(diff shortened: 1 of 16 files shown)


meta/src/parser.rs
```diff
@@ -1093,13 +1132,43 @@ mod tests {
         };
     }
 
+    #[test]
+    fn grammar_doc_and_line_doc() {
```
tomtau, line 1136: not sure if there could also be an assertion to check the comment was asserted correctly?

tomtau (commented):
great work! sadly, it looks like it's a semver breaking change:
```
Checking pest_meta
    Checking <unknown> v2.5.0 -> v2.5.2 (patch change)
   Completed [   0.141s] 33 checks; 31 passed, 2 failed, 0 unnecessary

--- failure constructible_struct_adds_field: externally-constructible struct adds field ---

Description:
A pub struct constructible with a struct literal has a new pub field. Existing struct literals must be updated to include the new field.
        ref: https://doc.rust-lang.org/reference/expressions/struct-expr.html
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.15.2/src/queries/constructible_struct_adds_field.ron

Failed in:
  field Rule.comments in meta/src/ast.rs:22
  field ParserRule.comments in meta/src/parser.rs:55
  field OptimizedRule.comments in meta/src/optimizer/mod.rs:105

--- failure enum_variant_added: enum variant added on exhaustive enum ---

Description:
A publicly-visible enum without #[non_exhaustive] has a new variant.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.15.2/src/queries/enum_variant_added.ron

Failed in:
  variant Expr:LineDoc in meta/src/ast.rs:93
  variant ParserExpr:LineDoc in meta/src/parser.rs:171
  variant OptimizedExpr:LineDoc in meta/src/optimizer/mod.rs:140
       Final [   0.161s] semver requires new major version: 2 major and 0 minor checks failed
```

not sure if it's possible to do in a non-breaking way (e.g. under a feature flag)... maybe mark that enum `#[non_exhaustive]` and make that `comments` field private (instead of public)?

meta/src/parser.rs
huacnlee: Sorry, I don't know what your means of this. Would you please explain more with informations?

meta/src/parser.rs
```diff
@@ -1347,7 +1419,11 @@ mod tests {
                             ))))
                         ))
                     ))))))
-                )
+                ),
+                comments: vec![
+                    "This is line comment".to_string(),
+                    "This is rule".to_string(),
+                ],
```
huacnlee, line 1426: > so having an assertion that [" hello", " world"] was correctly extracted out (just like in the ast test that asserts that the line/rule comments were extracted)

@tomtau Do you means this test?

In this test case, it covered for extract `/// This is rule` line comments into `["This is line comment", "This is rule"]`.

huacnlee:
Looks like we can't do it without the breaking changes.

We must add `comments` fields into `Rule`, `ParserRule` and `OptimizedRule` struct for storage the parsed comment strings. There no other way to store them.

-----

`Expr::LineDoc`, `ParserExpr::LineDoc` and `OptimizedExpr::LineDoc` I have removed, it not need.

generator/src/generator.rs
```diff
@@ -17,12 +17,28 @@ use pest::unicode::unicode_property_names;
...
+
+impl DocComment<'_> {
+    fn line_docs_for_rule<'a>(&self, rule_name: &str) -> Option<String> {
+        let idx = self.rules.iter().position(|r| r.name == rule_name)?;
+
+        self.line_docs.get(idx).map(|comments| comments.join("\n"))
+    }
+}
+
+pub(crate) fn generate(
```
huacnlee, line 35: This because of the`mod generator` is not public.

So I changed it to `pub(crate)`.

huacnlee:
@tomtau I have rewrite the implementation without make breaking changes.

meta/src/grammar.pest
```diff
@@ -7,11 +7,12 @@
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
 
-grammar_rules = _{ SOI ~ grammar_rule+ ~ EOI }
+grammar_rules = _{ SOI ~ grammar_doc* ~ (grammar_rule)+ ~ EOI }
 
 grammar_rule = {
```
tomtau, line 12: maybe it'll be good to "dogfood" this change in the meta grammar and use it to add doc comments, but that could be done in a follow up

tomtau (commented):
awesome work if it was made non-semver breaking! I'll have a look at in more detail hopefully soon

generator/src/generator.rs
```diff
@@ -17,12 +17,28 @@ use pest::unicode::unicode_property_names;
...
+    pub(crate) grammar_docs: Vec<&'a str>,
+    pub(crate) line_docs: Vec<Vec<&'a str>>,
+    pub(crate) rules: Vec<Rule>,
+}
+
+impl DocComment<'_> {
+    fn line_docs_for_rule(&self, rule_name: &str) -> Option<String> {
+        let idx = self.rules.iter().position(|r| r.name == rule_name)?;
+
+        self.line_docs.get(idx).map(|comments| comments.join("\n"))
```
tomtau, line 31: this may increase the compilation time for larger grammars a bit. So, my suggestion would be:
1. instead of constructing DocComment directly, make those inner fields private and have a `new` constructor function that takes pairs etc. and have that construction functionality (`consume_grammar_doc`, `consume_line_docs`) there instead of doing the construction externally (so that `DocComment` can maintain invariants about those inner fields)
2. not sure if there's any other purpose for `rules` other than looking up the index by the rule name? If no, maybe it'll be easier (and faster) just to have e.g. `line_docs: HashMap<String, Vec<&'a str>>`, so that line docs can be looked up directly?

generator/src/lib.rs
```diff
@@ -90,11 +93,65 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
...
+
+fn consume_line_docs(pairs: Pairs<'_, Rule>) -> Vec<Vec<&'_ str>> {
+    let mut docs = vec![];
+    let mut comments = vec![];
+
+    for pair in pairs {
+        if pair.as_rule() == Rule::grammar_rule {
+            if let Some(inner) = pair.into_inner().next() {
+                if inner.as_rule() == Rule::line_doc {
+                    comments.push(inner.as_str()[3..inner.as_str().len()].trim());
```
tomtau, line 139: the magic number `3` is perhaps obvious in the context of this PR, but just for any potential future changes, it may be worth adding a small comment to note it corresponds to either `///` or `//!` (or perhaps extract it to a constant with a meaningful name, e.g. `DOC_PREFIX_LEN`)

tomtau (changes requested):
the non-breaking version seems all right, but if possible, could `DocComment` hide its internal fields (for a better "encapsulation") ?

meta/src/parser.rs
tomtau: yes

meta/src/grammar.pest
tomtau: I could possibly do that in a follow up PR

generator/src/generator.rs
huacnlee: Done

tomtau:
https://github.com/pest-parser/pest/pull/770 was merged -- could you rebase on top of the latest master?

generator/src/lib.rs
```diff
@@ -90,11 +94,83 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
...
+fn consume_grammar_doc(pairs: Pairs<'_, Rule>) -> String {
+    let mut docs = vec![];
+    for pair in pairs {
+        if pair.as_rule() == Rule::grammar_doc {
+            let inner_doc = pair.into_inner().next().unwrap();
+            docs.push(inner_doc.as_str());
+        }
+    }
+
+    docs.join("\n")
```
tomtau, line 127: this may potentially be replaced with string building, i.e. have an empty string, and then `docs.push_str(inner_doc.as_str())` and `docs.push('\n')`

generator/src/generator.rs
```diff
@@ -17,12 +18,42 @@ use pest::unicode::unicode_property_names;
...
+    /// e.g.
+    ///
+    /// ```ignore
+    /// { "foo": "line doc 1\nline doc 2", "bar": "line doc 3" }
+    /// ```
+    line_docs: HashMap<String, String>,
+}
+
+impl DocComment {
+    pub fn new(grammar_doc: String, line_docs: HashMap<String, String>) -> Self {
```
tomtau, line 42: maybe `new` could take `pairs: Pairs<'_, Rule>` and call `consume_grammar_doc` / `consume_line_docs` (it could potentially fuse them in one iteration or still call them separately)?

generator/src/lib.rs
```diff
@@ -90,11 +94,83 @@ pub fn derive_parser(input: TokenStream, include_grammar: bool) -> TokenStream {
...
+///
+/// /// Line doc 2
+/// /// Line doc 3
+/// bar = {}
+/// ```
+///
+/// Will returns `{ "foo":  "This is line comment", "bar": "Line doc 2\n/// Line doc 3" }`
+fn consume_line_docs(pairs: Pairs<'_, Rule>) -> HashMap<String, String> {
+    let mut docs: HashMap<String, String> = HashMap::new();
+    let mut comments = vec![];
```
tomtau, line 146: ditto for the string construction

tomtau (approved):
lgtm, just a small nit about the string construction, but ok to merge

generator/src/lib.rs
huacnlee: That will have a extra `\n` in tail.

generator/src/generator.rs
huacnlee: Do you means move `consume_grammar_doc` and `consume_line_docs` implementation into DocComment?

huacnlee: How about this change?

https://github.com/pest-parser/pest/pull/765/commits/4d9d1af01251ec4dfb63e563656a024765ff23ea

generator/src/lib.rs
huacnlee: Or I use `push_str` and `pop` last char?

```
let out = String::new();

for pair in pairs {
  out.push_str(inner_doc.as_str())
  out.push('\n')
}

out.pop()
out
```

tomtau: I guess it'll need a check if non-empty, but it could work

generator/src/generator.rs
tomtau: yes, that works to logically group those functions in a module

generator/src/lib.rs
huacnlee: This part have updated.

PR #768, opened 2023-01-12
Fix incorrect `col` calculate for `pair.line_col` method, and add integration test for cover it.

huacnlee:
Hi @tomtau, when I use the new `pair.line_col` method in my project.

I found that `col` value is not correct:

image

https://github.com/pest-parser/pest/actions/runs/3899423161/jobs/6659084167

So I try to fix it, and add a complex integration test for cover it.


pest/src/iterators/pairs.rs
```diff
@@ -249,14 +249,7 @@ impl<'i, R: RuleType> Pairs<'i, R> {
...
         let part = &input[self.cursor.end..end];
-        let (l, c) = position::line_col(part, part.len());
-
-        // Because the `original_line_col` returns (line, col) is start from 1
-        let l = l - 1;
-        let mut c = c - 1;
-        if c < 1 {
-            c = 1
-        }
+        let (l, c) = position::original_line_col(part, part.len(), (0, 0));
```
tomtau, line 252: Can this still use the line_col() as before (so that it could benefit from fast-line-col when it's enabled)?

tomtau (approved):
Would it be possible to leverage the original line-col function (to give it an extra argument)?

pest/src/iterators/pairs.rs
huacnlee: If I do that, I will change the line_col to add a argument, is this method is private?

tomtau: line_col() is pub(crate), so it should be only exposed within the crate (in other words, changing it shouldn't be a semver-breaking change)

NoahTheDuke:
The code itself looks solid. I'm a little confused by why this change is needed. I think I just don't quite understand how the lines and cols are determined. Why do we need to start at 1,1 sometimes and 0,0 other times?

huacnlee:
@NoahTheDuke In `Pairs#move_cursor`, it calculates the `(line, col)` position by uses a part of string by keep accumulating. 

So it call `Position#line_col` just need get a number of offset of that str need to move (This is not like `line, col`. `line, col` is start from `1,1` buf offset need start with `0,0`).

PR #770, opened 2023-01-18
CI use Rust 1.62.0 for fix rustyline 10.1.0 `#[default]` compile error.

huacnlee:
https://github.com/pest-parser/pest/actions/runs/3946148398/jobs/6753649401

Ref:

- https://github.com/kkawakam/rustyline/compare/v10.0.0...v10.1.0
- https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1620-2022-06-30
- https://github.com/rust-lang/rust/pull/94457

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           kind: check
-          toolchain: 1.61.0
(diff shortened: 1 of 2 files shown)


tomtau (approved):
one alternative would be to fix rusty line to the previous version, but ok to bump to 1.62 (the MSVR is still checked for the pest crate)

PR #774, opened 2023-01-21
doc: added doc comments to the meta-grammar

tomtau:
it makes use of https://github.com/pest-parser/pest/pull/765

### DIFF
--- a/meta/src/grammar.pest
+++ b/meta/src/grammar.pest
@@ -6,23 +6,34 @@
 // license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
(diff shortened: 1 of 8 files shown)


NoahTheDuke (approved):

meta/src/grammar.pest
```diff
@@ -6,42 +6,64 @@
 // license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. All files in the project carrying such notice may not be copied,
 // modified, or distributed except according to those terms.
+//! Pest meta-grammar
 
+/// The top-level rule of the meta-grammar.
```
birkenfeld, line 11: This is a bit pedantic, but shouldn't "meta" be removed he and below? The rules are passing a normal grammar after all?

PR #779, opened 2023-01-27
Feature: Simple update checker for pest_debugger

Jamalam360:
This PR adds a simple synchronous update checker for pest_debugger. If an update is found, the new version is printed to the console:

`A new version of pest_debugger is available: v3.9.4` 

Else, 

`pest_debugger is up to date`

is printed.

The implementation discards any errors that occur during update checking since the user doesn't really need to know about them - update checking is non-essential. It's using the crates.io API, which is subject to their [policies](https://crates.io/policies). From what I understand, making one request every time someone starts pest_debugger will be fine.

Open for any feedback, my first contribution to an open source Rust project.

### DIFF
--- a/debugger/src/main.rs
+++ b/debugger/src/main.rs
@@ -311,6 +311,7 @@ fn main() -> rustyline::Result<()> {
                 "/",
                 env!("CARGO_PKG_VERSION")
             ))
+            .timeout(Some(Duration::from_secs(5)))
             .build()
             .ok();
 
--- a/debugger/src/main.rs
+++ b/debugger/src/main.rs
@@ -315,8 +315,7 @@ fn main() -> rustyline::Result<()> {
             .ok();
 
         if let Some(client) = client {
-            let opt = check_for_updates(client);
-            if let Some(new_version) = opt {
+            if let Some(new_version) = check_for_updates(client) {
                 println!(
                     "A new version of pest_debugger is available: v{}",
                     new_version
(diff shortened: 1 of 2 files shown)


Jamalam360:
Formattting fixed

debugger/src/main.rs
```diff
@@ -337,3 +366,23 @@ fn main() -> rustyline::Result<()> {
...
 }
+
+fn check_for_updates(client: Client) -> Option<String> {
+    let response = client
+        .get("https://crates.io/api/v1/crates/pest_debugger")
+        .send();
+
+    if let Ok(response) = response {
+        response.json::<serde_json::Value>().ok().and_then(|json| {
+            let version = json["crate"]["max_version"].as_str()?;
```
tomtau, line 377: I guess this may panic if the object didn't happen to contain those... it could be made more robust with checked value getting, but probably ok as it is

debugger/src/main.rs
```diff
@@ -337,3 +366,23 @@ fn main() -> rustyline::Result<()> {
...
+
+            if version != VERSION {
```
tomtau, line 379: this logic may perhaps not work in some edge cases -- e.g. if I build a local "newer" version or if the upstream version was yanked or something like that... maybe a more appropriate would be to compare semvers with `>`, but that would require defining and parsing version numbers which may be an overkill

tomtau (approved):
LGTM, thanks for the PR! I'm happy to merge it as it is; I added two small comments how it could be made more robust, but those may be an overkill for this small feature

PR #782, opened 2023-01-30
Update validator for ignore `Rule::line_doc` for fix #781 "already defined" error.

### LINKED ISSUE #781: Line doc "already defined" error
If we define line comment has multiple same contents, then will get compile error.

```rust
/// Matches dar
/// 
/// Match dar description
/// 
dar = { "da" }
```

This line doc has 2 empty lines, then will error:

```
thread 'tests::test_generate_doc' panicked at 'grammar error

  --> 19:1
   |
19 | ///
   | ^--^
   |
   = rule ///  already defined', /Users/jason/github/pest/meta/src/lib.rs:42:9
```

huacnlee:


tomtau (approved):
Is there a similar problem with grammar docs or only line docs?

huacnlee:
only in line doc, because they defined in `Rule::grammar_rule`.

PR #785, opened 2023-01-31
Add `LineIndex` instead of `Pairs::move_cursor` for improve `Pairs::next` and `Pair::line_col` performance.

### LINKED ISSUE #784: Pairs#move_cursor is slow down iter
Hi @tomtau,

I am very sorry about that. 😥

I have found `Pairs#iter` performance issue when we merged #754.

Because when we iter pairs, each iter next will called `Position#line_col` once to move cursor. That will slow down the iter . 

Especially that, in sometimes we just need get a little pair's line,col, not all of pairs. So to do this is not value of money.

By my test:

```rs
fn bench_pairs_iter(c: &mut Criterion) {
    let data = include_str!("data.json");

    fn iter_all_pairs(pairs: pest::iterators::Pairs<autocorrect::Rule>) {
        for pair in pairs {
            iter_all_pairs(pair.into_inner());
        }
    }

    c.bench_function("nested iter", |b| {
        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, &data).unwrap();

        b.iter(move || iter_all_pairs(pairs.clone()));
    });

    c.bench_function("flatten iter", |b| {
        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, &data).unwrap();

        b.iter(move || {
            for _pair in pairs.clone().flatten().into_iter() {
                // do nothing
            }
        });
    });
}
```

```
nested iter                             time:   [258.27 µs 260.05 µs 262.64 µs]
nested iter (fast-line-col)             time:   [14.943 µs 14.963 µs 14.993 µs]
flatten iter                            time:   [2.0367 µs 2.1104 µs 2.2144 µs]
```

> In a large files (or have a lot of pairs, For instance: enabled `WHITESPACE`, each spaces will a pair item), iterate all pairs may take a long time than before.

So we need decide to how to do, remove it or find some solution to fix that.

huacnlee:
Add `LineIndex` instead of `Pairs::move_cursor` for improve Pairs iterate and `Pair::line_col` performance.

Resolve #784
Ref: https://github.com/rust-lang/rust/blob/1.67.0/src/tools/rust-analyzer/crates/ide-db/src/line_index.rs

Benchmark result:

> (Updated on 2022.02.02 18:00) After I fix to for support utf-8, the new `line-index` implementation of line_col benchmark result is close to `fast-line-col`.
> But when In a last file, `line-index` will faster. 

Benchmark file: https://github.com/pest-parser/pest/blob/16de4fa29228ab07a8c49279de6c5f9814496ed6/grammars/benches/json.rs

```
// With 500 times iter
pair.line_col (with LineIndex)               time:   [22.426 µs 22.489 µs 22.556 µs]
position.line_col                            time:   [212.59 µs 213.38 µs 214.29 µs]
position.line_col (with fast-line-col)       time:   [18.241 µs 18.382 µs 18.655 µs]

// With 1000 times iter
pair.line_col (with LineIndex)               time:   [41.160 µs 41.969 µs 43.478 µs]
position.line_col                            time:   [1.7199 ms 1.7246 ms 1.7315 ms]
position.line_col (with fast-line-col)       time:   [48.498 µs 49.450 µs 50.877 µs]

// With 10K times iter
pair.line_col                                time:   [471.75 µs 475.03 µs 478.94 µs]
position.line_col (with fast-line-col)       time:   [8.1941 ms 8.3278 ms 8.5144 ms]
```

And I also tested the cost of `Pairs` iterate.

```
// Before changes
pairs nested iter                    time:   [2.0168 ms 2.0381 ms 2.0725 ms]
pairs flatten iter                   time:   [4.5973 µs 4.6132 µs 4.6307 µs]

// After added LineIndex
pairs nested iter (with LineIndex)   time:   [14.716 µs 14.822 µs 14.964 µs]
pairs flatten iter (with LineIndex)  time:   [5.4637 µs 5.6061 µs 5.7886 µs]
```

### DIFF
--- a/grammars/benches/json.rs
+++ b/grammars/benches/json.rs
@@ -21,7 +21,7 @@ fn bench_json_parse(c: &mut Criterion) {
     let data = include_str!("data.json");
 
     c.bench_function("json parser", |b| {
-        b.iter(|| JsonParser::parse(Rule::json, &data).unwrap())
+        b.iter(|| JsonParser::parse(Rule::json, data).unwrap())
     });
 }
 
@@ -55,16 +55,20 @@ inner_string = @{ ("\"" ~ (!(NEWLINE | "\"") ~ ANY)* ~ "\"") }
 // pair.line_col (with LineIndex)               time:   [41.160 µs 41.969 µs 43.478 µs]
 // position.line_col                            time:   [1.7199 ms 1.7246 ms 1.7315 ms]
 // position.line_col (with fast-line-col)       time:   [48.498 µs 49.450 µs 50.877 µs]
+//
+// With 10K times iter
+// pair.line_col                                time:   [471.75 µs 475.03 µs 478.94 µs]
+// position.line_col (with fast-line-col)       time:   [8.1941 ms 8.3278 ms 8.5144 ms]
 fn bench_line_col(c: &mut Criterion) {
     let data = include_str!("main.i18n.json");
-    let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, &data)
+    let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, data)
         .unwrap()
         .flatten();
 
     c.bench_function("pair.line_col", |b| {
         b.iter(|| {
             let mut pairs = pairs.clone();
-            for _ in 0..500 {
+            for _ in 0..1000 {
                 pairs.next().unwrap().line_col();
             }
         })
@@ -73,7 +77,7 @@ fn bench_line_col(c: &mut Criterion) {
     c.bench_function("position.line_col", |b| {
         b.iter(|| {
             let mut pairs = pairs.clone();
-            for _ in 0..500 {
+            for _ in 0..1000 {
                 pairs.next().unwrap().as_span().start_pos().line_col();
             }
         });
@@ -94,15 +98,15 @@ fn bench_pairs_iter(c: &mut Criterion) {
     }
 
     c.bench_function("pairs nested iter", |b| {
-        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, &data).unwrap();
+        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, data).unwrap();
         b.iter(move || iter_all_pairs(pairs.clone()));
     });
 
     c.bench_function("pairs flatten iter", |b| {
-        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, &data).unwrap();
+        let pairs = autocorrect::JsonParser::parse(autocorrect::Rule::item, data).unwrap();
 
         b.iter(move || {
-            for _pair in pairs.clone().flatten().into_iter() {
+            for _pair in pairs.clone().flatten() {
                 // do nothing
             }
         });
--- a/grammars/benches/json.rs
+++ b/grammars/benches/json.rs
@@ -85,6 +85,7 @@ fn bench_pairs_iter(c: &mut Criterion) {
 
     fn iter_all_pairs(pairs: pest::iterators::Pairs<autocorrect::Rule>) {
         for pair in pairs {
+            pair.line_col();
             iter_all_pairs(pair.into_inner());
         }
     }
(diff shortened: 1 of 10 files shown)


tomtau (commented):
I think the solution seems all right, but now, instead of the runtime penalty, there'll be a memory penalty. I recall @CAD97 's old comment that it was fine merging the cursor solution, because it had a fixed memory cost.
So perhaps to keep the runtime/memory behaviour of the original iterator, as discussed on the issue, maybe the middle ground will be to feature-guard it or create a separate wrapper iterator (`LocatablePairs` ?) that uses this `LineIndex`?

huacnlee:
This change just spend a Vec<usize> memory size that depends on number of lines. Because the LineIndex just storage the line offset, not str.

huacnlee:
And this still need more test. When I use this changes in my project, it looks like slower than before, not same as the benchmark result.

huacnlee:
I found it, `Pair#into_inner` not use the `line_index` ref.

After change, new benchmark result:

> And run the new changes in my project, benchmark come to fast.

```
pair.line_col           time:   [10.611 µs 10.651 µs 10.691 µs]
                        change: [+1.3490% +1.9534% +2.4931%] (p = 0.00 < 0.05)
                        Performance has regressed.

position.line_col       time:   [211.16 µs 211.41 µs 211.68 µs]
                        change: [-0.0250% +0.1523% +0.3160%] (p = 0.07 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

pairs nested iter       time:   [18.914 µs 19.036 µs 19.216 µs]
                        change: [-99.044% -99.041% -99.037%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

pairs flatten iter      time:   [17.058 µs 17.087 µs 17.119 µs]
                        change: [+0.2530% +0.4559% +0.6712%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

CAD97:
This is in fact a reduction in memory cost; the `Rc<LineIndex>` replaces line/col information wherever it's stored, allowing deriving of line/col in `O(col + line)`.

I was and still am hesitant to any changes which require increasing the size of `Pair`, since that impact is multiplied to be roughly linear to input size, but this change is reducing that size cost.

(The point of the cursor/view idea was to not have the line/col information on `Pair`, but apparently that didn't happen?)

huacnlee:
And I have also test this changes in my project: AutoCorrect

Here is the [benchmark file](https://github.com/huacnlee/autocorrect/blob/v2.5.8/autocorrect/benches/example.rs#L55) test a 2k json file ([Expected ~9ms / iter]((https://github.com/huacnlee/autocorrect/tree/v2.5.0#benchmark)))


## Before use `Position::line_col` in Pest 2.5.2 by enable `fast-line-col`

```
format_json_2k          time:   [19.625 ms 21.815 ms 24.671 ms]
                        change: [+109.28% +132.64% +161.98%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

## After use `Pair::line_col` with `line-index` feature.

```
format_json_2k          time:   [9.3551 ms 9.3770 ms 9.3995 ms]
                        change: [-1.3461% -0.8035% -0.2937%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

`9.3ms` is same fast like my self implementation of move_cursor in AutoCorrect project.

tomtau (commented):
Shall we remove fast-line-col, since it's not used in line index?

huacnlee:
> Shall we remove fast-line-col, since it's not used in line index?

Or we can just deprecated it, so we don't need make a breaking change.

Search code on GitHub, there is looks like not too much used it.

https://github.com/search?q=fast-line-col&type=code

huacnlee:
Test on https://github.com/kiibo382/prql/blob/main/prql-compiler/benches/bench.rs

```
// Before, Pest 2.5.0
variables-query         time:   [591.85 µs 593.78 µs 595.93 µs]

// After
variables-query         time:   [580.14 µs 581.95 µs 584.15 µs]
```

tomtau (approved):
will likely remove that feature, it won't be a "breaking change" in terms of semver, because it was an optional feature

PR #787, opened 2023-02-06
chore: added docs for semver changes in grammar + remove fast-col-line

tomtau:
using line_col from pairs is recommended: https://github.com/pest-parser/pest/pull/785#issuecomment-1413604569 as it's faster or similar performance after the line-index patch

### DIFF
--- a/debugger/Cargo.toml
+++ b/debugger/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "pest_debugger"
 description = "pest grammar debugger"
-version = "2.5.4"
+version = "2.5.5"
 edition = "2021"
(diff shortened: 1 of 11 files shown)


tomtau:
The main project I see that used `fast-col-line` is https://github.com/VirusTotal/yara-x/blob/ff32319be15114de42d79965f22e7cf787b0d7bd/yara-x-parser/Cargo.toml#L33

@VirusTotal @plusvic just a heads up that we plan to remove it, given the default line-index for line_col on Pairs: https://github.com/pest-parser/pest/pull/785#issuecomment-1413604569 or do you have any special use case for using line_col on Position?

NoahTheDuke (approved):
That warning is good.

plusvic:
Thanks for the heads-up!. I'm not using `fast-col-line` for anything special, I'm going to remove my dependency from that feature.

PR #790, opened 2023-02-09
Fix multiple grammars not always recompile issue #789

### LINKED ISSUE #789: Multiple grammars not always recompiled
**Describe the bug**
When using multiple grammars as described in #758:

```rust
use pest::Parser;
use pest_derive::Parser;

#[derive(Parser)]
#[grammar = "base.pest"]
#[grammar = "core.pest"]
struct SplitGrammarParser;
```

The parser is only recompiled with new grammars from both files when the last grammar is changed.  I.e., changes to only `base.pest` aren't captured in debug builds until something in `core.pest` changes.

**To Reproduce**

- Create a simple grammar split across two files
- Use derive as above
- Experiment with changes in either or both grammar files (e.g., with `cargo watch -x test`)

**Expected behavior**

A change to any grammar should cause the parser to recompile with updates from all grammar files.

huacnlee:
Closes #789


tomtau (approved):
LGTM, so the root cause was that only a single path was passed to that function at a time which prevented regenerating the other paths (only one of them was tracked)?

huacnlee:
I realize that maybe we don't need set all of pest file contents as a value, just only give a file updated at or a content md5. When the `.pest` files has changed, it's also can trigger change events.

NoahTheDuke:
Looks like a solid fix to me.

jayhale:
Thank you!

PR #793, opened 2023-02-10
Improve the `LineIndex` for as lazy calculate, it only calculates on the first `line_col` call.

huacnlee:


tomtau:
This seems to break semantic versioning that Pair, Pairs and FlatPairs no longer (auto-)derive UnwindSafe and RefUnwindSafe traits

pest/src/iterators/line_index.rs
```diff
@@ -7,38 +7,55 @@
...
 use alloc::vec::Vec;
+use core::cell::RefCell;
 
 #[derive(Clone)]
-pub struct LineIndex {
+pub struct LineIndex<'a> {
+    input: &'a str,
     /// Offset (bytes) the the beginning of each line, zero-based
-    line_offsets: Vec<usize>,
+    line_offsets: RefCell<Option<Vec<usize>>>,
```
tomtau, line 16: To preserve those UnwindSafe auto-traits, this would need to be Mutex or something like that instead of RefCell... but not sure if the performance improvement is still there in that case

huacnlee:
So, do you mean we can't use `RefCell` here?

tomtau:
Yes, I guess that's the cause of that semver breaking change that those auto traits were not derived

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.75s
Checking pest
    Checking <unknown> v2.5.0 -> v2.5.5 (patch change)
   Completed [   0.395s] 40 checks; 39 passed, 1 failed, 0 unnecessary
--- failure auto_trait_impl_removed: auto trait no longer implemented ---
Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.17.1/src/queries/auto_trait_impl_removed.ron
Failed in:
  type Pair is no longer RefUnwindSafe, in pest/src/iterators/pair.rs:39
  type Pair is no longer UnwindSafe, in pest/src/iterators/pair.rs:39
  type Pairs is no longer RefUnwindSafe, in pest/src/iterators/pairs.rs:35
  type Pairs is no longer UnwindSafe, in pest/src/iterators/pairs.rs:35
  type FlatPairs is no longer RefUnwindSafe, in pest/src/iterators/flat_pairs.rs:24
  type FlatPairs is no longer UnwindSafe, in pest/src/iterators/flat_pairs.rs:24
```

tomtau:
Alternatively, the RefCell version could be there under a feature flag, so that developers who don't need UnsafeUnwind on those structs could just enable that feature for a better performance

tomtau:
Or this instead of RefCell: https://docs.rs/ghost-cell/latest/ghost_cell/ghost_cell/struct.GhostCell.html ?

Issue #817, opened 2023-03-03
Parsing error (expected "EOI", "grammar_rule")


huacnlee:
In Move language case:

https://github.com/victoryang00/move-fmt/blob/4e6e47adf766177f48387b4216f0b161f0bce937/src/grammar.pest

```
////////////////////////////////////////////////////////////////////////////////////////////////////
/// structs
////////////////////////////////////////////////////////////////////////////////////////////////////
a = { "a" }
```

Will raise Parsing error:

```
Parsing error (expected "EOI", "grammar_rule")
```

PR #818, opened 2023-03-03
Fix grammar `////` parsing error.

huacnlee:
Solve #817

### DIFF
--- a/meta/src/grammar.pest
+++ b/meta/src/grammar.pest
@@ -171,6 +171,6 @@ space       = _{ " " | "\t" }
 /// A top-level comment.
 grammar_doc = ${ "//!" ~ space? ~ inner_doc }
 /// A rule comment.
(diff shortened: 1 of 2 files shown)


meta/src/grammar.pest
```diff
@@ -171,6 +171,6 @@ space       = _{ " " | "\t" }
 /// A top-level comment.
 grammar_doc = ${ "//!" ~ space? ~ inner_doc }
 /// A rule comment.
-line_doc    = ${ "///" ~ space? ~ !"/" ~ inner_doc }
```
huacnlee, line 174: Before this rule was following the Rust `line_doc` rule, wants to be let the `////` as a normal comment.

But I think, maybe we don't need to do that, just let it as a `line_doc` is also ok. So it will not come to a complex rule.

NoahTheDuke (approved):

PR #824, opened 2023-03-18
Fixup lifetimes of PrattParser

matanui159:
Currently the typings require the passed in callbacks to have the lifetime `'i`, and thus any values it reference must also have the same lifetime. If these callbacks return a value with lifetime `'i`, the borrow checker incorrectly assumes that those referenced values must also outlive the return value.

For example this code:
```rust
fn test_pratt<'a>(pairs: Pairs<'a, Rule>, message: &str) -> &'a str {
    PRATT_PARSER
        .map_primary(|pair| {
            // println!("{message}");
            pair.as_str()
        })
        .map_infix(|lhs, pair, rhs| {
            println!("{message}");
            pair.as_str()
        })
        .parse(pairs)
}
```
fails to compile with this error:
```
error[E0621]: explicit lifetime required in the type of `message`
   --> src/compiler/expression.rs:285:5
    |
284 |   fn test_pratt<'a>(pairs: Pairs<'a, Rule>, message: &str) -> &'a str {
    |                                                      ---- help: add explicit lifetime `'a` to the type of `message`: `&'a str`
285 | /     PRATT_PARSER
286 | |         .map_primary(|pair| {
287 | |             // println!("{message}");
288 | |             pair.as_str()
...   |
293 | |         })
294 | |         .parse(pairs)
    | |_____________________^ lifetime `'a` required
```

This is a small PR that adds a new lifetime to `PrattParserMap` so the callbacks can have a separate lifetime. This fixes the above issue.

### DIFF
--- a/pest/src/pratt_parser.rs
+++ b/pest/src/pratt_parser.rs
@@ -235,10 +235,10 @@ impl<R: RuleType> PrattParser<R> {
     }
 
     /// Maps primary expressions with a closure `primary`.
-    pub fn map_primary<'pratt, 'i, X, T>(
+    pub fn map_primary<'pratt, 'a, 'i, X, T>(
         &'pratt self,
         primary: X,
-    ) -> PrattParserMap<'pratt, 'i, R, X, T>
+    ) -> PrattParserMap<'pratt, 'a, 'i, R, X, T>
     where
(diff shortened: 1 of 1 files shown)


tomtau (approved):
Makes sense, thanks for the PR!

PR #826, opened 2023-03-26
Update syn to v2

CosmicHorrorDev:
Just fixes up the breakages from the `syn` update

It seems like the only thing that needed changing was from `meta` becoming a field of `Attribute` with nested meta parsing getting moved to a separate method entirely


tomtau (approved):

tomtau:
looks like it, thanks!

PR #832, opened 2023-04-08
Add basic support for `node_tag`, iterator, and meta grammar

tomtau:
revived https://github.com/pest-parser/pest/pull/552 by @oovm 

The original PR had an empty "tags" module in iterators, I assume it was meant for a custom "tagged pairs" iterator as sketched out in https://github.com/pest-parser/pest/discussions/550 with `pairs.find_first_mark` and `pairs.find_marks`. With the implementation in #552, it seems there is no need for another iterator type, so I just added two methods on `Pairs` (`find_first_tagged` and `find_tagged`).

I was also looking into extending the meta grammar, but encountered a few hurdles: https://github.com/pest-parser/pest/discussions/550#discussioncomment-5559434 so may attempt that in a separate PR.


tomtau:
added a meta grammar extension, but not sure if there's a meaningful difference between node and branch tags

NoahTheDuke:
Yeah, I don't see a lot of difference between the two kinds of tags. Are their optimizations or ambiguities that are covered by one or the other that can't be done in both?

If choosing between the two, I would lean towards NodeTag, as `=` assignment is fairly "natural" compared to binding to the right with no sigil.

What happens when both are used in the same term? Does that change parsing or the resulting nodes?

oovm:
There is no difference in existing weakly typed pests, but it is very necessary to distinguish when generating strongly typed AST.

It can be understood that the node tag means `field`, and the branch tag means `variant`.

---

I have used this feature here:

https://github.com/ygg-lang/yggdrasil-rs/blob/dfa64ab9cf918572c70962f67e01858b9db28d5c/projects/ygg-core/src/bootstrap.ygg#L59-L70

This is a parser preprocessor grammar, here `:` is used for node markers and `#` for branch markers

```js
class ImportStatement {
    | "import" (path:String) ("as" alias:Identifier)?              #FromPath
    | "import" (path:String) '{' @join(items:ImportItem, ',')? '}' #FromPathItems
}
```

Finally, the following strongly typed AST will be generated:

```rust
pub enum ImportStatement {
    FromPath {
        path: String,
        alias: Option<Identifier>,
    },
    FromPathItems {
        path: String,
        items: Vec<ImportItem>,
    },
}
```

If there's no branch mark, it will generate like:

```rust
pub struct ImportStatement {
    pub path: String,
    pub alias: Option<Identifier>,
    pub items: Option<Vec<ImportItem>>,
}
```

This also implies such a constraint that if a certain branch is marked, then all branches must be marked, otherwise the correct enum structure cannot be generated.

NoahTheDuke:
@oovm Thanks for the explanation. That example is helpful. I didn't think through how the branch tag would "consume" every node in the chosen branch whereas the node tag would only "consume" the next.

The usage of post-position still throws me off a little, but I am averse to falling into bike-shedding and don't have a ready proposition.

tomtau:
@oovm thanks for your reply, the example is in a different language though. Does that use case need it in the current 2.X "weakly typed" pest? If so, is the implementation in this PR ok?
If not, I'd then suggest just adding node tags (as plain labelling still has other use cases in the current pest) and add branch tags later (either in pest3 or if other use cases are found in the mean time, e.g. through pest-ast uses).

oovm:
This is only needed for pest3 in the future, it’s too early to think about it now, I think we can shelve the branch mark first.

derive/tests/implicit.pest
```diff
@@ -1,9 +1,9 @@
 program = _{ SOI ~ implicit ~ EOI  }
-implicit= ${ or ~ (WHITESPACE+ ~ or )* }
+implicit= ${ #head = or ~ #tail = (WHITESPACE+ ~ or)* }
 
-or  = !{ and ~ (or_op ~ and)+ | and }
-and = { comp ~ (and_op ~ comp)+ | comp }
-comp = { array ~ eq_op ~ array | array }
+or  = !{ #more_and = and ~ (or_op ~ and)+ | #one_and = and }
```
NoahTheDuke, line 4: The `#more_and` here only grabs the `and`, not all of `and ~ (or_op ~ and)+`, right?

NoahTheDuke (approved):
Cool, looks useful.

derive/tests/implicit.pest
tomtau: yes, I think so, it'd need extra brackets to capture the full expression

PR #833, opened 2023-04-12
Implement `ExactSizeIterator` for `Pair` iterators

MucTepDayH16:
PR adds impl `ExactSizeIterator` and `size_hint` accordingly for `Pairs`, `Tokens`, `FlatPairs` and `LinesSpan`. This can hint rust compiler to allocate proper memory amount while collecting to vector or similar containers

### DIFF
--- a/pest/src/iterators/flat_pairs.rs
+++ b/pest/src/iterators/flat_pairs.rs
@@ -102,6 +102,22 @@ impl<'i, R: RuleType> FlatPairs<'i, R> {
     }
 }
 
+impl<'i, R: RuleType> ExactSizeIterator for FlatPairs<'i, R> {
+    fn len(&self) -> usize {
+        let mut start = self.start;
+        let mut count = 0;
+        while start < self.end {
+            start += 1;
+            while start < self.end && !self.is_start(start) {
+                start += 1;
+            }
+
+            count += 1;
+        }
+        count
+    }
+}
+
 impl<'i, R: RuleType> Iterator for FlatPairs<'i, R> {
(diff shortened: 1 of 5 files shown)


pest/src/iterators/flat_pairs.rs
```diff
@@ -102,6 +102,22 @@ impl<'i, R: RuleType> FlatPairs<'i, R> {
...
+        while start < self.end {
+            start += 1;
+            while start < self.end && !self.is_start(start) {
+                start += 1;
+            }
+
+            count += 1;
+        }
+        count
+    }
```
tomtau, line 118: is it ok for a `len` to have a linear complexity / a loop there? maybe we can check with this benchmark https://github.com/pest-parser/pest/blob/master/grammars/benches/json.rs#L91 to see if it had any impact? (@huacnlee used to to check that the impact of the line index )
alternatively, perhaps there could be a private field to maintain the length iteratively?

pest/src/iterators/tokens.rs
```diff
@@ -92,6 +92,12 @@ impl<'i, R: RuleType> Tokens<'i, R> {
     }
 }
 
+impl<'i, R: RuleType> ExactSizeIterator for Tokens<'i, R> {
+    fn len(&self) -> usize {
+        self.end - self.start
```
tomtau, line 97: likely ok, but perhaps good to check / add a test that this works with non-ASCII UTF-8 characters

tomtau (approved):
thanks for the PR! I think this is ok to merge, but would probably check if there is any impact on the iterator performance

PR #834, opened 2023-04-13
fix: feature-guard debugger CLI dependencies

tomtau:
when trying to build the debugger "lib" for `wasm32-unknown-unknown`, it failed due to those CLI-only dependencies


tomtau:
Hm, this is still not that useful in `wasm32-unknown-unknown`, because spawning a thread would panic: https://github.com/tomtau/pest-web-debug#in-detail

PR #839, opened 2023-04-17
Fix #838 grammar error when only have `//!`, `///` or `//`, allow writing empty pest grammar.

### LINKED ISSUE #838: Lines beginning with //! result in an error
**Describe the bug**
lines beginning with //! result in
```
  |
6 | #[derive(pest_derive::Parser)]
  |          ^^^^^^^^^^^^^^^^^^^
  |
  = help: message: error parsing
            --> 16:1
             |
          16 | //!
             | ^---
             |
             = expected EOI or rule
```

**To Reproduce**
load a .pest file with a line starting with //!

**Expected behavior**
The line is ignored

**Additional context**
I'm using pest 2.5.7
The issue was not present in pest 2.1.3

this looks potentially similar to https://github.com/pest-parser/pest/pull/818.

huacnlee:


meta/src/grammar.pest
```diff
@@ -16,102 +16,126 @@
 //! future (e.g. by increasing MSRV and non_exhaustive annotations).
 
 /// The top-level rule of a grammar.
-grammar_rules = _{ SOI ~ grammar_doc* ~ (grammar_rule)+ ~ EOI }
+grammar_rules = _{ SOI ~ grammar_doc* ~ grammar_rule* ~ EOI }
```
huacnlee, line 19: This first just changed this line from `(grammar_rule)+` to `grammar_rule*`.

The other changes was made by pest-fmt format.

meta/src/parser.rs
```diff
@@ -1171,7 +1171,7 @@ mod tests {
             parser: PestParser,
             input: "0",
             rule: Rule::grammar_rules,
-            positives: vec![Rule::grammar_rule, Rule::grammar_doc],
+            positives: vec![Rule::EOI, Rule::grammar_rule, Rule::grammar_doc],
```
huacnlee, line 1174: Because the `grammar_rules` now is not requirement at least one grammar rule, so this test need to change.

meta/src/grammar.pest
```diff
@@ -1,6 +1,6 @@
 // pest. The Elegant Parser
 // Copyright (c) 2018 Dragoș Tiselice
-//
+//
```
NoahTheDuke, line 3: No need for this change.

NoahTheDuke (approved):

meta/src/grammar.pest
huacnlee: It also by pest-fmt

PR #850, opened 2023-04-28
ci: fix for cargo-msrv needing at least 1.59 now

tomtau:


Tartasprint:
Isn't updating the `rust-version` in each Cargo.toml also necessary ?

tomtau:
@Tartasprint no, pest's MSRV is still 1.56... this one was for building the `cargo-msrv` tool which is for checking MSRV (and the tool itself requires 1.59 to build it, but it can check lower MSRV AFAIK)

PR #854, opened 2023-05-06
Removed TODO about removed box syntax

Tartasprint:
I removed the multiple TODO comments about Rust's box syntax, because it has been removed according to rust-lang/rust/pull/108471


tomtau (approved):
oh, i  didn't know it was removed. based on the discussion, those performance difference seem mainly around boxed array, so probably no need to worry about it here

PR #860, opened 2023-05-26
feat: add simple Position/Span From implementations for LineColLocation

HoloTheDrunk:
# What?
Adds `From<Position<'_>>` and `From<Span<'_>>` implementations to `LineColLocation`.

# Why?
This will make integrating Pest errors and types into custom error types slightly less verbose.
It's technically possible to create a pest error from a span and then get the LCL from that, but it's an unnecessary overhead.

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -74,6 +74,19 @@ pub enum LineColLocation {
     Span((usize, usize), (usize, usize)),
 }
 
+impl From<Position<'_>> for LineColLocation {
+    fn from(value: Position<'_>) -> Self {
+        Self::Pos(value.line_col())
+    }
+}
+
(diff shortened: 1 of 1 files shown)


NoahTheDuke (changes requested):
Looks good. Can you write a simple test showing these off?

HoloTheDrunk:
Never bad to have unit tests, though I can't seem to distill the use case where these implementations are important enough to put it in one given that the issue only crops up when making decently complex wrapping error types.

NoahTheDuke (approved):

NoahTheDuke:
Yeah, I understand how that's hard to distill down for a test. Having a simple unit test works for me, just something to cover the barest case to prevent obvious regressions.

PR #866, opened 2023-06-16
Conditional source-tree bootstrapping

tomtau:
Based on #495
Use cargo to run the bootstrapper.
The default feature was flipped compared to #495,
so that the existing CI workflows continue working as before.

The projects that want to depend on git-based pest can do as follows:
`pest_derive = { git = "...", features = ["not-bootstrap-in-src"] }`

### DIFF
--- a/meta/build.rs
+++ b/meta/build.rs
@@ -1,8 +1,19 @@
+#[cfg(feature = "not-bootstrap-in-src")]
+use cargo::{
+    core::{resolver::CliFeatures, Workspace},
+    ops,
+    ops::{CompileOptions, Packages},
+    util::command_prelude::CompileMode,
+    Config,
+};
 use sha2::{Digest, Sha256};
 use std::env;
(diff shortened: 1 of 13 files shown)


meta/build.rs
```diff
@@ -43,19 +54,56 @@ fn main() {
             let mut hash_file = File::create(hash_path).unwrap();
             writeln!(hash_file, "{}", current_hash).unwrap();
 
-            // This "dynamic linking" is probably so fragile I don't even want to hear it
-            let status = Command::new(manifest_dir.join("../target/debug/pest_bootstrap"))
-                .spawn()
-                .unwrap_or_else(|_| {
+            #[cfg(not(feature = "not-bootstrap-in-src"))]
+            {
+                // This "dynamic linking" is probably so fragile I don't even want to hear it
```
NoahTheDuke, line 59: Lol

NoahTheDuke: That's great, nicely found.

PR #869, opened 2023-06-17
bump criterion and cargo dependencies

tomtau:


tomtau:
@NoahTheDuke FYI bumping up MSRV to 1.60 is for the latest dependencies (like once_cell) as well as being able to disable the optional dependencies as implicit features (so that there aren't so many feature flag combinations)

PR #873, opened 2023-06-20
Add reference to json5-nodes

jlyonsmith:
A library that uses a JSON5 Pest grammar to parse JSON5 into "nodes" that include the location of the JSON element in the original text.


tomtau (approved):
Thanks! You can also consider contributing it to this list: https://github.com/pest-parser/awesome-pest#projects

PR #874, opened 2023-06-21
fix: expose the unicode emoji builtin rules

tomtau:
ref: https://github.com/pest-parser/pest/discussions/872


tomtau:
> `EMOJI, EMOJI_COMPONENT, EMOJI_MODIFIER, EMOJI_MODIFIER_BASE, EMOJI_PRESENTATION, EXTENDED_PICTOGRAPHIC` where are you get this code?
> 
> In Pest, there have a script (`update_unicode.sh`) to generate the Unicode names.

from here: https://pest.rs/book/grammars/built-ins.html#binary-properties

huacnlee:
Oh, never mind on my last comment.
You do the right way. The Emoji unicode script has been generated before, we just not publish in pest `mod.rs`

NoahTheDuke (approved):

PR #887, opened 2023-07-14
Merging Spans :recycle: 

0nyr:
Hello Pest team!

I'm working on a Ctiny language interpreter for a university course. You can find my repository [here | GitHub](https://github.com/0nyr/ctiny_interpreter).

While working on the AST, I have faced a problem.

Considering the following grammar:

```
expression = { disjunction } // top level expression is disjunction
disjunction = { conjunction ~ (disjunction_operator ~ conjunction)* }
conjunction = { equality ~ (conjunction_operator ~ equality)* }
equality = { relation ~ (equality_operator ~ relation)* }
relation = { addition ~ (relation_operator ~ addition)? }
addition = { term ~ (addition_operator ~ term)* }
term = { factor ~ (multiplication_operator ~ factor)* }
factor = { unary_operator? ~ primary }
```

and my AST node structs:

```
// Base AST node
#[derive(Debug, PartialEq)]
pub struct Node<T> {
    pub sp: Span,   // contains information about the node's position (position of span) to be matched to string in the source code
    pub data: T,            // contains the data, wrapped into an inner type
}

// Expressions
#[derive(Debug, PartialEq)]
pub enum Expression {
    Literal(Literal), // direct value
    UnaryExpression(UnaryExpression),
    BinaryExpression(BinaryExpression),
    FunctionCall(FunctionCall),
    TypeCast(TypeCast),
    GetOrSetValue(GetOrSetValue),
}
```

While parsing the pairs of an expression, I want to be able to combine Spans to create Nodes of Expression::BinaryExpression for easier debugging later (giving the span of the Node to a pest::CustomError). But so far, there is no merge function for Span, and since there is no getter for the its `input` field... it would require me to pass everywhere the &str of the original input string... which is really annoying, especially considering I'm only manipulating pairs and spans that all have the reference I'm looking for. So the best approach to this is really to add a getter `get_input` and a `merge` function to Span.

You can have a look at this fork in action, specifically on span merging [here](https://github.com/0nyr/ctiny_interpreter/blob/2a0a4abbf4d2e4602a8a67b3e87b46c09c9f938b/src/abstract_syntax_tree/expressions.rs). Especially at the following code: 
```
let common_span = merge_spans_no_check!(
                &left_operation.sp,
                &right_operation.sp
            ).ok_or(make_ast_error_from_pair(
                $input_pair.clone(), 
                format!("🔴 Couldn't build a span from the left and right operations. 
                    str_len: {}, left_start: {}, left_end: {}, right_start: {}, right_end: {}",
                    $input_pair.as_str().len(),
                    left_operation.sp.start(),
                    left_operation.sp.end(),
                    right_operation.sp.start(),
                    right_operation.sp.end(),
                ).as_str())
            )?;
```

I'm not the only one facing this issue. A friend of mine, also working on this course is facing a similar issue.

That's what I have done with this PR. I also added meaningful documentation in comments as well as tests.

### DIFF
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -215,7 +215,7 @@ impl<'i> Span<'i> {
     /// Returns the input string of the `Span`.
     ///
     /// This function returns the input string of the `Span` as a `&str`. This is the source string
-    /// from which the `Span` was created. The returned `&str` can be used to examine the contents of 
+    /// from which the `Span` was created. The returned `&str` can be used to examine the contents of
     /// the `Span` or to perform further processing on the string.
     ///
     /// # Examples
@@ -223,7 +223,7 @@ impl<'i> Span<'i> {
     /// ```
     /// # use pest;
     /// # use pest::Span;
-    /// 
+    ///
     /// // Example: Get input string from a span
     /// let input = "abc\ndef\nghi";
     /// let span = Span::new(input, 1, 7).unwrap();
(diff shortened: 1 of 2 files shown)


pest/src/span.rs
```diff
@@ -288,6 +309,57 @@ impl<'i> Hash for Span<'i> {
...
+/// let input = "abc\ndef\nghi";
+/// let span1 = Span::new(input, 1, 7).unwrap();
+/// let span2 = Span::new(input, 8, 11).unwrap();
+/// let merged = merge_spans(&span1, &span2);
+/// assert!(merged.is_none());
+/// ```
+pub fn merge_spans<'i>(a: &Span<'i>, b: &Span<'i>) -> Option<Span<'i>> {
+    if a.end() >= b.start() && a.start() <= b.end() {
+        // The spans overlap or are contiguous, so they can be merged.
+        Span::new(a.get_input(), std::cmp::min(a.start(), b.start()), std::cmp::max(a.end(), b.end()))
```
tomtau, line 356: To keep it no-std-compatible, is this possible?
```suggestion
        Span::new(a.get_input(), core::cmp::min(a.start(), b.start()), core::cmp::max(a.end(), b.end()))
```

tomtau (changes requested):
looks good. Can you also run `cargo fmt` on that code?

pest/src/span.rs
0nyr: I just did it. Tests are good. It's commited.

tomtau (approved):

PR #900, opened 2023-07-23
Add feature gated `Cow` module paths

### LINKED ISSUE #899: Expansion error in `pest-derive` with tags
**Describe the bug**
On Rust 1.70 and 1.71, when using tags, the derive macro errors out with the following:

```
error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
 --> src/lib.rs:3:10
  |
3 | #[derive(Parser)]
  |          ^^^^^^ use of undeclared crate or module `alloc`
  |
  = help: add `extern crate alloc` to use the `alloc` crate
  = help: consider importing this enum:
          std::borrow::Cow
  = note: this error originates in the derive macro `Parser` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**To Reproduce**
Here is a minimum reproducible example:

```toml
# Cargo.toml
[package]
name = "pest-issue"
version = "0.1.0"
edition = "2021"

[dependencies]
pest = "2.7.1"
pest_derive = { version = "2.7.1", features = ["std", "grammar-extras"] }
```

```pest
program = { SOI ~ rule* ~ EOI }
rule = !{ "{" ~ (item ~ ";")* ~ (#tail_item = item)? ~ "}" }
item = _{ "foo" | "bar" }
```

```rust
// lib.rs
use pest_derive::Parser;

#[derive(Parser)]
#[grammar = "grammar.pest"]
struct ExampleParser;
```

**Expected behavior**
The generated code should not emit the error.

**Additional context**
- Adding `extern crate alloc` at the top of `lib.rs` makes the error go away -- it is a feasible workaround, but it should probably not be asked of the users to do that.
- This is the only snippet that appears to reference `alloc` in the expanded code:
  ```rust
  state
      .optional(|state| {
          self::item(state)
              .and_then(|state| {
                  state.tag_node(alloc::borrow::Cow::Borrowed("tail_item"))
              })
      })
  ```
- I found [this](https://github.com/pest-parser/pest/blob/4d51b5b75c56b03e13eb21d3e36a5bbe8285a640/generator/src/generator.rs#L544). I haven't had time to try it myself yet, but could prepending `::` so that we have `::alloc` be a potential solution?

Thank you for all your work on this library!

veeenu:
The generator currently wraps tags in an instance of `alloc::borrow::Cow`. This PR changes that to `std::borrow::Cow` conditional on the `std` feature flag being active.

Closes #899.

### DIFF
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -540,8 +540,14 @@ fn generate_expr(expr: OptimizedExpr) -> TokenStream {
         #[cfg(feature = "grammar-extras")]
         OptimizedExpr::NodeTag(expr, tag) => {
             let expr = generate_expr(*expr);
+            let tag_cow = {
+                #[cfg(feature = "std")]
+                quote! { ::std::borrow::Cow::Borrowed(#tag) }
+                #[cfg(not(feature = "std"))]
+                quote! { ::alloc::borrow::Cow::Borrowed(#tag) }
+            };
             quote! {
-                #expr.and_then(|state| state.tag_node(alloc::borrow::Cow::Borrowed(#tag)))
+                #expr.and_then(|state| state.tag_node(#tag_cow))
             }
         }
     }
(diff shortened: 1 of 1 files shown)


tomtau (approved):
thanks!

PR #901, opened 2023-07-25
propagate non_exhaustive attribute to the generated Rule enums

tomtau:


NoahTheDuke:
This seems like a big change. Is bumping the patch version enough?

Otherwise, code looks good.

tomtau:
> This seems like a big change. Is bumping the patch version enough?
> 
> Otherwise, code looks good.

The external interfaces remain the same, so at least semver-wise it should be, but it's a new "feature" in a way. Would you suggest bumping a minor version instead?

NoahTheDuke:
My apologies, I misread and thought `pub(crate)` was simply `pub` in generator. This is fine by me.

NoahTheDuke (approved):

Issue #326, opened 2018-10-28
Incompatibility with missing_docs lint


librelois:
I must have disabled `missing_docs` lint to compile my crate, it's very annoying, the compiler no longer checks that all the documentation is present : 

```
error: missing documentation for an enum
  --> documents/v10/mod.rs:38:10
   |
38 | #[derive(Parser)]
   |          ^^^^^^
   |
note: lint level defined here
  --> documents/lib.rs:20:5
   |
20 |     missing_docs,
   |     ^^^^^^^^^^^^

error: missing documentation for a variant
  --> documents/v10/mod.rs:38:10
   |
38 | #[derive(Parser)]
   |          ^^^^^^

error: aborting due to 2 previous errors
```

dragostis:
Hm, this is interesting. The `Rule` enum is mostly thought out to be used in the implementation of your crate and not in user-facing APIs. How are you using the derive?

librelois:
My Struct `ParserV10` is private.  
I migrate to pest a lib crate that is used to parse documents of a specific protocol.
This crate is called `duniter_documents` and is not intended for use in user-facingAPI.

Thank you for your reactivity, my code is here: https://git.duniter.org/nodes/rust/duniter-rs/blob/migrate_to_pest/documents/v10/mod.rs#L40
(Migration from regex to pest is a work in progress)

CAD97:
As a short-term solution, you can add `#[allow(missing_docs)]` to the module in which you're defining the parser.

attila-lin:
Hi. Any progress? I think the pr may work.

tomtau:
I pinged the author. If they don't respond, I can reopen a PR with a fix to add the missing comment to EOL

woodruffw:
I'm seeing a variant of this still with `pest 2.8.0`:

```
error: missing documentation for a variant
  --> crates/github-actions-expressions/src/lib.rs:12:10
   |
12 | #[derive(Parser)]
   |          ^^^^^^
   |
note: the lint level is defined here
  --> crates/github-actions-expressions/src/lib.rs:4:9
   |
4  | #![deny(missing_docs)]
   |         ^^^^^^^^^^^^
   = note: this error originates in the derive macro `
```

The code in question:

```rust
#[derive(Parser)]
#[grammar = "expr.pest"]
struct ExprParser;
```

Adding `pub(crate)` also doesn't suppress it, unfortunately.

woodruffw:
[...]

PR #923, opened 2023-09-03
Add allow(missing_docs) on Rule struct

pgimalac:
This PR adds `#[allow(missing_docs)]` to the `Rule` struct in order to fix https://github.com/pest-parser/pest/issues/326.

The only element which was missing documentation for me was the `EOI` variant, tell me if you would rather add some hardcoded documentation to that variant.

### DIFF
--- a/generator/src/generator.rs
+++ b/generator/src/generator.rs
@@ -220,7 +220,7 @@ fn generate_enum(
     let grammar_doc = &doc_comment.grammar_doc;
     let mut result = quote! {
         #[doc = #grammar_doc]
-        #[allow(dead_code, non_camel_case_types, clippy::upper_case_acronyms)]
+        #[allow(dead_code, non_camel_case_types, clippy::upper_case_acronyms, missing_docs)]
         #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
     };
     if non_exhaustive {


tomtau:
If we add a comment on EOI, does it work? Maybe that would be preferable? The comment could just say "end-of-input"?

tomtau:
@pgimalac

tomtau:
@pgimalac would this work: https://github.com/pest-parser/pest/pull/926/files#diff-b40e16517d351a155bd4ca1e43b33643158df95855f86a37f3b87f5a90afba30R237 ?

pgimalac:
@tomtau yes it works, that's what I did initially (before seeing the comment from @CAD97 on the issue I linked saying to just add an `allow(missing_docs)`)

One thing I'm not sure is whether other builtin rules could appear, for example I have the `WHITESPACE` rule in my project, but as I redefined it I was able to add a comment. This is the main reason why I think `#[allow(missing_docs)] might be a safer fix.

tomtau:
I think they don't, but if they do, we can see if we can add rustdocs to them?
One way to verify is you can use my pest fork as a git dependency, remove that whitespace rule and see if there are any warnings about missing comments?

tomtau:
added the comment to EOL: https://github.com/pest-parser/pest/pull/926 -- if it doesn't work, feel free to reopen this PR to add `#[allow(missing_docs)]`

attila-lin:
upgrade to `"2.7.4"`, still get warning

```
error: missing documentation for a variant
 --> rua\src\analyze\annotation\single_annotation_parser.rs:7:10
  |
7 | #[derive(Parser)]
  |          ^^^^^^
  |
note: the lint level is defined here
 --> rua\src\lib.rs:5:9
  |
5 | #![deny(missing_docs)]
  |         ^^^^^^^^^^^^
  = note: this error originates in the derive macro `Parser` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tomtau:
@attila-lin and your own grammar is all documented?

attila-lin:
Yes, I think so. Wonder if anyhow have the same situation?

tomtau:
@attila-lin i guess your code is not public, is it? I'm trying to look for some code to reproduce it -- the one posted originally by @librelois doesn't seem to be accessible now: https://git.duniter.org/nodes/rust/duniter-rs/blob/migrate_to_pest/documents/v10/mod.rs#L40

attila-lin:
Yes, not public.

I make it `pub(crate)` solve this problem. Thank you for your response. @tomtau :blush:

PR #924, opened 2023-09-07
Add clarification to the docs

DanikVitek:
Clarify that the `Pairs::{find_tagged, find_first_tagged}` methods flatten the `Pairs` iterator


NoahTheDuke (changes requested):
Both sentences should end in periods.

tomtau (approved):

PR #958, opened 2024-01-03
perf: use str instead of Cow to reduse Token size

sssooonnnggg:
"Cow<str>" has a relatively large layout size than &str, which can impact performance.
It seems that using '&str' to represent the tag is sufficient.
For my USD parser, the parsing time for a 50MB USD file decreased from 800 milliseconds to 600 milliseconds.
https://github.com/sssooonnnggg/rusd


NoahTheDuke:
Please don't use AI tools to make PRs to Pest.

NoahTheDuke:
Oh dang, my apologies. I thought that was brought in by the OP. If you enabled it, then I'll leave it be in the future. (I'm not a fan of such AI tools but that's just personal preference.) I only ask that folks don't come to use with their own AI tools.

sssooonnnggg:
> pest_vm fails to compile: https://github.com/pest-parser/pest/actions/runs/7396775314/job/20127215025?pr=958#step:4:238 I'm not sure if a string slice can be used there

Thank you for the review, 
Fixed the compilation error in the `vm/src/lib.rs`. I think we can use a string slice here, as it's used similarly in `lib.rs:188` with `match_string`.

sssooonnnggg:
It's an API breaking change, but from the user's perspective, this API is usually not directly utilized by consumers. In reality, it's used within the code generated by a derived macro, so I don't think it's a major issue.

Additionally, this optimization significantly improves execution speed and reduces memory usage. I believe it's unrelated to the std feature.

sssooonnnggg:
> > Thank you for the review,
> > Fixed the compilation error in the vm/src/lib.rs. I think we can use a string slice here, as it's used similarly in lib.rs:188 with match_string.
> 
> It still doesn't compile, unfortunately:
> 
> https://github.com/pest-parser/pest/actions/runs/7405863941/job/20155605432?pr=958#step:4:226

My fault :), I forgot to check all features, now it should pass all checks by `cargo check --all --features pretty-print,const_prec_climber,memchr,grammar-extras --all-targets` command, pelease review again

PR #961, opened 2024-01-04
Export generator and docs api in the pest_generator trait

marcfir:
### What it does
This PR exposes the docs module and the generator module of the pest_generator trait.
I added the missing docs for the newly exported functions, structs and modules.

### Background
We are currently working on a language workbench in rust based on pest. So far we have used a separate derive macro by. But it turns out that it is easier if we do our own preprocessing of the derive attributes and just generate the parser in our own derive macro with the `pest_generaotr::generator::generate` function.
 
### Improvements
We can add a feature flag as this API was not intended to be semver sensitive.

### DIFF
--- a/generator/src/docs.rs
+++ b/generator/src/docs.rs
@@ -1,9 +1,13 @@
+//! Type and helper to collect the gramamr and rule documentation.
+
 use pest::iterators::Pairs;
 use pest_meta::parser::Rule;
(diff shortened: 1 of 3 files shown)


marcfir:
Thanks for the quick review! I have added your comments. Do you have any other suggestions?

PR #964, opened 2024-01-09
Add docs to parsing source code

EmirVildanov:
While exploring source code (mostly `parser_state.rs`) I've left some comments for better understanding of parsing process.
I thought maybe they'll be useful for other contributors.


EmirVildanov:
> thanks! there's a small formatting issue: https://github.com/pest-parser/pest/actions/runs/7459853275/job/20296935592?pr=964 otherwise it should be good to go

Yeah, thanks for noting. I've fixed it.

Issue #541, opened 2021-09-25
Case-insensitive PUSH / PEEK / POP


mdharm:
For certain grammars, it would be helpful if `PUSH()`, `PEEK`, and `POP` all had case-insensitive variants.  The [HTML grammar used by the html_parser crate](https://github.com/mathiversen/html-parser/blob/master/src/grammar/rules.pest) is one example; it does not properly parse sequences such as `<BODY></body>` and there is no "easy" fix -- this is partially a side-effect of the grammar being written to accept arbitrary tags (rather than limiting to only the well-defined HTML tags such as), so it does not explicitly list all of the valid tag tokens.  Instead, it uses `PUSH()` and `POP` to find matching opening / closing tags.

I'm sure there are other cases where this would be handy.  It doesn't look like a major effort either, as there is already a case-insensitive comparison method `Position::match_insensitive()` that could be used in place of the existing `Position::match_string()`

yshavit:
Would you want case-insensitive variants for all the functions, or would it be enough to just have it on `PUSH`? If just having it on `PUSH` is enough, that'd be a smaller footprint in terms of the grammar rules people have to learn: otherwise we'd need variants of `PEEK`, `PEEK_ALL`, `POP`, and `POP_ALL` -- which isn't a _ton_, but not nothing.

Basically, I'm wondering if it'd be enough to have `PUSH_I` that semantically pushes a case-insensitive string; have the insensitivity be an attribute of the string being matched, not on the rule that matches it.

PR #1092, opened 2025-03-21
new rule: PUSH_LITERAL

yshavit:
Add a rule `PUSH_LITERAL("a")`. This rule always matches and never consumes any input, and it pushes `"a"` to the stack. The argument must be a literal.

> [!note]
> I haven't added any unit tests or documentation yet; I wanted to get a sense of whether you'd be amenable to this new rule in principle, before I did that. But, I tested it locally on a grammar I have, and it does work.

This new rule accomplishes the goals of pest-parser/pest#880 (in a slightly different way). Instead of `PUSH_OTHER("-", " ")`, this is just `PUSH_LITERAL(" ")`. If you wanted to match A and then push B, you just combine them via a sequence: `"-" ~ PUSH_LITERAL(" ")`.

This can be used to generalize quoted strings to a pair of matching but non-equal delimiters. For example, many programming languages let you express strings using either double or single quotes, and don't require escaping of the other:

```
string = ${ PUSH( "\"" | "'") ~ quoted_char* ~ POP }
quoted_char = @{!PEEK ~ ANY } // simple version without escape sequences

// will parse "a'b" or 'a"b'
```

This happens to work because the starting and ending char are the same, but that's not always desirable. For example, we may want to allow strings to be delimited by either `(` ... `)` or `<` ... `>`, with similar rules as above, such that `(a>)` and `<a)>` are both valid.

With `PUSH_LITERAL`, this becomes:

```
string = ${ quote_open ~ quoted_char* ~ POP }
quote_open = @{
  ( "(" ~ PUSH_LITERAL(")") )
| ( "<" ~ PUSH_LITERAL(">") )
}
quoted_char = @{!PEEK ~ ANY } // simple version without escape sequences
```

Happy to hear your thoughts!

### DIFF
--- a/pest/src/parser_state.rs
+++ b/pest/src/parser_state.rs
@@ -10,7 +10,7 @@
 //! The core functionality of parsing grammar.
 //! State of parser during the process of rules handling.
 
-use alloc::borrow::ToOwned;
+use alloc::borrow::{ToOwned};
 use alloc::boxed::Box;
 use alloc::collections::BTreeSet;
 use alloc::rc::Rc;
@@ -213,6 +213,52 @@ impl Display for ParsingToken {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+enum BorrowedOrRc<'i> {
+    Borrowed(&'i str),
+    Owned(Rc<String>)
+}
+
+impl From<&'static str> for BorrowedOrRc<'_> {
+    fn from(value: &'static str) -> Self {
+        BorrowedOrRc::Borrowed(value)
+    }
+}
+
+impl From<String> for BorrowedOrRc<'_> {
+    fn from(value: String) -> Self {
+        BorrowedOrRc::Owned(Rc::new(value))
+    }
+}
+
+/// A holder for a literal string, for use in `push_literal`. This is typically a `&'static str`, but is an owned
+/// `Rc<String>` for the pest vm.
+impl<'i> BorrowedOrRc<'i> {
+    fn as_str<'a: 'i>(&'a self) -> &'a str {
+        match self {
+            BorrowedOrRc::Borrowed(s) => *s,
+            BorrowedOrRc::Owned(s) => s,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+enum SpanOrLiteral<'i> {
+    Span(Span<'i>),
+    Literal(BorrowedOrRc<'i>),
+}
+
+impl<'i> SpanOrLiteral<'i> {
+    #[inline]
+    fn as_borrowed_or_rc(&self) -> BorrowedOrRc<'i> {
+        match self {
+            Self::Span(s) => BorrowedOrRc::Borrowed(s.as_str()),
+            Self::Literal(s) => s.clone(),
+        }
+    }
+}
+
 
 /// Structure that tracks all the parsing attempts made on the max position.
 /// We want to give an error hint about parsing rules that succeeded
@@ -423,7 +469,7 @@ pub struct ParserState<'i, R: RuleType> {
     atomicity: Atomicity,
     /// Helper structure tracking `Stack` status (used in case grammar contains stack PUSH/POP
     /// invocations).
-    stack: Stack<Span<'i>>,
+    stack: Stack<SpanOrLiteral<'i>>,
     /// Used for setting max parser calls limit.
     call_tracker: CallLimitTracker,
     /// Together with tracking of `pos_attempts` and `attempt_pos`
@@ -1064,6 +1110,34 @@ impl<'i, R: RuleType> ParserState<'i, R> {
         }
     }
 
+    /// Pushes the given literal to the stack, and always returns `Ok(Box<ParserState>)`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use pest;
+    /// # #[allow(non_camel_case_types)]
+    /// # #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+    /// enum Rule {}
+    ///
+    /// let input = "ab";
+    /// let mut state: Box<pest::ParserState<'_, Rule>> = pest::ParserState::new(input);
+    ///
+    /// let mut result = state.stack_push_literal("a");
+    /// assert!(result.is_ok());
+    /// assert_eq!(result.as_ref().unwrap().position().pos(), 0);
+    ///
+    /// let mut result = result.unwrap().stack_pop();
+    /// assert!(result.is_ok());
+    /// assert_eq!(result.unwrap().position().pos(), 1);
+    /// ```
+    #[inline]
+    #[allow(private_bounds)]
+    pub fn stack_push_literal(mut self: Box<Self>, string: impl Into<BorrowedOrRc<'i>>) -> ParseResult<Box<Self>> {
+        self.stack.push(SpanOrLiteral::Literal(string.into()));
+        Ok(self)
+    }
+
     /// Attempts to case-insensitively match the given string. Returns `Ok` with the updated
     /// `Box<ParserState>` if successful, or `Err` with the updated `Box<ParserState>` otherwise.
     ///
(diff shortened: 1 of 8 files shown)


yshavit:
> 1. This would need to be feature-guarded `#[cfg(feature = "grammar-extras")]`, because otherwise the different mixing of pest dependencies that cargo may resolve can fail to compile.

Happy to do that!

>     2. there was also this issue [Case-insensitive PUSH / PEEK / POP #541](https://github.com/pest-parser/pest/issues/541) -- not sure if making possible to push insensitive literals can be helpful here?

I don't think it's possible to combine this feature with that one, though I'm open if you have suggestions. Case-insensitivity feels pretty different, and I don't see any reasonably simple abstraction that combines it with matching delims.

But it sounds like, feature guarding aside (and I'd need to get the CIs passing, docs, etc.), this is something you'd be interested in accepting?

pest/src/parser_state.rs
```diff
@@ -214,6 +214,52 @@ impl Display for ParsingToken {
     }
 }
 
+#[derive(Debug, Clone)]
+enum BorrowedOrRc<'i> {
```
yshavit, line 218: I'd love some guidance / advice on this struct, because I'm still pretty new to rust, and not sure what the best approach here is.

Ultimately, I'm trying to be able to use the `&' str` when I can, which is:
* when `PUSH_LITERAL` provides a `&'static str`
* when `Span<'i>` provides a `&'i str`

The owned `Rc` only comes into play in the `vm` crate, which provides a `String`. (And if I'm reading it right, it's meant to be used in a playground/REPL context, so it really does need to be a String, not a `&'static str`.) I couldn't get a Cow to work without having to clone actual Strings, since I'd be working with a borrowed `&Cow<'static, str>` instead of the owned.

I originally had this struct as private with no `From`s, and then had two variants of `pub fn stack_push_literal`:

```rust
pub fn stack_push_literal_string(mut self: Box<Self>, string: String) -> {...}
pub fn stack_push_literal_str(mut self: Box<Self>, string: &'static str) -> {...}
```

But it felt ugly to expose two essentially equivalent methods, so I used the `impl Into` to unify them. That's fine, except that it now at least half-exposes this `BorrowedOrRc` struct, which feels like an internal implementation detail. I do have it as private while exposing it in the `impl Into`, but that also feels a bit wonky (and it compiles for me, but it looks like it broke CI).

So I basically don't know of any approach that ticks all the aesthetic boxes. Do you have any suggestions?

yshavit:
> 2. there was also this issue [Case-insensitive PUSH / PEEK / POP #541](https://github.com/pest-parser/pest/issues/541) -- not sure if making possible to push insensitive literals can be helpful here?

Actually, the `SpanOrLiteral` I have could be extended pretty easily to handle case-insensitivity. I'll hop over to that other issue for a clarifying question, though.

pest/src/parser_state.rs
yshavit: Huh, I tried it yesterday and couldn't get it to work... but now it did. Maybe I was just staring at it too long yesterday! :-)

yshavit:
Bumping this PR to Draft status until it's ready for a re-review.

pest/src/parser_state.rs
yshavit: Ah, I remember now -- it compiles, but it'll introduce additional String copies.

At the core of it, there are two requirements that together cause a problem:

1. When we work with these SpanOrLiteral instances (e.g. in `stack_peek`), we're working with borrows of the structs, not owned copies: `self.stack.peek().expect(...)` returns a `&SpanOrLiteral`
2. We can't hold onto anything that indirectly holds onto a borrow of `self`, because we then take that `&str` and pass it to `match_string`, which takes a `mut self` (and thus can't have any outstanding borrows on `&self`).

In the old version of this code, and also in the `SpanOrLiteral::Span` variant, this is fine because the underlying string is tied to the `'i` lifetime. So, we can just basically carve off a new version of that slice from `Span::as_str`, and it doesn't require a borrow against ParserState's `self`.

But with `SpanOrLiteral::Literal`, that doesn't work. For example, let's say we did use a Cow:

```rust
#[derive(Debug, Clone)]
enum SpanOrLiteral<'i> {
    ...
    Literal(Cow<'i, str>),
}
```

When we get this from `self.stack.peek()`, ultimately what we'd have is a `&'1 Cow<'i, str>`, where `1` is the lifetime of the borrow on the ParserState's `self`. Since we're about to invoke a `mut self` fn, we need to clone this in order to get rid of the borrow. That's fine in the `Cow::Borrowed` variant, but for `Cow::Owned`, that clone will end up copying the whole string.

That's what the Rc is here for. It lets us clone that `SpanOrLiteral::Literal` instance without actually copying any underlying bytes, since `Rc::clone` [just increments the reference count][1], it doesn't copy the whole String.

[1]: https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-Clone-for-Rc%3CT,+A%3E

yshavit:
Should be ready now! I added tests and the grammar-extra cfg.

grammars/Cargo.toml
```diff
@@ -15,7 +15,7 @@ rust-version = "1.65"
 
 [dependencies]
 pest = { path = "../pest", version = "2.7.15" }
-pest_derive = { path = "../derive", version = "2.7.15" }
+pest_derive = { path = "../derive", version = "2.7.15", features = ["grammar-extras"] }
```
yshavit, line 18: Alternatively: Would it be better to create a separate `./grammars_with_extras` crate? That way, we can also validate that the ones in this crate require only vanilla pest.

If you'd like to do that, I wonder if it'd make sense to just disable the new grammar, and then put that new work in as a separate PR? Just to keep the diff down on this one.

pest/src/parser_state.rs
```diff
@@ -86,6 +87,26 @@ pub enum MatchDir {
     TopToBottom,
 }
 
+/// A literal value, for use in `push_literal`. Use the `Static` variant when possible (for example, within a derived
```
yshavit, line 90: coderabbitai wants all of this new code guarded by `#[cfg(feature = "grammar-extras")]` -- what do you think?

I don't think there's any harm in doing so, but I also don't know if there's a need. My understanding (please correct me if I'm wrong!) is that the feature flag is there so that users can opt into new grammar features, in case those collide with their existing grammars. For example, someone could have created a grammar with a rule named `PUSH_LITERAL`. But that concern doesn't apply to this file.

grammars/Cargo.toml
yshavit: Roger, I'll move this.

grammars/src/lib.rs
```diff
@@ -64,6 +64,26 @@ pub mod sql {
     pub struct SqlParser;
 }
 
+/// Grammar rules of an SQL parser
```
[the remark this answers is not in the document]
yshavit: Whoops! ty

pest/src/parser_state.rs
yshavit: Ah, okay -- good to know. I think by that reasoning, I still don't need to use `#[cfg` for this, right?

pest/src/parser_state.rs
```diff
@@ -423,7 +471,7 @@ pub struct ParserState<'i, R: RuleType> {
     atomicity: Atomicity,
     /// Helper structure tracking `Stack` status (used in case grammar contains stack PUSH/POP
     /// invocations).
-    stack: Stack<Span<'i>>,
+    stack: Stack<SpanOrLiteral<'i>>,
```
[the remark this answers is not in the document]
yshavit: Just because of the Send/Sync, right? (I want to make sure I didn't miss any other aspect -- I wouldn't think so, since this is all just private to `parser_state.rs`)

yshavit:
Updated `parser_state.rs` to use Arc instead of Rc. `cargo semver-checks` now passes for me (I was able to repro the failure before I made that change), except for `pest-meta`, which fails to build because of windows path issues in importing the grammars.

yshavit:
Will do!

PR #1096, opened 2025-04-06
Debugger next and empty line repeat

rushsteve1:
Adds two new features to the Pest debugger
- A `next` command that single-steps the debugger. Most useful after hitting a breakpoint.
- Empty commands repeat the last command (with a message) similar to GDB. Most useful with `next` or `continue`.


rushsteve1:
In comparison to #1080 this PR is dramatically simpler, focused on minor QoL features.

PR #1102, opened 2025-06-10
remove once_cell (upon next MSRV bump!)

2ndDerivative:
When the next MSRV bump might be coming up once_cell can be cut in favor of std's own Lazy primitives.

This should probably be done only then, after all this is pretty minor, but here's the change so it's on the list.

## Summary by CodeRabbit

- **Chores**
  - Updated the minimum required Rust toolchain version to 1.80.
  - Removed an unused dependency to streamline the project.
  - Replaced an external crate with the standard library for lazy static initialization, reducing external dependencies.

### DIFF
--- a/meta/src/validator.rs
+++ b/meta/src/validator.rs
@@ -12,7 +12,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
-    sync::LazyLock
+    sync::LazyLock,
 };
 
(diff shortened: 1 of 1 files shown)


2ndDerivative:
I think this is correct? should definitely be reviewed, I don't know a lot about the CI stuff here.

I just found this crate in my dependency tree

PR #1120, opened 2025-09-18
deps: drop `thiserror` for a manual impl

CosmicHorrorDev:
`thiserror` largely wasn't getting much use in `pest`, and it's a rather heavy dependency for an otherwise lightweight crate. The only other crate in the workspace that uses `thiserror` is `debugger` which doesn't seem worth trying to remove

## Summary by CodeRabbit

- New Features
  - None

- Refactor
  - Reworked error handling to use explicit standard error trait implementations and clarify integrations between core and optional adapters.

- Chores
  - Removed an external error dependency and simplified feature flags to reduce configuration complexity.

- Performance
  - Potentially faster builds and smaller dependency footprint due to dependency and feature cleanup.

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -29,7 +29,6 @@ use crate::RuleType;
 
 /// Parse-related error type.
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
-#[cfg_attr(feature = "std", derive(thiserror::Error))]
 pub struct Error<R> {
     /// Variant of the error
     pub variant: ErrorVariant<R>,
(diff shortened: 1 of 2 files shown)


pest/src/error.rs
```diff
@@ -43,9 +42,11 @@ pub struct Error<R> {
     parse_attempts: Option<ParseAttempts<R>>,
 }
 
+#[cfg(feature = "std")]
+impl<R: RuleType> core::error::Error for Error<R> {}
```
CosmicHorrorDev, line 46: oh awkward mix of `std` and `core` here, but worth asking. Would you be fine bumping the MSRV up to 1.81.0 for `core::error::Error`? It's only one version up and I believe debian's next release will be using 1.84

PR #1121, opened 2025-09-19
feat: impl Error from core

CosmicHorrorDev:
Followup to #1120. This switches to using `core::error::Error` and bumps the MSRV up to 1.81 to match. It looks like all of CI was already using at least 1.82, so I don't think that requires any additional changes


CosmicHorrorDev:
Done!

Issue #999, opened 2024-03-24
Document panics


tomtau:
Currently, panics may not be consistently documented (as a follow-up to PR that removed unnecessary unsafe).

> Whether Pest documents panics in a # Panics section following the standard library pattern. Note that unlike safety docs, panic docs are not required for soundness.

It doesn't, at least not consistently, but it should.

_Originally posted by @tomtau in https://github.com/pest-parser/pest/pull/998#issuecomment-2016653683_

PR #1168, opened 2026-04-19
test(position): add should_panic tests for documented panic conditions

MasterOfFeelingFish:
Three panic conditions in \Position\ were documented but had no tests:

- \span()\ panics when positions come from different inputs
- \line_col()\ panics when position is out of bounds
- \line_of()\ panics when position is out of bounds

Added \#[should_panic]\ tests for each. The out-of-bounds tests use \
ew_internal\ to construct invalid positions since the public \Position::new()\ validates bounds.

Related to #999

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for boundary validation in position handling, ensuring robustness of error conditions.

### DIFF
--- a/pest/src/position.rs
+++ b/pest/src/position.rs
@@ -659,4 +659,31 @@ mod tests {
 
         positions.insert(start);
     }
+
+    #[test]
+    #[should_panic(expected = "span created from positions from different inputs")]
+    fn span_panic_different_inputs() {
+        let input1 = "a";
+        let input2 = "b";
+        let pos1 = Position::from_start(input1);
+        let pos2 = Position::from_start(input2);
+
+        pos1.span(&pos2);
+    }
+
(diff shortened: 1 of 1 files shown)


pest/src/position.rs
```diff
@@ -659,4 +659,31 @@ mod tests {
...
+        pos.line_col();
+    }
+
+    #[test]
+    #[should_panic(expected = "position out of bounds")]
+    fn line_of_panic_out_of_bounds() {
+        let input = "ab";
+        let pos = Position::new_internal(input, 5);
+        pos.line_of();
+    }
```
tomtau, line 688: these two tests don't seem to pass

tomtau (changes requested):
looks like the expected panic message may be different?

PR #1175, opened 2026-07-02
Reject unescaped quote inside char literal in meta grammar

### LINKED ISSUE #853: Minor inconsistency between pest meta grammar and some tools
**Describe the bug**
There is a difference between how `pest_meta` and the other tools/libraries handle the `'` in range expressions.
In [meta/src/grammar.pest](https://github.com/pest-parser/pest/blob/7f8411a4d5b9c9ac6fcec06e293878982cd320c8/meta/src/grammar.pest) the following is a valid range expression `'''..'''` as there is no check for it:
https://github.com/pest-parser/pest/blob/7f8411a4d5b9c9ac6fcec06e293878982cd320c8/meta/src/grammar.pest#L167-L179
There is no check in the meta's parser either:

https://github.com/pest-parser/pest/blob/7f8411a4d5b9c9ac6fcec06e293878982cd320c8/meta/src/parser.rs#L428-L436

Meanwhile other tools do not accept this syntax.
The highlighter for the playground [has this line](https://github.com/pest-parser/site/blob/076892ebd2d505b19bd576db54f9ee4df5d575a8/static/scripts/editor.ts#L1-L40) for handling ranges:

> ```
> { regex: /'(?:[^'\\]|\\(?:[nrt0'"]|x[\da-fA-F]{2}|u\{[\da-fA-F]{6}\}))'/, token: "string" },
> ```

And a similar rule seem to be defined for IDEs syntax highlighting.

Searching for usage the unescaped version `'''` in public GitHub repositories [gave no results](https://github.com/search?q=path%3A%2F.*%5C.pest%2F+%28%27%27%27..+OR+..%27%27%27%29&type=code), and [some results](https://github.com/search?q=path%3A%2F.*%5C.pest%2F+%28%27%5C%27%27..+OR+..%27%5C%27%27%29&type=code) for the escaped version `'\''`.

**To Reproduce**
Writing the following pest grammar in a file using a supported IDE or in the playground:
```
r={'''..'''}
```
Is deemed valid by Pest, the rule `r` matches the input `'`, but breaks the syntax-highlighting. Might break other things, I didn't check.

**Expected behavior**
I expected the behavior of the different tools listed up there to be consistent.

**Additional context**
I don't know if `'''` being valid was a design choice that wasn't followed in the playground/ide tools, or if checking for this edge case was simply forgotten in pest meta grammar. I'm aware this "bug" doesn't have a great impact (maybe none?), but still I think it should be addressed. I'd happily submit changes to address it, just not sure if that is wanted.

chatman-media:
`inner_chr` was defined as `escape | ANY`, so a bare unescaped `'` could sneak through the `ANY` branch. That meant a range like `'''..'''` would compile as a char literal matching `'`, even though the only documented way to write that is the escaped form `'\''`. It's inconsistent with how `inner_str` already excludes its own delimiter, and with basically every `.pest` syntax highlighter (including the playground), which flag `'''` as invalid. Reported in #853.

Fix excludes `'` and `\` from the `ANY` branch, same as `inner_str` does. Regenerated `grammar.rs` via `cargo bootstrap` and added two tests in `meta/src/parser.rs` — one confirming `'''` is now rejected, one confirming the escaped form still parses the same as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened character literal parsing so unescaped single quotes and backslashes are no longer accepted inside single-quoted characters.
  * Confirmed that escaped single quotes continue to parse correctly.
* **Tests**
  * Added coverage for rejected unescaped quote input and accepted escaped quote input in character parsing.


tomtau (approved):
thanks. I assume people didn't rely on the old semantics/bug, but we'll see

PR #1185, opened 2026-08-05
Add support for a fully-functional const PrattParser with optional macro for increased ergonomics

NiltonVolpato:
## Summary

Adds `ConstPrattParser<R, N>`, a fully-functional Pratt parser that can be built entirely in a `const` context and stored in static memory. This is especially useful in `no_std` environments where resources are generally limited and the availability of threads and synchronization primitives is greatly reduced, since it removes the need for `lazy_static`/`LazyLock` or similar runtime initialization.

Key changes:
- `ConstPrattParser::new_const(ops: [(Op<R>, u8); N])` builds the operator table at compile time. The `u8` is a precedence delta relative to the previous operator; the first delta is `0`.
- Adds the optional `pratt_precedence!` declarative macro for convenient grouping:
  ```rust
  static PRATT: ConstPrattParser<Rule, 7> = ConstPrattParser::new_const(pratt_precedence![
      Op::infix(Rule::add, Assoc::Left) | Op::infix(Rule::sub, Assoc::Left),
      Op::infix(Rule::mul, Assoc::Left) | Op::infix(Rule::div, Assoc::Left),
      Op::infix(Rule::pow, Assoc::Right),
      Op::prefix(Rule::neg),
      Op::postfix(Rule::fac),
  ]);
  ```
- Duplicate rules are resolved by the last configuration: `ConstPrattParser::get` scans the array in reverse, matching `PrattParser::op` behavior.
- Shares parse logic between `PrattParser` and `ConstPrattParser` via a `PrattParserOps` trait.
- Updates calculator tests (raw-array case + macro case) and adds a Pratt benchmark.

## Related work

The `experimental-simd-pratt` branch explored SIMD-accelerated const lookup for 2-byte rule enums on aarch64 NEON. It showed promising numbers (~40–44 ns vs ~83 ns for the runtime BTreeMap on Apple Silicon), but it is intentionally incomplete and was **not** included here. It would need u8/u16 lanes for aarch64 and x86_64, plus runtime feature detection, before it could be considered general.

## Test plan

- [x] `cargo test` passes in `pest/pest`
- [x] `just verify` passes in the Melbi const-pratt worktree using the new API

## Summary by CodeRabbit

* **New Features**
  * Added support for constructing Pratt parsers at compile time.
  * Added a precedence macro for defining operator configurations.
  * Exposed operator precedence and associativity types for advanced parser configuration.
  * Added benchmark coverage comparing runtime and compile-time operator lookups.

* **Tests**
  * Added calculator tests verifying compile-time parser construction and precedence configuration preserve existing arithmetic behavior.

### DIFF
--- a/pest/src/pratt_parser.rs
+++ b/pest/src/pratt_parser.rs
@@ -300,13 +300,14 @@ pub struct ConstPrattParser<R: RuleType + 'static, const N: usize> {
 }
 
 impl<R: RuleType + 'static, const N: usize> ConstPrattParser<R, N> {
-    /// Create a `ConstPrattParser` from a static array of (`[`Op`]`, `u8`)
+    /// Create a `ConstPrattParser` from a static array of (`[`Op`]`, `bool`)
     /// pairs.
     ///
-    /// Just like [`PrattParser::op`], but for use in a `const` context. The `u8`
-    /// in each pair is a precedence delta relative to the previous operator:
-    /// use `0` to keep the same precedence, and `1` (or higher) to move to the
-    /// next precedence level. The first delta must be `0`.
+    /// Just like [`PrattParser::op`], but for use in a `const` context. The
+    /// `bool` in each pair tells whether the operator starts a new precedence
+    /// level (`true`) or shares the previous operator's level (`false`).
+    /// Levels are ordered from lowest to highest precedence; the first
+    /// operator must start a new level (`true`).
     ///
     /// # Example
     ///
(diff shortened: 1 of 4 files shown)


pest/src/pratt_parser.rs
```diff
@@ -41,15 +45,23 @@ pub struct Op<R: RuleType> {
     next: Option<Box<Op<R>>>,
 }
 
-enum Affix {
+/// The position of an operator relative to its operand.
+///
+/// This is an implementation detail used by the Pratt parser internals.
+#[doc(hidden)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Affix {
```
tomtau, line 53: why are `Prec` and `Affix` made public?

tomtau: is it because of PrattParserOps?

pest/src/pratt_parser.rs
```diff
@@ -252,34 +264,163 @@ impl<R: RuleType> PrattParser<R> {
...
+}
+
+/// Internal trait for types that provide operator metadata to [`PrattParserMap`].
+///
+/// This trait is an implementation detail and not intended to be implemented
+/// by downstream code.
+///
+/// [`PrattParserMap`]: struct.PrattParserMap.html
+#[doc(hidden)]
+pub trait PrattParserOps<R: RuleType> {
```
tomtau, line 280: if it's internal, does it need to be `pub`?

pest/src/pratt_parser.rs
NiltonVolpato: Making them private will result in two things:

* These warnings:
#[allow(private_interfaces)]
#[allow(private_bounds)] 
because while not used directly by the user, PrattParserOps is a bound in PrattParserMap, which is public.

* The benchmarks will not compile. At least as they are written right now.

But I can make them private, and silence the warnings, and figure out a way to rewrite the tests using only the public API. Let me know if you'd prefer that.

pest/src/pratt_parser.rs
NiltonVolpato: Same case as above. It can be private, but since it's used in PrattParserMap it will generate warnings that would have to be silenced. Just let me know what you prefer.

NiltonVolpato:
Thanks for the review. Just let me know what you prefer.

pest/src/pratt_parser.rs
tomtau: >  rewrite the tests using only the public API.

how complex do you estimate it'd be?

Anyway, I guess it may be all right to go with the public definitions (just the price to pay for fully-const is that these internal types surface as public)

tomtau (approved):
looks all right, thanks!
I'll merge the current iteration. Depending on how complex the test rewrite is, I think surfacing some of these internal types as public may be preferable / the price to pay here

PR #1187, opened 2026-08-05
fix(debugger): upgrade rustyline to 13 for loongarch64 support

tangaac:
pest_debugger fails to compile on loongarch64-unknown-linux-gnu because rustyline 10 pulls in nix 0.25, whose ioctl module does not support the loongarch64 architecture. Bump to rustyline 13, which uses nix 0.27 and adds loongarch64 support.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved debugger command-history handling for more reliable interactive sessions.
  * Added clearer feedback when a command cannot be saved to history.
  * Updated debugger terminal input support for compatibility with the latest functionality.
  * Improved persistence and loading of command history between debugger sessions.
  * Enhanced stability when entering and recalling commands from the debugger history.

### DIFF
--- a/debugger/Cargo.toml
+++ b/debugger/Cargo.toml
@@ -25,7 +25,7 @@ reqwest = { version = "= 0.11.13", default-features = false, features = [
     "json",
     "default-tls",
 ] }
-rustyline = "13"
+rustyline = { version = "13", default-features = false, features = ["custom-bindings", "with-file-history"] }
 serde_json = "1"
 thiserror = "2"
 
--- a/debugger/Cargo.toml
+++ b/debugger/Cargo.toml
@@ -25,7 +25,7 @@ reqwest = { version = "= 0.11.13", default-features = false, features = [
     "json",
     "default-tls",
 ] }
-rustyline = "10"
+rustyline = "13"
 serde_json = "1"
 thiserror = "2"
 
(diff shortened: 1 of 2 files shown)


tomtau (changes requested):
looks ok, but may require bumping the CI toolchain:
> error: rustc 1.86.0 is not supported by the following package:
>  home@0.5.12 requires rustc 1.88

tangaac:
Hi @tomtau, thanks for the review!

That error came from rustyline's `with-dirs` feature pulling in `home@0.5.12` (MSRV 1.88). It's now removed — rustyline uses `default-features = false` with only `custom-bindings` and `with-file-history`. `home` is no longer in the dependency tree, so no toolchain bump is needed and MSRV stays at 1.86.

tomtau (approved):
> Hi @tomtau, thanks for the review!
> 
> That error came from rustyline's `with-dirs` feature pulling in `home@0.5.12` (MSRV 1.88). It's now removed — rustyline uses `default-features = false` with only `custom-bindings` and `with-file-history`. `home` is no longer in the dependency tree, so no toolchain bump is needed and MSRV stays at 1.86.

thanks @tangaac! that's a better solution (hopefully it also reduces the amount of transitive dependencies that sometimes break CI)

PR #1194, opened 2026-08-18
Add PairsBuilder to construct Pairs without a parser

### LINKED ISSUE #469: Add a simple way to manually generate Pair
Hi, I am writing some unit tests for a function that takes `Pair<'_, Rule>` as an input. The problem is that I am forced to call the parser to generate those pairs, thus making my unit tests not that independent.

Example: I have

```rust
let input = EbolParser::parse(Rule::Integer, "42").unwrap();
println!("{:?}", input); // Prints: [Pair { rule: Integer, span: Span { str: "42", start: 0, end: 2 }, inner: [] }]
my_function(input.peek());
```

But I would like to have something like:
```rust
let input = Pair { rule: Rule::Integer, span: Span { str: "42", start: 0, end: 2 }, inner: [] };
my_function(input);
```

ChrisJr404:
Closes #469.

### Problem

Code that consumes parser output usually takes a `Pair<'_, Rule>` or `Pairs<'_, Rule>`. Unit-testing such code today means running a real parse first, just to get tokens of the right shape — which couples the test to the grammar and makes it awkward to exercise edge cases (a specific nesting, a particular tag) in isolation. #469 asks for a simple way to hand-build a `Pair` for exactly this.

### What this adds

A `PairsBuilder` in `pest::iterators` that describes the expected token tree directly and turns it into real `Pairs`:

```rust
use pest::iterators::PairsBuilder;

let pairs = PairsBuilder::new("1+2")
    .rule_with(Rule::sum, 0, 3, |inner| {
        inner
            .rule(Rule::number, 0, 1).tag("lhs")
            .rule(Rule::number, 2, 3).tag("rhs")
    })
    .build();

let sum = pairs.peek().unwrap();
assert_eq!(sum.as_rule(), Rule::sum);
```

- `new(input)` — start a builder over the input string.
- `rule(rule, start, end)` — append a leaf pair spanning `input[start..end]` (same `[start, end)` byte convention as `Span`).
- `rule_with(rule, start, end, |inner| ...)` — append a pair whose inner pairs are built by the closure.
- `tag(tag)` — attach a node tag to the most recently appended pair (so `find_tagged` / `as_node_tag` work).
- `build()` — flatten into the same `Start`/`End` queue the parser produces and return `Pairs`.

Because it reuses the existing queue representation, the resulting pairs behave exactly like parsed ones — `as_str`, `as_span`, `line_col`, `into_inner`, `tokens`, and the tag lookups all work unchanged. `build()` validates every span against the input (ascending range, on UTF-8 char boundaries), mirroring the invariant `Span::new` enforces, and panics with a descriptive message otherwise.

### Notes

- Purely additive — one new module, exported as `pest::iterators::PairsBuilder`; nothing else changes.
- `no_std`-friendly (only `alloc`), consistent with the rest of the crate.
- Tests cover leaves, multiple top-level pairs, nesting, tags, token round-tripping, `line_col`, multibyte spans, the empty builder, and the panic paths; docs include runnable examples. `cargo fmt --check` and `cargo clippy` are clean.

The method names / shape are of course open to bikeshedding — happy to adjust naming or the leaf-vs-`rule_with` split if you'd prefer different ergonomics.

## Summary by CodeRabbit

* **New Features**
  * Added a public builder for constructing parser rule pairs without parsing input.
  * Supports nested rules, tags, UTF-8 spans, empty inputs, and line/column calculations.
  * Provides depth-first token output through a simple fluent API.
  * Added documentation examples and coverage for common and invalid span scenarios.

### DIFF
--- a/pest/src/iterators/mod.rs
+++ b/pest/src/iterators/mod.rs
@@ -13,11 +13,13 @@ mod flat_pairs;
 mod line_index;
 mod pair;
 pub(crate) mod pairs;
+mod pairs_builder;
 mod queueable_token;
 mod tokens;
 
 pub use self::flat_pairs::FlatPairs;
 pub use self::pair::Pair;
 pub use self::pairs::Pairs;
+pub use self::pairs_builder::PairsBuilder;
 pub(crate) use self::queueable_token::QueueableToken;
 pub use self::tokens::Tokens;
--- a/pest/src/iterators/pairs_builder.rs
+++ b/pest/src/iterators/pairs_builder.rs
@@ -1,0 +1,464 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. All files in the project carrying such notice may not be copied,
+// modified, or distributed except according to those terms.
+
+use alloc::rc::Rc;
+use alloc::vec::Vec;
+
+use super::line_index::LineIndex;
+use super::pairs::{self, Pairs};
+use super::queueable_token::QueueableToken;
+use crate::RuleType;
+
+/// A builder that constructs [`Pairs`] manually, without running a parser.
+///
+/// This is meant for unit tests: code that consumes parser output often takes a
+/// [`Pair`] or [`Pairs`] as input, and testing it would otherwise require
+/// running a real parse just to obtain the right shape of tokens. With a
+/// `PairsBuilder` the expected tree can be spelled out directly.
+///
+/// A pair is described by a [`Rule`], a `start` byte offset and an `end` byte
+/// offset into `input` (the same `[start, end)` convention used by
+/// [`Span`]), and, optionally, inner pairs and a node tag.
+///
+/// [`Pair`]: struct.Pair.html
+/// [`Rule`]: ../trait.RuleType.html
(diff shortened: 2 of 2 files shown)


tomtau (commented):
thanks! Could you rebase it on top of the latest master?

One other thing I'm thinking is that given 1. this is not used anywhere inside the `pest` crates, 2. it's meant for testing (without invoking the parser), would it make sense to feature-guard under a "test" feature flag? (I know `fails_with` and `parses_to` macros are kind of in the same category, but they are used within `pest` tests)

ChrisJr404:
Rebased onto the latest master, just the one CI toolchain commit on top of the old base, no conflicts.

On the feature flag, good question. A couple of thoughts.

I'd steer away from `#[cfg(test)]` specifically: that only compiles during pest's own test builds, so downstream crates (which are the real audience here, testing their own code that consumes `Pairs`) would never see it, which defeats the point of #469.

A regular cargo feature that's off by default is the right mechanism if the goal is keeping it out of the default build. One caveat worth naming: cargo feature unification means that if a crate pulls pest in both as a normal dependency and as a dev-dependency with the feature on, the feature ends up enabled for the normal build too. So it isn't a hard guarantee, it just keeps the code out for anyone who never turns it on anywhere.

My weak preference is to leave it ungated, for two reasons: `parses_to` and `fails_with` already ship ungated in the same testing-helper category, and this is a small, purely additive, alloc-only module, so the weight it adds to the default build is tiny. That said I don't feel strongly. If you'd rather have it behind a non-default feature (something like `builder` or `test-utils`), say the word and I'll add it to this PR, it's a quick change.

tomtau (approved):

PR #895, opened 2023-07-19
Update mod.rs

TheVeryDarkness:
Fix an error of implementation of Display for OptimizedExpr, which will break the line when it contains a control character such as \n or \r.

PR #1197, opened 2026-09-03
Make Display for Expr and OptimizedExpr re-parse to the same AST

dylanpulver:
`Display for Expr`/`OptimizedExpr` prints pest source that does not always re-parse to the expression it came from.

**1. Postfix binds tighter than `&`/`!`.** `Opt(NegPred(e))` prints `!e?`, re-parsed as `NegPred(Opt(e))`:

```
r = { (!"a")? ~ "b" }           parses "b"
printed: r = { (!"a"? ~ "b") }  parses nothing (`!(a?)` can never succeed)
```

Of 224 prefix×postfix×atom grammars (5×8×8) pest accepts, **128 print to a different AST**. After: 0.

**2. `{:?}` emits `\u{1}` for U+0001..=U+000F**, but pest's own rule is `unicode = @{ "u" ~ "{" ~ hex_digit{2, 6} ~ "}" }`, so that output is not valid pest:

```
r = { "\u{0001}" }  ->  r = { "\u{1}" }  ->  parse error: expected quote
```

Over `Str`/`Insens`/`Range` for U+0000..U+3000, **36 of 36,867 literals fail to re-parse**. After: 0.

`fmt_literal` keeps every escape `{:?}` already got right (that is what keeps control characters out of doc comments, #895) and pads those twelve.

Reverting the impl fails all 5 new tests, no existing one. Two naive alternatives rejected: parenthesising *every* postfix operand round-trips but breaks 7 existing Display tests; relaxing the grammar to `hex_digit{1, 6}` instead **panics** at `parser.rs:447`, where `unescape` enforces the minimum too.

Not addressed: `a ~ b ~ c` round-trips to a left-associated `Seq` (3 of 414 rules here); semantically identical, and the rotator normalises it.

Local: 641 tests pass (636 on master); fmt, clippy `-Dwarnings`, check, doc, bootstrap clean. CI has not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected expression formatting so postfix operators preserve predicate grouping when displayed.
  * Improved escaping for low-valued control characters, ensuring displayed expressions can be parsed reliably again.
  * Updated optimized expression output to apply the same literal escaping and predicate parenthesization rules, including restored predicates.

* **Tests**
  * Added coverage for formatted expression round-trips, optimized grammars, predicate postfix operators, and control-character escaping.

### DIFF
--- a/meta/src/ast.rs
+++ b/meta/src/ast.rs
@@ -458,6 +458,48 @@ impl Iterator for ExprTopDownIterator {
     }
 }
 
+/// Grammars whose printed form has to parse back to what it was printed from.
+///
+/// Both layers round-trip the same corpus: [`Expr`] here, and
+/// [`crate::optimizer::OptimizedExpr`] in `optimizer::tests::display`. A new `Display` case
+/// belongs here rather than in a one-off test, so that both layers get it.
+#[cfg(test)]
+pub(crate) const ROUND_TRIP_GRAMMARS: &[&str] = &[
+    r#"r = { (!"a")? ~ "b" }"#,
+    r#"r = { (&"a"){2, 3} }"#,
+    r#"r = @{ (!"a"){2} }"#,
+    r#"r = { "\u{0001}" ~ ^"\u{000b}" ~ '\u{0007}'..'\u{000f}' }"#,
+    r#"r = { PUSH("a") ~ PEEK[0..-1] ~ POP }"#,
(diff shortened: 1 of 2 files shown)


tomtau (commented):
@dylanpulver maybe there could be a few E2E tests for this parsing vs display?
would this manifest https://github.com/pest-parser/pest/pull/1197#discussion_r3927490744 ?

PR #1201, opened 2026-09-07
Stop Span::lines from yielding a line past the span

### LINKED ISSUE #1205: Span::lines() yields the following line when a span ends on a line break
`Span::lines()` and `Span::lines_span()` yield one line too many when the span ends on a line break. `LinesSpan::next` (`pest/src/span.rs`) stops only once its cursor is strictly past `span.end`, but after it yields the line that ends there the cursor is exactly `span.end`, so the iterator carries on and yields the whole following line, which the span does not cover.

Spans are end-exclusive everywhere else. Take the input `"abcdef\nghi"` and the span `"abcdef\n"`, bytes 0..7: `end_line_col` already says it ends on line 1, at `(1, 8)`. `lines()` is the only thing claiming a line 2.

`Error::new_from_span` consumes that iterator and keeps its last item as `continued_line`, so the extra line shows up in rendered errors. On master:

```
 --> 1:1
  |
1 | abcdef␊
1 | ghi
  | ^-----^
  |
  = error: big one
```

Line number 1 is printed twice and `ghi` is not part of the span. The formatter itself is fine: a span that really covers two lines (0..8, `abcdef\ng`) renders as `1 | abcdef` / `2 | ghi`. The double `1` only shows up when `lines()` hands it a line the span doesn't touch.

Expected:

```
 --> 1:1
  |
1 | abcdef␊
  | ^-----^
  |
  = error: big one
```

A zero-width span should still yield the one line it sits on.

#1201 has a proposed fix, with tests for both the line-break and the empty-span case.

MaxFreedomPollard:
`Span::lines()` and `Span::lines_span()` yield one line too many whenever the span ends on a line break. `LinesSpan::next` (pest/src/span.rs, line 367 on master) stops only once its cursor is strictly past `span.end`, but the cursor is exactly `span.end` after it yields the line that ends there, so the iterator carries on and yields the whole following line, which the span does not cover.

`Error::new_from_span` consumes that iterator and keeps its last item as `continued_line`, so the extra line shows up in rendered errors. For input `"abcdef\nghi"` and a span over `"abcdef\n"`, master prints:

```
 --> 1:1
  |
1 | abcdef␊
1 | ghi
  | ^-----^
  |
  = error: big one
```

Line number 1 is printed twice and the second line of text is not part of the span. When the line that leaks in ends in a line break of its own, that break is emitted raw and splits the block with a blank line, because `continued_line` skips the whitespace visualization that the first line gets. With this change the same error prints:

```
 --> 1:1
  |
1 | abcdef␊
  | ^-----^
  |
  = error: big one
```

The fix is to stop the cursor when it reaches `span.end` instead of when it passes it. A zero-width span still sits on one line, so the guard keeps yielding on the first call, when the cursor is still `span.start`; `lines_of_empty_span` pins that behaviour down and passes both with and without the fix.

Verification on 66513d2 with Rust 1.95, macOS aarch64:

```
cargo fmt --all -- --check                                                     # clean
cargo clippy -p pest --features pretty-print,const_prec_climber,memchr,miette-error --all-targets -- -Dwarnings   # clean
cargo test -p pest --features pretty-print,const_prec_climber,memchr,miette-error   # 101 lib, 7 calculator, 11 json, 83 doctests, all pass
cargo test -p pest_meta --features grammar-extras                              # 183 pass
cargo test -p pest_derive --features grammar-extras                            # 101 pass, 12 ignored
```

With the one-line guard reverted and the tests kept, `span::tests::lines_ending_on_line_break` fails with 2 lines where 1 is expected and `error::tests::display_custom_span_end_after_inner_newline` fails with the duplicated line number shown above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected multi-line error rendering when an error span ends immediately after a line break.
  - Improved line-range handling for spans that end on a line break.
  - Preserved correct behavior for zero-width spans, ensuring their associated line remains available for display.

- **Tests**
  - Added coverage for custom error formatting, line-ending spans, and empty spans to help prevent regressions.

### DIFF
--- a/pest/src/error.rs
+++ b/pest/src/error.rs
@@ -1048,6 +1048,35 @@ mod tests {
     }
 
     #[test]
+    fn display_custom_span_end_after_inner_newline() {
+        let input = "abcdef\nghi";
+        let start = Position::new(input, 0).unwrap();
+        let end = Position::new(input, 7).unwrap();
+        assert!(start.at_start());
+        assert!(!end.at_end());
+
+        let error: Error<u32> = Error::new_from_span(
+            ErrorVariant::CustomError {
+                message: "error: big one".to_owned(),
+            },
+            start.span(&end),
+        );
+
+        assert_eq!(
+            format!("{error}"),
+            [
+                " --> 1:1",
+                "  |",
+                "1 | abcdef␊",
+                "  | ^-----^",
+                "  |",
+                "  = error: big one"
+            ]
+            .join("\n")
+        );
+    }
+
+    #[test]
     fn display_custom_span_empty() {
         let input = "";
         let start = Position::new(input, 0).unwrap();
--- a/pest/src/span.rs
+++ b/pest/src/span.rs
@@ -364,7 +364,11 @@ pub struct LinesSpan<'i> {
 impl<'i> Iterator for LinesSpan<'i> {
     type Item = Span<'i>;
     fn next(&mut self) -> Option<Self::Item> {
-        if self.pos > self.span.end {
+        // `pos` holds the start of the next line to yield, so every line the span
+        // covers has been yielded once `pos` has reached the end of the span. The
+        // exception is a zero-width span, which still sits on one line: that line is
+        // yielded by the first call, while `pos` is still the start of the span.
+        if self.pos >= self.span.end && self.pos > self.span.start {
             return None;
         }
         let pos = position::Position::new(self.span.input, self.pos)?;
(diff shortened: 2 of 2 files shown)


pest/src/error.rs
```diff
@@ -1047,6 +1047,35 @@ mod tests {
...
+            format!("{error}"),
+            [
+                " --> 1:1",
+                "  |",
+                "1 | abcdef␊",
+                "  | ^-----^",
+                "  |",
+                "  = error: big one"
+            ]
+            .join("\n")
```
tomtau, line 1075: is this test a correct expectation?
a - 0
b - 1
c - 2
d - 3
e - 4
f - 5
\n - 6
g - 7

so the error on master looks correct what it points to, but just incorrect how it's rendered (counting lines or making a line break):
```
 --> 1:1
  |
1 | abcdef␊
1 | ghi
  | ^-----^
  |
  = error: big one
```

pest/src/span.rs
```diff
@@ -364,7 +364,11 @@ pub struct LinesSpan<'i> {
 impl<'i> Iterator for LinesSpan<'i> {
     type Item = Span<'i>;
     fn next(&mut self) -> Option<Self::Item> {
-        if self.pos > self.span.end {
+        // `pos` holds the start of the next line to yield, so every line the span
+        // covers has been yielded once `pos` has reached the end of the span. The
+        // exception is a zero-width span, which still sits on one line: that line is
+        // yielded by the first call, while `pos` is still the start of the span.
+        if self.pos >= self.span.end && self.pos > self.span.start {
```
tomtau, line 371: not sure if this check is ok

tomtau (commented):
@MaxFreedomPollard I'm not sure if the fix is correct. could you perhaps first open an issue to discuss it?

MaxFreedomPollard:
Fair question, I went and checked rather than argue it.

The span is `abcdef\n`, bytes 0..7. pest already agrees that ends on line 1: `end_line_col` for it comes out as `(1, 8)`. `lines()` was the only thing claiming a line 2, and that's why you see `1 |` twice, the formatter is using `le = 1` for both rows and faithfully printing the extra line it was handed.

The formatter itself is fine. A span that genuinely does cover two lines (0..8, `abcdef\ng`) renders as `1 | abcdef` / `2 | ghi` on master today. So the double `1` isn't a numbering bug, it only shows up when `lines()` yields a line the span doesn't touch.

I also ran every span over a handful of inputs, 188 in total, before and after. 24 outputs change, every one of them a span whose end sits just after a newline, and the fix never adds a line, only drops the one past the end. The empty-span cases you flagged are byte-identical either way, which is what the `pos > span.start` half of the check is there for.

Happy to open an issue if you'd still rather work through it there.

tomtau:
@MaxFreedomPollard yes, maybe better to open an issue first to frame what the problem is: if I understand it correctly, it's that span rendering should be end-exclusive but newline endings break it to look end-inclusive?

tomtau (approved):
ok, I tried the PR and there were some rendering differences, e.g. `Span::new("abcdef\nghi", 0, 3)` used to be:

```
 --> 1:1
  |
1 | abcdef␊
  | ^-^
  |
```

but it'll be
```
 --> 1:1
  |
1 | abcdef
  | ^-^
  |
```

but I guess that's OK. Those newline characters may be a bit noisy and it's perhaps better to only render them when they are actually pointed to.

tomtau:
thanks for the fix!

